Issue #151, opened 2026-06-18
Harden scripts/backup-device-key.py per Gemini review (private-key leak on no-response + capture robustness)


Strycher:
## Summary
`scripts/backup-device-key.py` (exports a device identity **private key** over the serial CLI)
needs hardening per its Gemini review before it's committed/used. The happy path is fine; the
error/capture paths are not. Found 2026-06-17.

## Findings
- **[CRITICAL] Key leak on regex mismatch** — the no-response branch writes the *entire raw
  serial buffer* (which may contain the private key) to `<name>-prvkey-NORESPONSE.log`. Fix:
  don't dump the raw buffer; report a char-count only.
- **[HIGH] Truncated-key capture** — the fixed-time read can stop mid-transmission; the `{32,}`
  regex then matches a partial key and saves it as "success." Fix: read until the reply is
  complete (key + trailing newline / next prompt), not a fixed timer.
- **[HIGH] Ambiguous regex** — `>\s*([0-9A-Fa-f]{32,})` grabs the first hex (could match a
  pubkey/EUI). Mitigated by `reset_input_buffer()`; anchor it + enforce the MeshCore key length
  (128 hex).
- **[MEDIUM] Unhandled `bytes.fromhex` ValueError** — odd-length capture → traceback leaks
  `key_hex` to stderr. Wrap in try/except.
- **[MEDIUM] Silent overwrite** of an existing backup — add `--force`.
- **[LOW] gitignore reliance** — add a loud "ensure device_backups/ is gitignored" note.

Audit log: `docs/llm-consultations/2026-06-17-key-backup-tooling-review-gemini-gemini-2.5-pro.log`

Strycher:
Hardened (Gemini re-review clean) + bench-verified on the 1W RAK3401: key saved, sha256 head matches the prior backup, key never printed. Landing via PR from fix/151-harden-key-backup.

PR #155, opened 2026-06-19
fix(#151): harden backup-device-key.py (key-leak + capture robustness)

Strycher:
Hardens the device private-key backup tool per its Gemini review (#151), plus a fix from bench-testing.

## Fixes
- **[CRITICAL]** no-response path no longer dumps the raw serial buffer (may contain the key) to a log
- **[HIGH]** read-until-complete `> <hex>\n` line (no truncated-key saves); **[HIGH]** regex requires ≥64 hex (rejects short EUI/pubkey)
- **[MED]** `bytes.fromhex` try/except (no traceback leak); **[MED]** `--force` overwrite guard
- **[LOW]** runtime gitignore check — corrected to check the key-file path, not the bare dir (a false-negative the bench-test caught)

## Verification
- 6/6 no-device parse tests pass; Gemini re-review **clean** (0 findings)
- **Bench-verified on the 1W RAK3401 (COM20):** key saved, sha256 head matches the prior backup, key never printed

For #151 (closed per the close-then-push convention).

### DIFF
--- a/scripts/backup-device-key.py
+++ b/scripts/backup-device-key.py
@@ -1,0 +1,147 @@
+#!/usr/bin/env python3
+"""Back up a device's MeshCore identity (private key) over the serial CLI.
+
+Sends `get prv.key` to the device's CommonCLI serial console (local-serial-only
+command) and saves the exported private key to
+device_backups/<name>-prv.key.txt (gitignored).
+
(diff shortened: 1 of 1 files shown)


Strycher:
Human approval received (Ben: "Merge 155"). Enabling auto-merge; merges when CI passes.

Issue #199, opened 2026-06-24
bug: xiao_c6 companion_radio_ble missing NimBLE dep — BLE env can't build


Strycher:
## Bug
`Xiao_C6_companion_radio_ble_` fails to build: `fatal error: NimBLEDevice.h: No such file or directory`. The BLE companion env legitimately uses BLE (compiles `helpers/esp32/SerialBLEInterface.cpp`), but its `lib_deps` is **missing** `h2zero/NimBLE-Arduino @ ^2.0.0` — the dep our other working BLE envs (heltec_v3/v4, xiao_s3_wio) all declare.

**Distinct from #89 / #90**: those are the *non-BLE* envs (`_usb`/`_serial`/`_wifi`) wrongly compiling SerialBLEInterface. This is a *BLE* env missing the dep outright. xiao_c6 isn't in #90's affected list.

Surfaced 2026-06-24 while porting the MeshSmith Photon-1W (#193); the Photon BLE env had the identical gap (fixed in-port by adding the NimBLE dep).

## Fix
Add `h2zero/NimBLE-Arduino @ ^2.0.0` to `Xiao_C6_companion_radio_ble_`'s `lib_deps` (1 line; matches the Photon fix + heltec/xiao_s3_wio pattern).

## Also audit
Sweep all `*_companion_radio_ble` envs for the same missing dep, so no other BLE env is silently broken.

## Acceptance
- [ ] `Xiao_C6_companion_radio_ble_` builds clean.
- [ ] No other BLE companion env missing the NimBLE dep (or each tracked).

Strycher:
## The "Also audit" sweep is done — this is systemic, and 4 **shipped** envs are broken

Ran the audit this issue asked for while building for #298 (which hit the same error on `heltec_tracker_v2_companion_radio_ble`). It is not one or two stragglers.

### Root cause is structural
Only **3 of 32** ESP32 variants that define a `*companion_radio_ble` env declare a NimBLE dep: `heltec_v3`, `heltec_v4`, `xiao_s3_wio`. And **no shared/base config provides it** (nothing in the root `platformio.ini` or `arch/`), so every ESP32 BLE env must declare it individually — which is exactly why this keeps recurring one board at a time.

*(nRF52 BLE envs are irrelevant here — they use the Nordic/Bluefruit stack, not NimBLE. An earlier naive sweep flagged ~60 "suspects"; filtering to ESP32-only is what makes the list meaningful.)*

[...]

Issue #90, opened 2026-06-11
bug: non-BLE companion envs (heltec_v3/v4 + xiao serial/wifi) share #288 NimBLE build regression


Strycher:
## Bug (deferred — "fix properly later")
Same root cause as Strycher/Crosswire#89: the #288 NimBLE migration added NimBLE-Arduino **only to the `_ble` envs**, but the **`_usb` / `_serial` / `_wifi` companion envs** on every NimBLE board still compile `SerialBLEInterface.cpp` (via `+<helpers/esp32/*.cpp>`) without the dep → they don't build.

## Affected (to verify + fix)
- `heltec_v3` companion `_usb` / `_serial` / `_wifi`
- `heltec_v4` companion `_usb` / `_serial` / `_wifi`
- `xiao_s3_wio` companion `_serial` / `_wifi` (the `_usb` is fixed under #89)

## Fix options
- Exclude `SerialBLEInterface.cpp` from each non-BLE companion env's `build_src_filter` (matches #89), **or**
- Guard `SerialBLEInterface.h` behind `NIMBLE_ENABLED` so it compiles to nothing without BLE (cleaner, addresses the "transitional" #288 state at the source).

## Acceptance
- [ ] All non-BLE companion envs on heltec_v3/v4/xiao_s3_wio build clean.

Strycher:
Fixed by the #199 work (commit `428658e8`). All non-BLE companion envs on heltec_v3 / heltec_v4 / xiao_s3_wio now either exclude `SerialBLEInterface.cpp` via the new shared `${esp32_no_ble.build_src_filter}` or legitimately carry NimBLE. Took the #89 exclusion pattern rather than the NIMBLE_ENABLED header guard, to keep src/ aligned with upstream MeshCore for clean cherry-picks. Verified: Heltec_v3_companion_radio_wifi, heltec_v4_companion_radio_wifi, Xiao_S3_WIO_companion_radio_serial all build.

PR #200, opened 2026-06-24
fix(#89): exclude SerialBLEInterface.cpp from Xiao_S3_WIO USB companion

### LINKED ISSUE #89: bug: Xiao_S3_WIO_companion_radio_usb fails to build — NimBLEDevice.h (regression from #288)
## Bug
`pio run -e Xiao_S3_WIO_companion_radio_usb` fails to compile:
```
src/helpers/esp32/SerialBLEInterface.h:4: fatal error: NimBLEDevice.h: No such file or directory
```

## Root cause
The #288 NimBLE migration (`562347be`, `b25bc3fd`) added the NimBLE-Arduino lib **only to the `_ble` env**, but this `_usb` env still compiles `SerialBLEInterface.cpp` via the greedy `build_src_filter += <helpers/esp32/*.cpp>`. The USB build never *uses* BLE — `main.cpp:56` gates the BLE interface behind `defined(BLE_PIN_CODE)`, which this env does not set. So the file is compiled but never referenced. Full analysis: Strycher/Crosswire#86 (comment).

## Fix (verified safe)
Add `-<helpers/esp32/SerialBLEInterface.cpp>` to `Xiao_S3_WIO_companion_radio_usb`'s `build_src_filter`. No missing symbol (USB build never references it).

## Scope
This env only — needed to unblock device provisioning (#86). The same regression on `heltec_v3`/`v4` + the `_serial`/`_wifi` companion envs is tracked separately ("fix properly later").

## Acceptance
- [ ] `pio run -e Xiao_S3_WIO_companion_radio_usb` builds clean.

Strycher:
## fix(#89): Xiao_S3_WIO USB companion NimBLE build regression

Closes #89.

`Xiao_S3_WIO_companion_radio_usb` failed with `fatal error: NimBLEDevice.h: No such file or directory` — the #288 NimBLE migration added NimBLE only to `_ble` envs, but the USB env still greedy-compiled `helpers/esp32/SerialBLEInterface.cpp` without the dep. The USB companion never uses BLE (gated on `BLE_PIN_CODE`, undefined here), so the fix excludes that NimBLE-only file from the env's `build_src_filter`.

1-line change; build verified green on current firmware-base (Flash 18.6%). Re-applies the original fix (`d89561c4`) that had been stranded on an unmerged stale branch.

Related (separate, tracked): **#90** (non-BLE serial/wifi envs across boards), **#199** (xiao_c6 BLE missing NimBLE dep).

### DIFF
--- a/variants/xiao_s3_wio/platformio.ini
+++ b/variants/xiao_s3_wio/platformio.ini
@@ -141,6 +141,7 @@ build_flags =
 build_src_filter = ${Xiao_S3_WIO.build_src_filter}
   +<helpers/ui/SSD1306Display.cpp>
   +<helpers/esp32/*.cpp>
+  -<helpers/esp32/SerialBLEInterface.cpp>   ; #89: USB companion has no BLE (gated on BLE_PIN_CODE); drop NimBLE-only file
   +<helpers/ui/MomentaryButton.cpp>
   +<../examples/companion_radio/*.cpp>
   +<../examples/companion_radio/ui-new/*.cpp>


Strycher:
Human merge approval received (Ben, 2026-06-24). Enabling auto-merge (--rebase) per DW flow.

Issue #95, opened 2026-06-12
Epic: default the MQTT broker pool pre-config (CoreScope/LetsMesh/Eastme.sh/MeshMapper/EU/AU + per-device auth)


Strycher:
## Goal
Ship the observer with the core MQTT broker pool **pre-configured by default** so operators don't hand-enter every broker. Extends the partial Phase-2 work (Crosswire-03m) with this session's verified auth findings.

## Broker layout (owner-specified)
| Slot | Broker | Transport | ca_cert | jwt_audience | Status |
|---|---|---|---|---|---|
| 0 | **CoreScope (Dayton)** | tcp / anon | — | — | known: `mqtt.w8oof.net:1883` |
| 1 | **LetsMesh.net (US)** | wss/jwt | `gts-r4` | `mqtt-us-v1.letsmesh.net` (bare) | known + verified |
| 2 | **Eastme.sh (US)** | wss/jwt | `letsencrypt` | `mqtt.eastme.sh` | known + verified |
| 3 | **MeshMapper** | TBD | TBD | TBD | **NEW — never tried; research host/cert/auth** |
| 4 | **LetsMesh (EU)** | wss/jwt | TBD (likely `gts-r4`) | TBD | **verify EU site + host/audience** |
| 5 | **Eastmesh.au** | wss/jwt | `letsencrypt` | `mqtt2.eastmesh.au` | **verify still exists** (`wss://mqtt2.eastmesh.au:443/mqtt` per NVS test fixture) — may have consolidated into eastme.sh |
| 6 | **User config** | — | — | — | left blank for the operator |

## CRITICAL design constraint — per-device auth ("confirm this works")
`username` and `jwt_owner` are **device-specific** (the node's own Ed25519 pubkey) and **cannot be static defaults**:
- `mqtt.broker.<N>.username` = **`v1_<UPPERCASE 64-hex pubkey>`** (broker strips `v1_`, verifies the token's `publicKey` claim)
- `mqtt.broker.<N>.jwt_owner` = **`<UPPERCASE 64-hex pubkey>`** (own pubkey; #63 normalizes to uppercase)
- `jwt_email` = operator email (per-user; default from a configured owner-email if present)
- password = the minted JWT (firmware mints; #68)

[...]

PR #282, opened 2026-07-06
Updated EastMe.sh to CoreComms.net

EldoonNemar:


Strycher:
Hi @EldoonNemar — first, an apology for how long this sat without a response; that's on us, not you.

Your rename is now landed on firmware-base via #679 (merge commit 1a0b7edb), with you credited as Co-Authored-By on the commit and thanked in the CHANGELOG. We re-implemented rather than merging this branch only because the files it touches (ConfigSchema, MqttAuth) were heavily restructured after July 6 (slot-layout #95, reseeds #390/#593, the 1.17.0 base update), which left this PR in an unresolvable conflict state — nothing wrong with the change itself.

Two things your contribution directly triggered: we live-verified mqtt.corecomms.net (WSS/MQTT upgrade + full TLS chain to GTS Root R4 — your gts-r4 anchor choice was exactly right), and in the process we found the Eastmesh.au slot's old letsencrypt pin no longer validates at all, so that got fixed in the same landing.

Thanks for the contribution and the rebrand heads-up — PRs like this one are genuinely welcome, and we'll be faster next time.

PR #291, opened 2026-07-14
fix(#290): parse ESP32-C6/H2 base MAC in pio-flash bootstrap

### LINKED ISSUE #290: bootstrap mis-parses ESP32-C6/H2 MAC (records EUI-64 head, not base MAC)
## Summary

`scripts/pio-flash.py` → `cmd_bootstrap` records the **wrong MAC** for ESP32-C6 / ESP32-H2 (any IEEE 802.15.4-capable chip). It stores a corrupted address with the EUI-64 fill pattern `ff:fe` embedded instead of the chip's true base MAC.

## Root cause

The MAC is parsed with:

```python
m = re.search(r"MAC:\s*([0-9a-fA-F:]{17})", result.stdout)
```

This grabs the **first** `MAC:` line and truncates it to 17 chars (6 bytes). On 802.15.4 chips, esptool v5.x prints the **8-byte EUI-64** as the first `MAC:` line, whose first 6 bytes are the EUI-64 head (`...:ff:fe:...`), **not** the base MAC. The correct value is on the separate `BASE MAC:` line.

## Evidence (verified 2026-07-14)

> **Note:** MAC values below are **synthetic placeholders** illustrating the output *shape*. The real device MAC is not reproduced here — it lives only in the gitignored `hardware-devices.yaml` (redaction discipline; device MACs are treated as sensitive).

Registering an ESP32-C6 via bootstrap, `esptool v5.2.0 read-mac` prints this shape:

```
Chip type:  ESP32-C6FH4 (QFN32) (revision v0.2)
MAC:        02:71:bc:ff:fe:12:34:56      <- 8-byte EUI-64 (parsed by mistake)
BASE MAC:   02:71:bc:12:34:56            <- true 6-byte MAC (should be used)
MAC_EXT:    ff:fe
```

- Bootstrap stored: `02:71:bc:ff:fe:12`  ❌ (first 6 bytes of the EUI-64)
- True base MAC:     `02:71:bc:12:34:56`  ✅

## Impact

Every C6/H2 device registered via `bootstrap` gets a wrong MAC silently. The MAC is used for the device-identity collision cross-check, so a mis-registered MAC weakens that safety check for the whole 802.15.4 device class. S3/C3 (which print a plain 6-byte first `MAC:` line) are unaffected.

## Proposed fix

Prefer the `BASE MAC:` line; fall back to a strict 6-byte `MAC:` match (with a negative lookahead so it never captures the first 6 bytes of an 8-byte EUI-64). Add a unit test covering both C6 (EUI-64 + BASE MAC) and S3 (plain MAC) esptool output shapes.

## Note — canonical copy

The wrapper's canonical copy lives in `C:\Dev\LoRa` (per the `block-raw-flash` hook message). The same fix should be applied there in a separate PR against that repo; this issue covers `OffbandMesh/meshcore-firmware` only.

**Affected:** `scripts/pio-flash.py` (`cmd_bootstrap`, MAC-parse regex ~L1373)

Strycher:
## What & why

`pio-flash bootstrap` recorded the **wrong MAC** for ESP32-C6 / H2 (802.15.4 chips). `esptool read_mac` prints an 8-byte **EUI-64** as the first `MAC:` line; the old regex `MAC:\s*([0-9a-fA-F:]{17})` grabbed its first 6 bytes (the `ff:fe` EUI-64 head) instead of the true base MAC on the `BASE MAC:` line. This silently mis-registers every C6/H2 device and weakens the MAC-based identity collision check. Found while registering a Photon-1W XIAO ESP32-C6.

Closes #290.

## Change

- **`scripts/pio-flash.py`** — new `parse_base_mac()`: prefers the `BASE MAC:` line, falls back to a strict 6-byte `MAC:` match whose negative lookahead refuses to capture the head of an 8-byte EUI-64. `cmd_bootstrap` now calls it. S3/C3 (plain 6-byte `MAC:`) behavior unchanged.
- **`scripts/test_pio_flash_mac.py`** — self-contained regression test (no pytest dep). Covers C6 (EUI-64+BASE), S3 plain, S3+BASE, uppercase, no-MAC→None, plus a guard asserting the *old* regex reproduces the bug.

## Test evidence

```
[PASS] C6 EUI-64 -> BASE MAC   [PASS] S3 plain MAC   [PASS] S3 with BASE MAC
[PASS] uppercase normalized    [PASS] no MAC -> None
All 5 cases passed.
```
Red→green confirmed (test failed pre-fix: helper absent). `python -m py_compile` clean.

## Gemini adversarial review

Verdict **Ship**. One MINOR — flagged the `2026-07-14` comment date as a "future typo (→2024)". **Rejected:** 2026-07-14 is the actual current date; the finding is an artifact of the reviewer's training cutoff. No other findings.

## Redaction

Fixtures + a code comment initially carried a real device MAC; now fully synthetic (`02:…` locally-administered). Real MACs live only in the gitignored `hardware-devices.yaml`. Issue #290 body was also redacted to placeholders.

## Follow-up (not in this PR)

The wrapper's **canonical copy in `C:\Dev\LoRa`** needs the same fix in a separate PR against that repo.

### DIFF
--- a/scripts/pio-flash.py
+++ b/scripts/pio-flash.py
@@ -1316,6 +1316,29 @@ def cmd_erase_region(args, registry):
     return rc
 
 
+# --- MAC parsing (#290) ------------------------------------------------------
+# On IEEE 802.15.4 chips (ESP32-C6 / ESP32-H2) `esptool read_mac` prints an
+# 8-byte EUI-64 as the FIRST "MAC:" line (e.g. 02:71:bc:ff:fe:12:34:56); its
+# first 6 bytes carry the ff:fe EUI-64 fill and are NOT the device base MAC.
+# The real 6-byte address is on the "BASE MAC:" line. S3/C3 print only a plain
+# 6-byte "MAC:" line (some builds also add "BASE MAC:"). Prefer BASE MAC; fall
+# back to a strict 6-byte MAC whose negative lookahead refuses to capture the
+# head of an 8-byte EUI-64.
+_MAC6 = r"([0-9a-fA-F]{2}(?::[0-9a-fA-F]{2}){5})"
+_BASE_MAC_RE = re.compile(r"BASE\s+MAC:\s*" + _MAC6, re.IGNORECASE)
+_PLAIN_MAC_RE = re.compile(r"MAC:\s*" + _MAC6 + r"(?![0-9a-fA-F:])", re.IGNORECASE)
+
+
+def parse_base_mac(stdout):
+    """Extract the device base MAC (lowercased) from `esptool read_mac` output.
+
+    Returns the 6-byte MAC as "aa:bb:cc:dd:ee:ff", or None if none is found.
+    See #290 for the C6/H2 EUI-64 mis-parse this replaces.
+    """
+    m = _BASE_MAC_RE.search(stdout) or _PLAIN_MAC_RE.search(stdout)
+    return m.group(1).lower() if m else None
+
+
 def cmd_bootstrap(args, registry):
     """
     Register a new device. User-initiated only. Performs ONE authorized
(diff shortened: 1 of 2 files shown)


Strycher:
Human approval received from owner (Ben) to merge. Enabling auto-merge (--rebase); will land when CI is green.

PR #328, opened 2026-07-20
fix(#199, #90): correct all 64 ESP32 envs that compile SerialBLEInterface.cpp

### LINKED ISSUE #199: bug: xiao_c6 companion_radio_ble missing NimBLE dep — BLE env can't build

Strycher:
Fixes #199 and #90.

## What was wrong

The reported symptom was 4 shipped `*_companion_radio_ble` envs that could not build (`fatal error: NimBLEDevice.h`). The actual scope was **64 ESP32 envs**, and **2 of them are in the release matrix** — `Heltec_v2_companion_radio_usb` and `Xiao_C3_companion_radio_usb`. Because `release.yml` runs `fail-fast: false`, those boards' binaries were **silently absent from v1.1.2** rather than failing the release. Verified against the published release assets: both boards have repeater and room-server artifacts but no companion.

Root cause is not the missing dependency. Companion envs pull the whole helpers directory in with `+<helpers/esp32/*.cpp>`, and `SerialBLEInterface.h` includes `<NimBLEDevice.h>` unconditionally — no `#ifdef` guard. So an env compiles the BLE interface whether or not the board uses BLE.

## The split

`examples/companion_radio/main.cpp:64` includes that header only under `#elif defined(BLE_PIN_CODE)`, which makes `BLE_PIN_CODE` the authoritative discriminator:

| Class | Count | Remedy |
|---|---|---|
| Genuinely BLE | 30 | `${esp32_ble.lib_deps}` |
| USB / WiFi / serial | 34 | new `${esp32_no_ble.build_src_filter}` |

Excluding the file rather than handing non-BLE envs the dependency is deliberate — declaring NimBLE would make the error vanish while linking a full Bluetooth stack into firmware that never uses it. Follows #89, which fixed `Xiao_S3_WIO_companion_radio_usb` the same way.

## Prevention

`scripts/check_esp32_ble_deps.py` asserts the invariant against the **resolved** PlatformIO config (these values inherit through `extends`, so grepping `.ini` files cannot see them), wired into CI as a fast `config-lint` job covering every env rather than a curated subset. Both previously-broken shipped envs added to the build matrix.

## Verification

- Resolved-config diff across every env, before vs after: 30/30 now carry NimBLE, 34/34 now exclude the file, **0 envs outside the work list changed, 0 lost a dependency**.
- All 64 changed envs were already unbuildable, so this cannot regress a shipping binary.
- Built: `Heltec_v2_companion_radio_usb`, `Heltec_v2_companion_radio_wifi`, `Tbeam_SX1262_companion_radio_ble`, `Xiao_C3_companion_radio_usb`, `M5Stack_Unit_C6L_companion_radio_ble` (C6 platform package), plus #90's named envs `Heltec_v3_companion_radio_wifi`, `heltec_v4_companion_radio_wifi`, `Xiao_S3_WIO_companion_radio_serial`.
- Lint negative-tested: reverting one variant makes it exit 1 naming that env.

Gemini review: no defects; confirmed src-filter entries apply in order and `-<path>` reliably overrides an earlier `+<glob>`.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,6 +17,32 @@ on:
   workflow_dispatch:
 
 jobs:
+  # #199: build-config invariant check. Seconds, and it covers EVERY env rather
+  # than the curated build matrix below -- which is the point. The NimBLE defect
+  # recurred four times board-by-board because the affected envs were neither in
+  # CI nor able to fail a release (release.yml is fail-fast: false, so a broken
+  # env just silently produced no binary). A per-env build matrix would never
+  # have caught it without building all 114; this asserts the rule instead.
+  config-lint:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v5
+
+      - uses: actions/setup-python@v6
+        with:
+          python-version: '3.12'
+
+      - name: Install PlatformIO
+        run: pip install --upgrade platformio
+
+      - name: Generate dummy secrets stub (config interpolation only)
+        run: |
+          printf '[node_secrets]\nadmin_password = ci-dummy\n[wifi_secrets]\nssid = ci-dummy\npassword = ci-dummy\n[mqtt_secrets]\nhost = mqtt.example.com\nport = 1883\nuser = ci-dummy\npassword = ci-dummy\n[ota_secrets]\ntrigger_secret = ci-dummy\n[cmdrelay]\nurl = http://cmdrelay.example.com/cmd\n' > "$RUNNER_TEMP/ci-secrets.ini"
+          echo "PIO_SECRETS_FILE=$RUNNER_TEMP/ci-secrets.ini" >> "$GITHUB_ENV"
+
+      - name: Every ESP32 env compiling SerialBLEInterface.cpp declares NimBLE
+        run: python scripts/check_esp32_ble_deps.py
+
   build:
     runs-on: ubuntu-latest
     strategy:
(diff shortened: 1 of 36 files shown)


Strycher:

Issue #318, opened 2026-07-19
bug: Heltec V4 OLED reports no FEM LNA capability — instrument FEM detection to establish root cause


Strycher:
**Symptom:** a Heltec V4 OLED (lab device ST-P) running `heltec_v4_companion_radio_ble` @ `cf95650` reports **no** FEM LNA capability to the client — `OFFBAND_CAP_FEM_LNA` (`0x04`) is clear in the device-info caps byte, so the client correctly hides the setting.

**Why that is not credible as-is:** the same physical board previously ran as the patio repeater with FEM + LNA **enabled**, at ~30 dBm combined FEM + TX, under the pre-Offband LoRa repeater firmware. It was ordered in the same order, same day, same package as the HV4 currently on the patio, which runs FEM enabled today. A claim that this unit "has no FEM" contradicts operating history and is rejected.

## Claim status

`[verified: source]` — `LoRaFEMControl::init()` (`variants/heltec_v4/LoRaFEMControl.cpp`) calls `setLnaCanControl(true)` **only** in the `KCT8103L_PA` branch. The `GC1109_PA` branch leaves `lna_can_control` at its `false` default.

`[verified: source]` — `init()` is called from `HeltecV4Board::begin()`, so it does run on the companion.

`[verified: source]` — `canControlLoRaFemLna()` returns `loRaFEMControl.isLnaCanControl()`; `MyMesh` sets the cap bit directly from it.

`[verified: observed]` — the client shows no FEM/LNA setting for this device, i.e. the bit is clear.

**`[hypothesis: untested]`** — the board is a GC1109 (V4.2) part.
**`[hypothesis: untested]`** — the detection misfires and a KCT8103L reads as GC1109.
**Not established:** which FEM chip is physically on this board. Neither hypothesis has been tested. No action may be taken on either until it is.

## Note on terminology

Both V4 FEM variants (KCT8103L and GC1109) include an LNA and amplify RX. `lna_can_control` is **not** "has a FEM" — it is "this code exposes an independent LNA bypass line for the detected part." Conflating the two is what produced the wrong framing above.

## The test

Instrument `LoRaFEMControl::init()` to log, at boot, before and after detection:

- raw `digitalRead(P_LORA_KCT8103L_PA_CSD)` value (the actual pin level the decision is made on)
- resolved `fem_type`
- resolved `lna_can_control`
- `getManufacturerName()` result

Flash to ST-P, capture serial. One line of output resolves it:

[...]

PR #329, opened 2026-07-20
feat(#298): apply FEM LNA on all FEM-capable boards + 0xC3 client control surface

### LINKED ISSUE #298: Apply the FEM LNA setting on all FEM-capable boards (companion role never applied radio_fem_rxgain)
**Context:** upstream report [meshcore-dev/MeshCore#2732](https://github.com/meshcore-dev/MeshCore/issues/2732) — MeshCore 1.16 disables the Heltec V4 (V4.3 / KCT8103L) FEM LNA by default with no user control. Offband already solved the control half for **repeaters**: persisted `radio_fem_rxgain` pref (kept at pre-1.16 offset 291, `src/helpers/CommonCLI.cpp:171-178`), `fem on|off|status` CLI (`src/helpers/CommonCLI.cpp:570-584`), and repeater boot default LNA **ON** applied to hardware (`examples/simple_repeater/MyMesh.cpp:922` + `:973-974`).

**Cause:** `board.setLoRaFemLnaEnabled(...)` is called from `simple_repeater` only — verified by repo-wide grep. Companion (and room_server/sensor) roles never apply the pref, so the FEM falls to the class default `lna_enabled=false` (`variants/heltec_v4/LoRaFEMControl.h:27`) = LNA bypassed. Heltec V4 companion builds run with degraded RX sensitivity — the exact 1.16-base behavior complained about upstream. Affects real fleet hardware (V4 OLED BLE companion).

**Ask:** apply `radio_fem_rxgain` at boot in the companion role (same pattern as `simple_repeater/MyMesh.cpp:973-974`, guarded by `board.canControlLoRaFemLna()`), with the same LNA-ON default. Decide whether room_server/sensor get the same treatment in this pass or a follow-up. Bench note: KCT8103L LNA gain is undocumented (see HARDWARE.md FEM section) — before/after RSSI/noise-floor readings on the V4 companion would quantify the win.

Strycher:
Fixes #298.

## What this delivers

The companion role never applied the FEM LNA setting, so V4 companions ran with degraded RX sensitivity while the repeater role had it right. This adds the boot-time apply, board support, and the client-facing control surface.

- **Boot apply** from persisted `radio_fem_rxgain` (default on), behind a capability check
- **Board support** on `heltec_v4`, `heltec_tracker_v2`, `heltec_t096`
- **`0xC3` companion-API command** — SET (`0x01`) / GET (`0x02`), reply `[0xC3][sub][value]`, malformed → `[0x01][0x06]`
- **`OFFBAND_CAP_FEM_LNA = 0x04`** cap bit, `FIRMWARE_VER_CODE` 15 → 16
- **LNA state in the device-info reply** at offset 83 (frame length now 84), appended unconditionally so the offset never depends on hardware

Wire contract was negotiated with the client-side agent and accepted verbatim. Three deviations were flagged on the issue rather than left to surface at integration: the SET reply carries post-apply hardware state rather than echoing the request; a non-capable board answers `ERR_CODE_UNSUPPORTED_CMD`; and `savePrefs()` only runs on an actual change (Gemini caught that an unconditional save would put a slider's `onChange` burst straight onto flash).

## Hardware-verified

Owner confirmed the setting works end-to-end with the client on **Firestar (KCT8103L / V4.3)**.

Capability is **per-unit, not per-model** — and that is now demonstrated, not asserted: two Heltec V4 OLEDs from the same order resolve differently. Firestar is KCT8103L and exposes the control; ST-P is GC1109 and correctly does not. The client must gate on the cap bit, never on a model allowlist. See #318 for the detection evidence.

`rak3401` stays excluded by design — its SKY66122 gates LNA and PA off one line, so a toggle would kill TX.

## Builds

`heltec_v4_companion_radio_ble` (FEM-capable), `Heltec_v3_companion_radio_ble` (base `MainBoard` virtuals, no FEM override), `RAK_4631_companion_radio_ble` (nRF52), plus all 10 non-companion envs on the two boards whose classes changed (repeater, room server, sensor, kiss modem, terminal chat).

### DIFF
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -241,6 +241,11 @@ void DataStore::loadPrefsInt(const char *filename, NodePrefs& _prefs, double& no
     file.read((uint8_t *)&_prefs.rx_boosted_gain, sizeof(_prefs.rx_boosted_gain));         // 89
     file.read((uint8_t *)_prefs.default_scope_name, sizeof(_prefs.default_scope_name));    // 90
     file.read((uint8_t *)_prefs.default_scope_key, sizeof(_prefs.default_scope_key));     // 121
+    // #298: APPEND-ONLY, must stay last. Prefs files written before this field ended at
+    // 137, so on those this read short-reads to EOF and leaves the caller-set default
+    // (LNA ON) intact. A uint8 append is all-or-nothing, so there is no partial-read
+    // hazard. Never insert a new field mid-sequence.
+    file.read((uint8_t *)&_prefs.radio_fem_rxgain, sizeof(_prefs.radio_fem_rxgain));      // 137
 
     file.close();
   }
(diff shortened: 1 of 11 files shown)


Strycher:

Issue #312, opened 2026-07-19
CommonCLI: unbounded strcpy of command text into tmp[132] (set radio / tempradio / sensor set)


Strycher:
**Context:** found by the Gemini review during #299 (CLI token matching). Pre-existing — not introduced by that change, and out of its scope, so filing separately.

**Cause:** `CommonCLI` copies attacker-influenced command text into a fixed member buffer with an unbounded `strcpy`. The destination is

```c
char tmp[PRV_KEY_SIZE*2 + 4];   // src/helpers/CommonCLI.h:128  -> 64*2+4 = 132 bytes
```

sized for the `prv.key` hex dump, not for command text. Three call sites copy into it without a length check:

| Site | Command |
|---|---|
| `src/helpers/CommonCLI.cpp:374` | `strcpy(tmp, &command[10]);` — `tempradio ` |
| `src/helpers/CommonCLI.cpp:417` | `strcpy(tmp, &command[11]);` — `sensor set ` |
| `src/helpers/CommonCLI.cpp:802` | `strcpy(tmp, &config[6]);` — `set radio ` |

**Reachability:** the remote-admin path passes a pointer directly into the decrypted packet payload —

```c
uint8_t temp[166];
char *command = (char *)&data[5];   // examples/simple_repeater/MyMesh.cpp:721
```

so `command` can carry on the order of ~179 bytes. A command such as `set radio ` followed by >131 characters overruns `tmp` and corrupts whatever `CommonCLI` members follow it. Reachable **over the air**, though it requires an authenticated admin client (the command path sits behind client login), and locally over serial. Room server and sensor roles route through the same `_cli.handleCommand`.

Not remote-unauthenticated, so this is not a drop-everything issue — but it is memory corruption from a network-supplied string, and the 132-byte bound is incidental rather than intentional.

**Ask:**
1. Bound all three copies to `sizeof(tmp)` (`StrHelper::strncpy` or `snprintf`) and reject over-length input with an explicit error rather than truncating silently — truncation would resurrect the same class of silent-accept bug #299 just fixed.
2. Audit `CommonCLI` for other unbounded copies out of `command` / `config`, and for any other buffer sized for one purpose but reused for command text.
3. Consider a host-side test in the `native` env feeding an over-length command through the dispatcher, which would pin this and #299 together.

[...]

PR #330, opened 2026-07-20
fix(#299): whole-token match for CLI get keys + reject extra set radio params

### LINKED ISSUE #299: CommonCLI 'get' prefix-matching accepts garbage suffixes (get radio foobar returns radio settings)
**Context:** confirmed locally while triaging upstream [meshcore-dev/MeshCore#2732](https://github.com/meshcore-dev/MeshCore/issues/2732) (same defect reported there against 1.15/1.16; inherited with the 1.16 base).

**Cause:** the `get` handler chain matches bare `memcmp` prefixes with no end-of-token check — e.g. `src/helpers/CommonCLI.cpp:1038` matches `radio`, so `get radio foobar` (or any garbage suffix) silently returns the radio settings instead of an unknown-command error. Same pattern applies to sibling `get` keys (`lat`/`lon`/etc. — audit the chain while in there). Upstream repro: user typed `get radio.fem.rxgain` (a command that does not exist in mainline) and got the radio settings back, concluding the knob was broken.

**Ask:** after each prefix match, require the next char to be `\0` or a space (or switch the chain to exact-token compare), and return the standard unknown-command error otherwise. Cosmetic severity, but it actively misleads users typing commands from other forks/versions — exactly what happened in the upstream thread. Host-side unit test on the `get` dispatcher would pin it.

Strycher:
Fixes #299.

## The first pass did not fix the reported bug

Worth stating plainly, because the issue's own suggested remedy is the broken one. #299 asks to *"require the next char to be `\0` or a space."* That was implemented first, and `get radio foobar` **still** returned the radio settings — the space clause is the hole.

The correct rule is **whole-token**: every key in `handleGetCmd` is valueless (each just prints a pref), so a trailing term is always a user error. `isKey()` now requires the key to consume the entire remaining token, tolerating only trailing whitespace.

That is the case that misled the upstream reporter in meshcore-dev/MeshCore#2732, who typed the non-existent `get radio.fem.rxgain`, got the radio settings back, and concluded the knob was broken.

## Same silent-accept on the set side

`Utils::parseTextParts` discards anything past `max_num` without comment, so

```
set radio 910.525,62.5,7,5,foobar
```

applied the first four values and dropped `foobar` silently. It now parses one extra part and reports "too many params", checked before the range validation so the error is accurate rather than confusing.

The rest of the `set` chain is deliberately left on `memcmp` — each `set` key embeds a trailing space (`"name "`, `"af "`) so it is already self-delimiting, and the chain has a proper `"unknown config: "` fallthrough. Blanket-rejecting extra tokens there would break setters that legitimately take spaces, e.g. `set name My Node`.

## Verification

Built on `Heltec_v3_repeater` and `Heltec_v3_companion_radio_ble`.

Gemini review confirmed the whole-token rule, the `num > 4` placement, no behaviour change for valid 4-part input, and no regression for `get radio\r` (which failed before this change too). It flagged `strcpy(reply, "Error, ...")` as an overflow — false positive, `reply` is 161 bytes and the literal is 47. It also flagged the pre-existing `strcpy(tmp, &config[6])`, which **is** a real latent overflow (`tmp` is 132 bytes, a mesh-payload command can be ~179) — filed separately as #312, not fixed here.

### DIFF
--- a/src/helpers/CommonCLI.cpp
+++ b/src/helpers/CommonCLI.cpp
@@ -995,64 +995,75 @@ void CommonCLI::handleSetCmd(uint32_t sender_timestamp, char* command, char* rep
   }
 }
 
+// #299: exact-token match for CLI 'get' keys. The bare memcmp prefix compares that
+// gated this chain let any garbage suffix through -- 'get radio foobar' matched the
+// "radio" prefix and silently returned the radio settings instead of the standard
+// unknown-command error (it also let a longer key be shadowed by a shorter sibling).
+// Require the key to be followed by end-of-string or a space. strncmp (not memcmp)
+// so a config shorter than the key cannot read past its NUL terminator.
+static bool isKey(const char* config, const char* key) {
+  size_t n = strlen(key);
+  return strncmp(config, key, n) == 0 && (config[n] == 0 || config[n] == ' ');
+}
+
 void CommonCLI::handleGetCmd(uint32_t sender_timestamp, char* command, char* reply) {
   const char* config = &command[4];
-  if (memcmp(config, "dutycycle", 9) == 0) {
+  if (isKey(config, "dutycycle")) {
     float dc = 100.0f / (_prefs->airtime_factor + 1.0f);
     int dc_int = (int)dc;
     int dc_frac = (int)((dc - dc_int) * 10.0f + 0.5f);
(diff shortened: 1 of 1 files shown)


Strycher:

Issue #321, opened 2026-07-19
task: adversarial verification — Heltec V4 revision/FEM/LNA claims vs manufacturer sources


Strycher:
Parent: #320

**Task:** run a devil's-advocate verification pass over every Heltec V4 FEM/LNA claim in this repo, and produce the ground-truth research note at `docs/architecture/2026-07-19-heltec-v4-fem-lna-research.md`.

**Claims under test** (each to be confirmed / corrected / flagged unverifiable, with source):
- V4.2 carries GC1109; V4.3 carries KCT8103L.
- FEM type is reliably auto-detected by reading the default pull level on shared GPIO2 (`P_LORA_GC1109_PA_EN` / `P_LORA_KCT8103L_PA_CSD`), GC1109 pulling LOW and KCT8103L pulling HIGH (`variants/heltec_v4/LoRaFEMControl.cpp:19-41`).
- LNA is controllable **only** on KCT8103L (`setLnaCanControl(true)` is set on that branch only).
- On KCT8103L, `CTX` LOW = LNA on and HIGH = LNA bypass (`LoRaFEMControl.cpp:77-82`).
- GC1109 `CPS` HIGH = full PA, LOW = bypass; CPS is "don't care" in RX (code comments, `:63`).
- `HARDWARE.md`: high-power V4.3 = +28 ±1 dBm at antenna; KCT8103L's advantage over GC1109 is RX noise figure, not TX gain; KCT8103L small-signal gain is undocumented publicly.
- Board self-identification strings `"Heltec V4.3 OLED"` vs `"Heltec V4 OLED"` are derived purely from FEM type (`HeltecV4Board.cpp:81-83`) — i.e. the firmware equates FEM part with board revision.

**Method:** manufacturer sources first — Heltec official schematics, Heltec docs/wiki, GC1109 and KCT8103L vendor datasheets, SX1262 datasheet. Code is evidence of firmware behavior only, never of hardware fact. Gemini adversarial review before PR.

**Out of scope:** firmware changes. This task delivers the verified record; remediation lands as follow-ups under #320.

Strycher:
Agent: BrownHawk (session 8a9de97d) — Citadel `Crosswire-sw2`

Strycher:
Research note delivered — commit e3de8cb5 on `task/321-heltec-v4-fem-research`. PR to follow. Remediation items filed under #320.

PR #332, opened 2026-07-20
fix(#318): correct FEM detect comment + land CSD probe behind FEM_DEBUG_PROBE

### LINKED ISSUE #318: bug: Heltec V4 OLED reports no FEM LNA capability — instrument FEM detection to establish root cause

Strycher:
Fixes #318.

## Two things

**1. The detection comment was wrong.** It claimed the CSD level comes from chip-internal pulls. The GC1109 datasheet documents no internal pulls on CSD/CPS/CTX. It is a **board strap**:

```
V4.2 / GC1109  : R4  10k from PA_CSD to RF_GND  (pull-down) -> reads LOW
V4.3 / KCT8103L: R33 10k from PA_CSD to Vfem    (pull-up)   -> reads HIGH
```

`R33` ties to **Vfem** — downstream of the FEM LDO enabled a few lines earlier — not to 3V3. That is the physical reason a settle window can exist here at all, and why the probe samples twice. Schematic and datasheet evidence came from an independent verification pass (#321/#322) whose conclusions matched this one separately.

**2. The probe, versioned and gated.** An earlier throwaway build carried this instrumentation, established the answer, and was reverted **without being committed** — leaving an unversioned binary on a freely-reflashable device as the only artifact. Another session caught that. This lands it properly, behind `-D FEM_DEBUG_PROBE`, **off by default**.

## Default build is unaffected

No second read, no delay, no print. The only always-on change is hoisting the existing `digitalRead` into `csd_early`, which is the value the decision already used.

```
-D FEM_DEBUG_PROBE                              -> 60 lines, ~60 s boot delay
-D FEM_DEBUG_PROBE -D FEM_DEBUG_PROBE_REPEAT=1  -> single line, no delay
```

The repeat exists because on ESP32-S3 native USB the port drops and re-enumerates across a reset — a host reader cannot hold the port through boot and can only re-attach afterwards, by which time a single print is long gone. At ~1 Hz it will drown CLI replies on the same endpoint, which is exactly why it is opt-in.

## Gemini review caught a real defect

`Serial.flush()` blocks when no host is attached to the USB CDC endpoint, which can trip the watchdog. Transmission is now guarded on `if (Serial)`. The `delay()` is deliberately left **outside** that guard — the reader attaches late by design, so gating the delay too would burn every iteration instantly and print nothing to a host that shows up at t+10 s.

## What it established

ST-P: `csd_early=0 csd_late=0 fem_type=GC1109(V4.2) lna_can_control=0`. LOW at 1 ms and still LOW at 50 ms, so the settle-time hypothesis is refuted and the classification is correct. GC1109 has no independent LNA bypass (datasheet Table 4: RX = `CSD=1, CTX=0, CPS=X`), so hiding the control there is right.

Built: `heltec_v4_companion_radio_ble` with and without the flag, and `heltec_v4_repeater`.

### DIFF
--- a/variants/heltec_v4/LoRaFEMControl.cpp
+++ b/variants/heltec_v4/LoRaFEMControl.cpp
@@ -16,13 +16,25 @@ void LoRaFEMControl::init(void)
         delay(1);  // FEM startup time after cold power-on
     }
 
-    // Auto-detect FEM type via shared GPIO2 default pull level.
-    // GC1109 CSD: internal pull-down → reads LOW
-    // KCT8103L CSD: internal pull-up → reads HIGH
+    // Auto-detect FEM type from the CSD level on the shared GPIO.
+    //
+    // #318: this is a BOARD STRAP, not a chip-internal pull. The previous comment
+    // here claimed "internal pull-down / internal pull-up"; that is wrong -- the
+    // GC1109 datasheet documents no internal pulls on CSD/CPS/CTX. From the Heltec
+    // schematics:
+    //   V4.2 / GC1109  : R4  10k from PA_CSD to RF_GND  (pull-down) -> reads LOW
+    //   V4.3 / KCT8103L: R33 10k from PA_CSD to Vfem    (pull-up)   -> reads HIGH
+    //
+    // R33 pulls to Vfem -- downstream of the FEM LDO enabled a few lines above --
+    // NOT to 3V3. That is why a settle window can exist at all here. Measured on a
+    // GC1109 unit: LOW at 1 ms and still LOW after 50 ms, so the strap is
+    // unambiguous there. Build with -D FEM_DEBUG_PROBE to re-measure on a board
+    // whose FEM is unknown or suspect.
     rtc_gpio_hold_dis((gpio_num_t)P_LORA_KCT8103L_PA_CSD);
     pinMode(P_LORA_KCT8103L_PA_CSD, INPUT);
     delay(1);
-    if(digitalRead(P_LORA_KCT8103L_PA_CSD)==HIGH) {
+    csd_early = digitalRead(P_LORA_KCT8103L_PA_CSD);   // the level the decision uses
+    if(csd_early==HIGH) {
         // FEM is KCT8103L (V4.3)
         fem_type= KCT8103L_PA;
         pinMode(P_LORA_KCT8103L_PA_CSD, OUTPUT);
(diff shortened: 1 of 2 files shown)


Strycher:

Issue #320, opened 2026-07-19
Epic: Establish verified ground truth — Heltec V4 revisions, FEM identification, LNA controllability


Strycher:
Parent: #319

**Cause:** Offband firmware, `HARDWARE.md`, and issue #298 all assert a specific Heltec V4 FEM story (V4.2 = GC1109, V4.3 = KCT8103L, CSD-pull auto-detect, LNA controllable only on KCT8103L). That story originates from inherited upstream code comments and prior-session reasoning — **not from manufacturer documentation.** It has never been checked against a Heltec schematic or a FEM vendor datasheet, and at least one prior session produced conclusions the owner does not trust.

**Ask:** produce a verified ground-truth record covering:
1. **Revision map** — which Heltec V4 hardware revisions exist, which FEM part each carries, and which are actually in circulation.
2. **Identification** — how to tell revisions apart visually (silkscreen, component markings/package) and by firmware probing; whether the existing GPIO2 pull-level probe is a sound discriminator or a coincidence.
3. **Control surface** — what each FEM part actually exposes (chip enable, TX/RX path select, LNA bypass, gain steps) per its datasheet, and therefore which knobs can legitimately be offered as a user preference.
4. **Corrections** — every in-repo claim confirmed, corrected, or flagged unverifiable, with the source cited.

**Evidence standard:** every claim carries `[verified: <source>]` or `[hypothesis: untested]`. Manufacturer schematic / vendor datasheet outranks code comments, wiki prose, and prior issues. Where no public datasheet exists, that is itself the finding — say so rather than inferring.

**Downstream:** #298 (companion never applies `radio_fem_rxgain`) is in scope of this Epic and should be re-adjudicated against the verified record, not merged on its own reasoning.

Strycher:
Closing as confirmed resolved — owner sign-off received 2026-07-19.

**Delivered:** verified ground-truth record at `docs/architecture/2026-07-19-heltec-v4-fem-lna-research.md`, merged via #322.

**Established against manufacturer sources:**
- FEM parts confirmed from Heltec schematics — GC1109 (U10) on V4.2, KCT8103L (U8) on V4.3; pin-compatible except pin 3.
- GC1109 has **no LNA bypass** (datasheet Table 4) — its RX path is permanently through the LNA. "LNA controllable only on KCT8103L" is correct, but because the older part is always in its best receive state, not because it is deficient.
[...]

PR #335, opened 2026-07-20
task(#327): report the FEM part in the model string; audit GPIO46 strapping

### LINKED ISSUE #327: task: Heltec V4 FEM code remediation — detect comment, board-name strings, GPIO46 audit
Remediation items from the #321 research note (§10) that were **not** actioned by that docs-only work. Filing them so they stay on the board now that Epic #320 is closed.

**1. Correct the FEM auto-detect comment** — `variants/heltec_v4/LoRaFEMControl.cpp:19-21` attributes detection to `GC1109 CSD: internal pull-down` / `KCT8103L CSD: internal pull-up`. **Wrong.** The GC1109 datasheet documents no internal pulls on CSD/CPS/CTX; the real mechanism is a board-level revision strap — V4.2 `R4` 10 kΩ → RF_GND, V4.3 `R33` 10 kΩ → **Vfem**. Anyone porting this probe to another board off the current comment will get it wrong. Name R4/R33 instead.

**2. Reconcile the board-name strings** — `HeltecV4Board.cpp:81-83` reports `"Heltec V4.3 OLED"` / `"Heltec V4.3 TFT"`, derived purely from FEM type. Heltec's published revisions are **V4.0** and **V4.3.1**; "V4.3" is a schematic filename, not a documented board revision. Either align with Heltec's numbering or stop reporting a revision the vendor does not publish (reporting the FEM part would be honest and more useful).

**3. Audit GPIO46 handling** — `P_LORA_GC1109_PA_TX_EN=46` is driven `OUTPUT` from `init()` and held. **GPIO46 is an ESP32-S3 strapping pin.** Upstream raised this in meshcore-dev/MeshCore#1249 (proposed: leave unconfigured at boot, set OUTPUT only around TX, release to INPUT after); that PR was closed and superseded by #1600, which retained only the deep-sleep `PA_POWER` hold. Confirm whether the strapping concern is still live in our tree and whether it matters in practice. Note this pin is GC1109-only — V4.3 freed GPIO46 and moved FEM control to GPIO5.

**Deliberately NOT included:**
- RF characterisation of the KCT8103L LNA (noise-floor A/B). Downgraded from load-bearing to optional in #321 §8a once the control was validated end-to-end on Firestar. Worth doing only if someone wants actual dB figures.
- `HARDWARE.md` corrections — owned by RedCreek, gitignored per-host file, not committable here.

**Context:** `docs/architecture/2026-07-19-heltec-v4-fem-lna-research.md` §10. Related: #318, #324, #298, #323.

Strycher:
Closes #327 · Epic context #320 · Agent: BrownHawk (session 8a9de97d) · Citadel `Crosswire-e2f`

Remediation items from the #321 research note. **Item 1 was already landed by RedCreek in `6b341951`** — only items 2 and 3 remained.

## User-visible change (owner-approved wording)

`getManufacturerName()` — the **Settings → Model** field in the client:

| | before | after |
|---|---|---|
| KCT8103L | `Heltec V4.3 OLED` | `Heltec V4 OLED (KCT8103L)` |
| GC1109 | `Heltec V4 OLED` | `Heltec V4 OLED (GC1109)` |

TFT equivalents likewise. Both old arms were wrong: **"V4.3" is a Heltec schematic filename, not a revision Heltec publishes** (their hardware log lists V4.0 and V4.3.1), and bare "V4" understated a GC1109 board whose silkscreen reads `HTIT-WB32LAF V4.2`. The detect reads the FEM strap, so the FEM part is the only thing this function actually knows — and it is what determines whether the LNA setting appears.

## GPIO46 — audited, not a defect

Comment only, no behaviour change. Verified against **ESP32-S3 Series Datasheet v2.2 §3**: strapping pins are explicitly *"freed up to be used as regular IO pins after reset"*, GPIO46 defaults to a weak pull-down (Table 3-1), and GPIO0=1 selects SPI boot for **any** GPIO46 value (Table 3-3). The pin is never `rtc_gpio_hold_en`-ed, so it cannot be latched HIGH through a reset. Upstream #1249's release-to-INPUT proposal is deliberately not adopted — closed and superseded by #1600, and it guards a hazard the reset behaviour already rules out.

## Unplanned safety fix

`sprintf` → `snprintf` on the OTA page ID buffer (`ESP32Board.cpp`). `id` is `node_name[32]`; with the longer model string the worst case lands on **exactly** the 60-byte capacity — zero margin on a previously unbounded copy. Bounded so future growth truncates rather than corrupts.

## Gemini review — 2 BLOCKERs, both upheld

1. **The GPIO46 safety argument was right for the wrong reason** — it credited `gpio_pulldown_en()` (a deep-sleep measure) for reset behaviour it has nothing to do with. Rewritten to name the actual mechanism: the output driver is disabled before the strapping latches sample. A reviewer "cleaning up" the sleep path would have been misled by the original.
2. **Downstream model-string consumers were unexamined.** Audited all four — MQTT status JSON (sanitized *display* field, not part of any topic or unique_id, so HA device identity is unaffected), device-info frame (25 chars into 40), OTA page ID, CLI reply. **Nothing parses it for identity.** Chasing this is what surfaced the unbounded sprintf.

Both MINORs (tone) also applied.

## Verification

`heltec_v4_repeater` + `heltec_v4_tft_repeater` — SUCCESS, covering both `#ifdef` arms.

Not hardware-tested. The model string is display-only and the GPIO46 change is a comment; the `snprintf` is the only behavioural delta and it is strictly narrowing. Client-side display lives in a separate repo — flagging for the client agent.

Requires human review and merge approval.


Strycher:
Human approval received from owner ("merge #335"). Enabling auto-merge per CLAUDE-BASE — agent does not merge manually.

Issue #357, opened 2026-07-23
Task: gated debug CLI (hangtest + resetreason) for watchdog/heartbeat bench verification


Strycher:
Task under Epic #255 / Feature #253. Instrumentation that produced the hardware evidence closing #257, #266 and #275.

## Why

The #257/#266 acceptance test ("induced hang -> auto-reboot -> RESETREAS=Watchdog") was **not executable** with the firmware as shipped:

- No way to induce a main-loop hang on demand.
- No way to read the reset reason after the fact. There is **no SWD/J-Link probe on this bench (and none wanted)**, so no RTT; the boot banner prints before a USB-CDC host can attach; and no CLI command reported it.

## What

Two DEBUG-ONLY commands in `CommonCLI::handleCommand`, gated behind `-D OFFBAND_WDT_HANGTEST=1`:

- **`hangtest`** — spins the main loop forever (`nop` prevents the compiler eliding a side-effect-free infinite loop) so the loop-fed watchdog starves. Proves both #257 (WDT trips) and #275 (loop-driven LED freezes) in one shot.
- **`resetreason`** — reports `getResetReason()` / `getShutdownReason()` via the `mesh::MainBoard` virtuals (`src/MeshCore.h:103-106`), which default to `"Not available"`, so it is portable rather than nRF52-only.

## Gate — proven, not asserted

`strings firmware.elf | grep hangtest` -> **ABSENT** in a normal build, **PRESENT** only with the flag. It cannot ship by accident.

## Evidence produced

```
> hangtest
hangtest: wedging main loop on purpose -- expect LED freeze then Watchdog reset (~30s)
> resetreason
  -> Last reset: Watchdog (0x2)
```
Control read immediately post-DFU: `Soft Reset (0x4)` — the command reports the real reason, not a constant.

## Follow-on

**#351** should promote/generalize `resetreason` into a persistent, permissioned, mesh-queryable form rather than starting from scratch. This gated command is bench-only: values live in RAM for the current boot and there is no permissioning.

Related: #255, #257, #266, #275, #351

Strycher:
## Closing — landed on `feat/275-heartbeat` (`083d639`)

Gated debug CLI committed. Gate proven empirically: `strings firmware.elf | grep hangtest` is **ABSENT** in a normal build, **PRESENT** only with `-D OFFBAND_WDT_HANGTEST=1`.

[...]

Issue #350, opened 2026-07-23
Epic: decouple CrashLog from wifi_observer — uniform boot-survival diagnostics across roles


Strycher:
## Problem

CrashLog lives at `src/helpers/wifi_observer/CrashLog.cpp` — structurally coupled to the WiFi/observer path. Every role that is **not** a WiFi observer therefore cannot use boot-survival diagnostics: nRF52 repeaters, room server, and (newly relevant) **sensor leaf nodes**.

This matters most exactly where it is hardest to debug: an unattended node in a field or on a pole has **no serial port and no operator**. A persistent "why did I reboot / where did I hang" record is the only diagnosis channel.

## Goal

Move CrashLog to a role-neutral home and make boot-survival diagnostics **uniform across all roles and both MCU families** (ESP32-S3 + nRF52), with one API and one boot-banner surface.

## Scope (to groom)

- **Relocate + de-WiFi** the module out of `helpers/wifi_observer/` to a role-neutral path; strip observer/WiFi assumptions.
- **Abstract the persistence backend** — ESP32 `RTC_NOINIT` vs nRF52 no-init RAM / `GPREGRET` / UICR.
  ⚠ **SafeBoot already owns retention state** (ESP32 `RTC_NOINIT_ATTR`; nRF52 `GPREGRET2` with a bit-7 ownership marker). The abstraction must **not collide** with SafeBoot's region.
- **Uniform reset-reason decoding** — `ESP32Board.h` and `NRF52Board.cpp` already decode; unify the surface.
- **Enable across envs** — sensor / repeater / room server builds, ESP32 + nRF52; CI green on the matrix.

## Coordination (important)

**#255 deliverable 2** ("nRF52-backed persistent crash breadcrumb") overlaps this directly. This Epic should provide the **shared store + abstraction** that #255's breadcrumb builds on — **not** a parallel implementation. Sequence and reconcile with #255/#257 before either lands.

## Related
- #349 (Sensor build — a primary consumer)
- #255 / #257 (nRF52 hang auto-recovery + breadcrumb)
- #276 (CrashLog lift into wadamesh HV4-TFT diag build)

Parentless by owner approval; may later sit under a reliability/diagnostics Feature.

Strycher:
Agent: RubyDog (session b3d7203d)

Starting this Epic — it is the prerequisite for #255 deliverable 2 (owner-sequenced: #350 first).

**Design decisions locked by owner:**
[...]

Issue #266, opened 2026-07-02
Task: wire the #257 watchdog into the repeater (simple_repeater, nRF52-guarded)


Strycher:
Task under Epic #255 / Feature #253. Extends the #257 nRF52 watchdog (companion) to the repeater.

The watchdog engine (`NRF52Board::startWatchdog()`/`feedWatchdog()`) already exists and is hardware-verified on the companion (#257). The **repeater example never calls it** (`simple_repeater/main.cpp`: grep `startWatchdog|feedWatchdog` = 0), so a repeater build carries the code dormant. This wires it in — important because a fixed/unattended repeater is exactly where auto-recovery matters most.

### Change (mirror the companion, `#if defined(NRF52_PLATFORM)`-guarded)
- `examples/simple_repeater/main.cpp`: `startWatchdog(30)` after boot/init; `feedWatchdog()` once per `loop()` pass (feed from the loop ONLY).
- Optional: unconditional `[boot] last reset: <reason>` banner (repeater has a readable USB console, so cheap + useful).

### Guard rationale (avoids a false-reset footgun)
`simple_repeater` is shared with the **ESP32 WiFi/MQTT-bridge** repeaters, which can block >30 s in a network call (would false-trip a WDT). The `NRF52_PLATFORM` guard excludes them — only nRF52 LoRa repeaters (RAK3401 1W, RAK4631, t096) arm the watchdog, and those do no blocking network I/O. Confirm no >30 s blocking op in the nRF52 repeater loop when placing the feed.

### Falsifiable bench test (same as #257, easier here — repeater has a text USB console)
Induced hang (temp `while(1){}`) on an nRF52 repeater → auto-reboot ≤30 s → boot banner reads `Watchdog` (RESETREAS=0x2). Plus a no-false-reset idle soak.

Build (nRF52 repeater envs) + Gemini review + bench verify before it's called done. Tier-2 build/flash gated.

Strycher:
## Closing — hardware-verified (owner-authorized 2026-07-23)

The falsifiable bench test in this issue was executed on an nRF52 repeater.

**Test:** induced hang on `wismesh1w-repeater` (RAK10724 WisMesh 1W, RAK3401 nRF52840 core, COM20) running the `RAK_3401_repeater` build with the wiring from `58b6fb11`:
```
> hangtest                       # gated debug cmd (-D OFFBAND_WDT_HANGTEST=1)
[...]

Issue #257, opened 2026-06-30
Task: nRF52 hardware watchdog (Board-layer, auto-recover + Watchdog reset-reason) (#255)


Strycher:
Task under Epic #255 / Feature #253.

Add an nRF52 hardware watchdog so a main-loop hang auto-reboots (service restored without physical access) and the reboot self-stamps `RESETREAS = Watchdog` (already decoded by `NRF52Board::getResetReasonString`).

### Exact changes (cited)
- `src/helpers/NRF52Board.{h,cpp}`: add `startWatchdog(uint32_t timeout_s)` (nrf_wdt HAL: CONFIG run-in-sleep + pause-under-debugger, CRV=timeout×32768, enable RR0, TASKS_START) + `feedWatchdog()` (reload RR0). Design as Board-layer so the rest of the nRF52 fleet inherits it.
- `variants/rak3401/RAK3401Board.cpp` (~`:62`, after `checkBootVoltage`): call `startWatchdog(30)`.
- `examples/companion_radio/main.cpp` `loop()`: feed once per pass on nRF52 **not** behind `OFFBAND_OBSERVER`; feed inside `NRF52Board::sleep()` before WFE (`:490-494`).
- `src/helpers/NRF52Board.cpp:60-65`: print the reset-reason boot line **unconditionally** (not MESH_DEBUG-only) so "Watchdog"/"CPU Lockup" shows on the companion boot banner.

### Design rule
Feed **only from the main loop** (a timer-ISR feed would mask a hung loop). Timeout (30 s, tunable) must exceed the longest legitimate loop gap during idle — **soak-test, do not assume.**

### Falsifiable bench tests (HW test — not native-unit-testable)
1. Induce hang (temp forced `while(1){}`) → expect auto-reboot ≤30 s + next boot banner reset reason = **"Watchdog"**.
2. No-false-reset soak: idle 10+ min, then BLE connect + 310-contact sync + settings refresh → expect uptime keeps climbing, no reboot.

Build (`pio run`) + flash are Tier-2 — gated on explicit human approval; flash only after confirming chip serial `2B1FE7EF20987013`.

---

### ⚠ Correction — bench target chip serial (2026-07-23, RubyDog)

[...]

PR #360, opened 2026-07-24
fix(#275/#257/#266): true ungated nRF52 heartbeat + hardware watchdog (repeater + companion)

### LINKED ISSUE #275: bug(P0): green (user) LED heartbeat is gated/suppressed — must be an absolute, ungated liveness indicator
Origin: #7 (RAK4631 companion heartbeat), #9 (repeater heartbeat). Suppressing mechanism: #208 (nRF52 power-save nap). Under Feature #253.

## REQUIREMENT (non-negotiable)
The green (user) LED heartbeat blink **must be an absolute, true indicator that the heartbeat / main loop is still running.**
- It **must NOT be gated behind any other pre-condition** — not `UITask`/any UI, not display state/presence, not connection state, not the power-save nap (`hasPendingWork`/`board.sleep`), not traffic.
- It **MAY also** indicate traffic (activity blink is allowed on top), **but it can NEVER FAIL to indicate the heartbeat.**
- Applies to every headless nRF52 role (companion + repeater) — where the LED is the *only* liveness signal, which is exactly why it must never lie.

## Defect (root cause, verified)
- **Companion:** heartbeat is driven by `UITask::userLedHandler()` (per #7), which only runs on a full main-loop pass. The nRF52 power-save nap (`if (!hasPendingWork()) board.sleep(0)`, 1.16.0 base / #208) suppresses loop passes when idle → the "heartbeat" degrades to an **activity light**: it blinks only when traffic/an event wakes the loop, and goes dark when the device is idle-but-alive.
- **Repeater:** the loop-direct blink (#9) has the **same** nap-suppression — #9's own code comment: "Cadence follows the loop rate; under nRF52 powersave it blips on each wake."
- Net: an idle-but-running device shows **no blink** and is indistinguishable from dead.

## Impact — why P0
The heartbeat is the primary liveness diagnostic. Its silent failure caused **running devices to be judged dead** (repeated field pulls of a live unit) and drove an entire wedge/watchdog investigation on a false signal. A diagnostic instrument that silently fails to indicate liveness is a P0 defect.

## Fix direction
A fixed-cadence (~1 Hz) heartbeat **independent of UI, display, connection, and the nap** — e.g. a periodic RTC/timer tick that toggles the LED regardless of loop-idle state (and, tying to #255, feeds the WDT on that same cadence). The blink must continue when the device is fully idle; it must stop **only** on a genuine loop/tick hang.

## Acceptance
- Idle nRF52 device (no traffic, no client connected) blinks the green LED at the fixed cadence.
- A genuine loop/tick hang stops the blink (and, with #255, trips the WDT).
- Verified on headless nRF52 companion AND repeater.

Strycher:
Agent: RubyDog (session b3d7203d)

Recovers work that was stranded local-only on an unpushed branch, finishes it, and verifies it on hardware.

## What this does

**#275 — the green LED lied about liveness (P0).** The heartbeat ran through `UITask::userLedHandler()` and was suppressed by the nRF52 power-save nap (#208), so an idle-but-running node went dark and was indistinguishable from dead. That false signal caused live units to be pulled from the field.

Fix: `NRF52Board::startHeartbeat()` / `heartbeatTick()` toggle the LED **from the main loop** on a `millis()` schedule; a repeating `SoftwareTimer` with a **deliberately empty callback** pulls the loop out of `sd_app_evt_wait()` at ~10 Hz. The wake is the point — the loop, not the ISR, owns the LED and the watchdog feed, so a genuine hang still freezes the blink. ~1 Hz low-duty blip (~90 ms on / ~890 ms off ⇒ ~10% duty) because the car companion is a solar node. `UITask` no longer writes `PIN_STATUS_LED` on nRF52 (two writers would contend and silently kill the heartbeat).

**#257 — nRF52 hardware watchdog.** `startWatchdog(30)` + `feedWatchdog()` fed **from the main loop only** (an ISR feed would mask a hung loop). A field hang now auto-recovers instead of wedging until someone drives to it.

**#266 — watchdog wired into `simple_repeater`**, `NRF52_PLATFORM`-guarded so ESP32 WiFi/MQTT-bridge repeaters (which can legitimately block >30 s) can't false-trip.

**#357 — gated debug CLI** (`hangtest`, `resetreason`) behind `-D OFFBAND_WDT_HANGTEST=1`. Without it the acceptance test was not executable: no probe on the bench (no RTT), and the boot banner prints before USB-CDC can attach.

## Hardware verification

| Criterion | Result |
|---|---|
| Companion idle blink (`wismesh1w-companion`) | ✅ steady ~1 Hz |
| Repeater idle blink (`wismesh1w-repeater`) | ✅ matches companion |
| Hang → LED freezes | ✅ observed by owner |
| Hang → unattended reboot | ✅ device re-enumerated |
| Reset reason | ✅ `Last reset: Watchdog (0x2)` |
| Control (post-DFU) | ✅ `Soft Reset (0x4)` — reason is read, not constant |
| No-false-reset soak | ✅ owner confirms LED continuity |

## Builds
`RAK_3401_repeater`, `RAK_3401_companion_radio_usb`, `RAK_4631_repeater`, `RAK_4631_companion_radio_ble` — all SUCCESS.

## Gemini adversarial review
Run on each increment (`8990857d`, `7f58fb56`, `2ec32720`). #275 findings applied: **CRITICAL** (file-static `SoftwareTimer` → board member), **MAJOR** (~50% duty → ~10%, solar node).

## Debug-command safety
`strings firmware.elf | grep hangtest` → **ABSENT** in a normal build, **PRESENT** only with the flag. Proven empirically, not asserted.

## Scope
Closes #275, #257, #266, #357 (all already closed with evidence).
**Epic #255 stays OPEN** — deliverable 2 (persistent crash breadcrumb store) is not implemented; it should consume #350's shared store rather than contend for the 8-bit `GPREGRET2` that SafeBoot already partly owns.


Strycher:
Human approval received from owner in chat (2026-07-24): "Merge #360". Enabling auto-merge (`--auto --rebase`) per CLAUDE-BASE — agent does not force-merge.

⚠ Note for the record: `firmware-base` has no required status checks, so `--auto` lands **immediately** rather than waiting on CI. This merge is gated by the hardware verification recorded on #275/#257/#266, not by CI.

Issue #366, opened 2026-07-24
task: detect config-provider key-space overlap instead of silently shadowing (first-provider-wins hazard)


Strycher:
Surfaced by the Gemini adversarial review of #364 (**MAJOR-2**). Review log: `docs/llm-consultations/2026-07-24-364-config-dispatch-review-gemini-gemini-2.5-pro.log`.

## The hazard

`offband::config::dispatchSet`/`dispatchGet` (added in #364) short-circuit on the **first provider that claims a key**. A provider registered earlier therefore **silently shadows** a later one for any key both would claim.

This is live, not theoretical, for **#301**: the observer provider claims the **entire `wifi.` prefix** (`wifi.ssid`, `wifi.pwd`) via

```cpp
if (strncmp(key, "wifi.", 5) == 0) return handleSetWifiField(...);
```

so a repeater provider registered after the observer will **never be asked** about those keys. On a combined repeater+observer build (#297) the repeater's WiFi config would be unreachable over the wire, with **no diagnostic** — it would simply behave as though the repeater's handler didn't exist.

## What #364 already did

Documented the constraint prominently in `src/helpers/config/ConfigDispatch.h` ("KEY SPACES MUST BE DISJOINT ACROSS ROLES — FIRST PROVIDER WINS") and made it an explicit manual review obligation when adding a role. That is mitigation by documentation only — it relies on a human remembering.

## Ask

Make the overlap **mechanically detectable** rather than a review obligation. Options (pick during design, not prescribed here):

- A debug/build-flag-gated `dispatchAudit()` that walks **all** providers without short-circuiting and logs when more than one claims the same key.
- Provider-declared key/prefix manifests checked for overlap at registration time (costs RAM — weigh against `kMaxProviders` and tight-RAM boards).
- A host-side unit test that asserts the registered key spaces are disjoint (zero firmware cost; needs the key sets to be enumerable).

## Acceptance criteria

- [ ] An overlap between two registered providers is surfaced automatically (build failure, test failure, or a loud runtime diagnostic under a debug flag) — not silently shadowed.
- [ ] Zero flash/RAM cost in release builds, or a cost quantified and accepted.
- [ ] Covers the concrete `wifi.` case above as a regression test.

## Sequencing

**Should land with or before #301**, which is where the second provider actually appears and where the trap would first bite. Not a blocker for #364 (single provider registered; no overlap possible yet).

Strycher:
[...]

Issue #365, opened 2026-07-24
Companion WiFi credentials at runtime: capability bit 0x08 + a non-observer config path


Strycher:
Companion firmware is gaining runtime WiFi credential config (owner, 2026-07-18). This is the firmware half; the app half is OffbandMesh/meshcore-client#375 and is **blocked on the contract decided here**.

The app already sets WiFi on Observer. The blocker is that today's gating and the config backend itself are both observer-only, so Companion cannot reuse either as-is.

## 1. Capability bit — proposed `0x08`

Current allocation [verified: `examples/companion_radio/OffbandConfigProtocol.h:113-115`]:

| Bit | Value | Meaning |
|---|---|---|
| 0 | `0x01` | `OFFBAND_CAP_WIFI_OBSERVER` — config backend compiled in |
| 1 | `0x02` | `OFFBAND_CAP_BLOCK` — user block list (#241) |
| 2 | `0x04` | `OFFBAND_CAP_FEM_LNA` — FEM LNA runtime control (#298) |
| 3 | **`0x08`** | **proposed `OFFBAND_CAP_WIFI_COMPANION`** |

**No upstream conflict.** `offband_caps` does not exist upstream [verified: `git grep OFFBAND_CAP` on `upstream/main` + `upstream/dev` → no matches]. It is a fork-only byte appended to `RESP_CODE_DEVICE_INFO` after `path_hash_mode`, additive so pre-v14 clients read a shorter frame and ignore it. Bits 3..7 are ours; four remain free after this one.

**Do not reuse `0x01`.** It gates the *whole* Observer settings category in the app — WiFi, MQTT brokers, and display together. A Companion advertising `0x01` would show broker/display config that does not apply to it.

## 2. The real constraint — the config backend is observer-only

Not just a gating problem. The backend (`ConfigSchema` + `dispatchObserverCli`) compiles **only under `OFFBAND_OBSERVER`**, and the header states there is *"NO display-without-observer path"* — display prefs persist via `ConfigSchema`, itself observer-only [verified: `OffbandConfigProtocol.h:104-115`; `#ifdef OFFBAND_OBSERVER` blocks throughout `MyMesh.cpp`].

**So a non-observer Companion build has no config subsystem at all.** Advertising a bit is not enough — Companion needs a config path that exists without `OFFBAND_OBSERVER`. That is the bulk of the work here.

[...]

Issue #305, opened 2026-07-19
Epic: GUI config — extend the companion-API contract with repeater keys + Offband client screens


Strycher:
Expose the repeater's config (WiFi, broker pool slots, OTA secret, command-queue endpoint/secret) over the **#143 companion-API config contract** so an operator configures everything from the **Offband client GUI — no CLI required**.

**Two-repo work** (firmware + meshcore-client): additive, capability-gated command codes so stock clients are unaffected; version/capability negotiation; secrets **write-only** in GET responses (mirror the observer redaction pattern). Coordinate with #142 (client settings UI) and #143 (contract).

**Acceptance:** a repeater is fully configurable from the app with zero `_sys` CLI commands; older clients degrade gracefully.

---
**Parent Feature:** #295 · **Scoped by:** #296 · **Design-of-record:** `docs/architecture/2026-07-17-repeater-telemetry-genericization-dor.md`

Strycher:
Cross-link: capability-bit requirement for the repeater config surface recorded on #301 (a repeater-config cap bit + client gating; the current model treats config as observer-only via OFFBAND_CAP_WIFI_OBSERVER 0x01). The client screens + contract bits land here in #305. Coordinate bit allocation with #365 (claims 0x08). Agent: CloudyCliff (session 83c8582b)

Issue #301, opened 2026-07-19
Epic: Repeater publish — separate BurstMqttPublisher + wifi.mode selector (observer pool untouched)


Strycher:
Converge the repeater telemetry publish path off `PubSubClient` onto the observer `esp_mqtt` `MqttBrokerPool`+`ConfigSchema` (DoR Decision D1 = Option A), and un-bake WiFi/MQTT config to runtime.

**Dual lifecycle:**
- **Persistent** — powered / observer / combined nodes: pool runs as today (must NOT regress).
- **Burst** — solar repeaters: begin→publish→drain→end with a bounded connect window; accept per-wake TLS handshake cost, or fall back to a TCP broker on the tightest solar budgets.

Unlocks TLS/wss/JWT brokers from a repeater (the driver for the combined repeater+observer role, #297).

**Acceptance:** repeater publishes via the pool over TCP *and* TLS/wss; observer regression-free; solar burst power budget re-validated.

---
**Parent Feature:** #295 · **Scoped by:** #296 · **Design-of-record:** `docs/architecture/2026-07-17-repeater-telemetry-genericization-dor.md`

Strycher:
**⚠ Approach REVISED by adversarial review (MAJOR 3) — issue retitled.**

**Original (rejected):** retrofit a dual burst/persistent lifecycle into `MqttBrokerPool`. Rejected for state-machine conflicts (teardown during backoff/reconnect), event races (`MQTT_EVENT_DISCONNECTED` mid-teardown → double-free / use-after-free), an ill-defined 'drain' silently dropping async publishes, and regression risk to **deployed observers**.

**Approved approach:** leave `MqttBrokerPool` **untouched**. Build a separate, small **`BurstMqttPublisher`** — still on `esp_mqtt_client` (TLS remains available), simple synchronous `connect → wait → publish → wait-for-ack → disconnect`. Both read the **same `ConfigSchema`**, so config + GUI are identical either way.

**Adds `wifi.mode` (owner-approved):**
```
wifi.mode = burst   → wake, publish, sleep   (default; solar)  → BurstMqttPublisher
wifi.mode = always  → stay connected         (PoE/powered)     → MqttBrokerPool
```
Extensible mode (not a boolean) so it renders as a GUI dropdown. Existing temporary overrides still apply — an OTA/keepalive command can hold a burst node online for a session, then it drops back.

[...]

PR #368, opened 2026-07-24
refactor(#364): extract role-agnostic config-command dispatch into src/helpers/config/

### LINKED ISSUE #364: task: extract role-agnostic config-command dispatch into src/helpers/config/ (Epic 300 item 1)
Child task of Epic #300 (Feature #295), **decomposition item 1**. Design-of-record: `docs/architecture/2026-07-17-repeater-telemetry-genericization-dor.md` §5 / D2.

## Goal

Extract the config-command **dispatch mechanism** out of `ObserverCli` into a role-agnostic component under `src/helpers/config/`, so #301/#305 can register repeater keys (`wifi.mode`, creds, cmd-queue, OTA) against the same surface. `ObserverCli` keeps every observer-specific handler; it registers them and routes through the shared dispatcher.

**This is a pure internal refactor. NO behavior change.**

## Current state `[verified: read this session]`

- `src/helpers/wifi_observer/ObserverCli.cpp:975-1031` — `configSet(key, value, reply, reply_size, MqttBrokerPool&)`
- `src/helpers/wifi_observer/ObserverCli.cpp:1033-1080` — `configGet(key, reply, reply_size)`
- Parse helpers already factored: `parseConfigBool` (:951), `parseBrokerKey` (:963), `parseSlot` (:42), `eq` (:25)
- Enumeration API (`configBrokerSlotCount` / `configBrokerSlotPopulated` / `configRenderBrokerSlot`, :1085-1151) is **broker-specific** — stays observer-side, out of scope.
- Sole wire caller: `examples/companion_radio/MyMesh.cpp:1278` (`OCFG_SET`) / `:1287` (`OCFG_GET`), via `CMD_OFFBAND_CONFIG`.
- `helpers/wifi_observer/*.cpp` is compiled by **exactly 4 envs** (`build_src_filter`): `Heltec_v3_companion_observer_wifi`, `heltec_v4_companion_observer_wifi`, `heltec_v4_tft_companion_observer_wifi`, `Xiao_S3_WIO_companion_observer_wifi`. nRF52 envs never compile it.

## Coupling to break

`configSet`/`configGet` hardwire observer keys (`mqtt.*`, `display.*`, `wifi.*`) and take `MqttBrokerPool&` directly. The **dispatch mechanism** must become role-agnostic; the **handlers** stay observer-owned.

## Hard constraints

1. **Wire contract (#143/#160) stays BYTE-COMPATIBLE.** The Offband client must see zero change. This includes the `reply` **text**: every `snprintf` format string (incl. every `ERROR: …` string) is observable by the client and must be preserved character-for-character. Return-value semantics (`false` = unknown key → `OCFG_R_ERR "unknown config key"`) unchanged.
2. **Key evaluation order preserved exactly.** `wifi.enabled` before the generic `wifi.` prefix; `mqtt.iata` / `mqtt.status_interval` before `mqtt.broker.`; the malformed-broker-key error branch after the well-formed parse. Reordering changes behavior.
3. **`enabled` / `clear` stay ACTIONS, not fields** (F6 / #166) — `enabled` written last as the activation guard; `clear` wipes a slot.
4. **Platform guards** — any platform API in the shared component guarded with `#if defined(ARDUINO) && defined(ESP_PLATFORM)` (convention: `src/helpers/wifi_observer/MqttBroker.h:15`).
5. **No heap** in the registration mechanism — fixed static storage (tight-RAM boards; matches the raw-function-pointer precedent in `ObserverCli.h`).
6. **No silent registration failure** — if a role fails to register, that must be loud, not a silently-dead config surface.

## Acceptance criteria

- [ ] `src/helpers/config/` exists with the role-agnostic dispatch surface + shared parse helpers + a no-heap registration mechanism.
- [ ] `ObserverCli` registers its observer keys and routes through the shared dispatcher; all handlers still observer-owned.
- [ ] Reply strings and key ordering byte-identical to pre-refactor (diff-audited, documented in the PR).
- [ ] All 4 observer envs build clean.
- [ ] One **nRF52** env builds clean (proves guards + proves the new directory did not leak into non-ESP32 builds).
- [ ] Gemini adversarial review run on the branch diff; every finding fixed or justified in the PR body.

## Out of scope

- Repeater key registration (#301 / #305).
- Broker-slot enumeration API (stays observer).
- The **observer hardware no-regression gate** — decomposition item 3, a separate task, MANDATORY before #301 proceeds.

Agent: DarkLynx (session 17550faa)

Strycher:
Closes #364. **Epic #300 item 1** of Feature #295. Design-of-record: `docs/architecture/2026-07-17-repeater-telemetry-genericization-dor.md` §5 / D2.

**Pure internal refactor — no behaviour change, wire contract byte-compatible.**

## What this does

Moves the config-command **dispatch mechanism** out of the observer role so other roles can register their own config keys. Every key **handler** stays owned by its role.

New `src/helpers/config/ConfigDispatch.{h,cpp}`:
- **Provider registry** — each role registers one `{SetFn, GetFn}` pair; dispatch walks providers in registration order, first to claim a key wins.
- **Shared role-agnostic parse helpers** — `parseBool`, `parseIndexedKey`, `strEq` (were `parseConfigBool` / `parseBrokerKey` inside `ObserverCli.cpp`).
- **Fixed static storage, no heap.** POD table in `.bss`.
- **Loud on every target** (Arduino *and* host) if the table is empty or full.

`ObserverCli`: the former public `configSet`/`configGet` become file-static providers registered by a file-scope `ProviderRegistrar`.

`MyMesh`: the two wire call sites now use `offband::config::dispatchSet` / `dispatchGet`. The pool is no longer threaded through the config path.

## Why "Option A" (one function pair per role, not a per-key table)

The observer's key chain is **order-sensitive** and the ordering is load-bearing: exact `wifi.enabled` must be tested before the generic `wifi.` prefix, or the boolean switch gets written to NVS as a wifi field literally named `enabled`. Same trap with exact `mqtt.*` ahead of the `mqtt.broker.` family.

Registering one function pair let that chain move across **unmodified**, so ordering is preserved *by construction* rather than re-encoded into a table's precedence rules. Two thin local adapters bind `OFFBAND_MAX_BROKERS` to the shared helpers so the chain body itself is untouched.

## Verification

| Gate | Result |
|---|---|
| Client-visible reply/format strings | **131 strings, empty diff** before vs after (mechanical extract + compare, re-run against the post-rebase base) |
| Parse-helper parity vs originals | **52/52 inputs, 0 failures** — out-of-range slot, `00`/`+1` indices, NULL, leading whitespace, case variants. Host build, `-Wall -Wextra` clean |
| Registration actually links **and** runs | `.init_array` entry 6 → ObserverCli's static-init ctor → single `call8` to `config::registerProvider`, read off the linked ESP32-S3 image |
| Builds | **5/5 SUCCESS** — `heltec_v4`, `Heltec_v3`, `heltec_v4_tft`, `Xiao_S3_WIO` observer envs + `RAK_4631_companion_radio_usb` (nRF52) |
| Non-observer compilation | Component compiled clean into the nRF52 env with **zero observer sources present** (temporary filter add, reverted) |

The reply text is part of the contract — the client displays it — which is why the string diff is the primary gate rather than an eyeball review.

## Gemini adversarial review — adjudicated

Review log: `docs/llm-consultations/2026-07-24-364-config-dispatch-review-gemini-gemini-2.5-pro.log` (untracked, per repo convention for consult logs).

**1 BLOCKER / 2 MAJOR / 3 MINOR raised. 4 accepted, 2 rejected.**

### BLOCKER-1 — "linker may discard the registrar, killing the whole config surface" → accepted as hardening, **severity rejected**
Claim: `--gc-sections`/LTO could drop the file-scope registrar, so every config key would answer "unknown config key" on a deployed fleet.

**Not a current defect** — verified on the linked image before the review even ran: the ctor **is** `.init_array` entry 6 and its entire body is one call to `registerProvider`. It also cannot be dropped as an unextracted archive member, because other symbols in that TU (`dispatchObserverCli`, `configBrokerSlotCount`) are referenced elsewhere, and ESP-IDF's linker script `KEEP`s `.init_array`.

Added `__attribute__((used))` anyway — free, and the failure mode is severe enough to insure against a future toolchain/flag change. The code comment records it as **insurance, not a fix**, so nobody later mistakes it for a bug that existed.

### MAJOR-1 — error reporting only under `#ifdef ARDUINO` → **accepted, fixed**
Empty/full provider table was silent on non-Arduino builds, so a host or unit-test build would hide a dead config surface. Added an `fprintf(stderr, …)` path for non-Arduino. Kept non-fatal rather than `assert` — aborting is a policy decision this library shouldn't make for its callers.

### MAJOR-2 — first-provider-wins silently shadows overlapping keys → **accepted**
Real and live, not theoretical: the observer provider claims the **entire `wifi.` prefix**, so a future repeater/companion provider registered after it would never be asked about `wifi.ssid`/`wifi.pwd` — no error, no diagnostic.

Documented prominently in `ConfigDispatch.h` with the concrete `wifi.` example. Mechanical detection filed as **#366**, sequenced with/before #301. Also cross-linked to **#365** (Companion WiFi at runtime), where this trap would bite first.

### MINOR-2 — provider returning `false` may leave stale `reply` → **accepted**
Contract documented ("a provider MUST NOT leave content in `reply` when it returns false"). Verified the observer providers comply: only two `return false` paths each — the NULL-arg guard (before any write) and the final unknown-key return (buffer holds only `'\0'`).

### MINOR-1 — "remove the redundant `reply[0]='\0'` from providers" → **REJECTED**
The redundancy is deliberate: it keeps each provider correct if ever called directly (e.g. from a test), and removing those lines would edit the very function bodies whose byte-identity is this change's core safety property. Cost is one store instruction. Intent documented instead.

### MINOR-3 — "replace `eq()` with NULL-safe `strEq()` everywhere" → **REJECTED, premise is factually wrong**
The existing `eq()` **is** NULL-safe (`ObserverCli.cpp:25`):
```cpp
static bool eq(const char* a, const char* b) {
    return a != nullptr && b != nullptr && strcmp(a, b) == 0;
}
```
The stated impact ("would crash on a NULL key") does not exist. Separately, `eq` has 71 call sites across the whole CLI path — out of scope, and churning them in a zero-behaviour-change refactor is exactly the risk this change is structured to avoid.

## Explicitly NOT in this PR

- **Observer hardware no-regression gate** — Epic #300 decomposition **item 3**. Operating a real observer on this firmware against the **currently released** app. **MANDATORY before #301 proceeds.** Not done, not claimed; needs its own task and bench time.
- Repeater key registration (#301 / #305).
- Broker-slot enumeration API (`configBrokerSlotCount` / `configBrokerSlotPopulated` / `configRenderBrokerSlot`) — deliberately left observer-owned.
- Extracting the `wifi.*` handlers into a shared `WifiConfigProvider` — needed by **#365**; owner decision pending on whether that lands under #300 or #365.

## Review notes

- Base moved during the work (`0ae494c5` → `b49babc0`); rebased onto current `firmware-base` and **all audits + all 5 builds re-run after the rebase**. No file overlap with the intervening commits.
- Branch was force-pushed once (`--force-with-lease`) to update it after that rebase — unshared branch, no prior PR.

⚠ **`firmware-base` has no required status checks, so `--auto` merges immediately.** Please merge deliberately; I have not enabled auto-merge.

Agent: DarkLynx (session 17550faa)

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -28,10 +28,13 @@
 #include "helpers/wifi_observer/CliPassthrough.h"
 // Epic F: the config command codes (#160 contract) live in OffbandConfigProtocol.h,
 // now included unconditionally above (#241). Its typed dispatch backend (#165):
-#include "helpers/wifi_observer/ObserverCli.h"   // configSet/configGet + dispatchObserverCli
+#include "helpers/wifi_observer/ObserverCli.h"   // dispatchObserverCli + broker enumeration
+#include "helpers/config/ConfigDispatch.h"       // #364: role-agnostic config set/get
 // wifiObserverPool() lives in WifiObserver.h (heavy transitive includes); forward-
-// declare it here as CliPassthrough.cpp does. configSet takes a pool ref (only the
-// broker branch -- F3 -- consumes it, but the signature requires it for flat keys too).
+// declare it here as CliPassthrough.cpp does. Still needed by the VIEW passthrough
+// and the F3 broker enumeration below; the config set/get path no longer takes a
+// pool ref (#364 -- the shared dispatcher must not know about MqttBrokerPool, so
+// the observer's provider fetches the pool itself).
 namespace offband { MqttBrokerPool& wifiObserverPool(); }
 #endif
 
@@ -1275,7 +1278,7 @@ void MyMesh::handleOffbandConfigCmd(size_t len) {
     char* sp = strchr(payload, ' ');                 // split on FIRST space; value = remainder
     if (sp == nullptr) { writeOffbandConfigScalar(offband::OCFG_R_ERR, "set: expected '<key> <value>'"); return; }
     *sp = '\0';
-    if (!offband::configSet(payload, sp + 1, reply, sizeof(reply), offband::wifiObserverPool())) {
+    if (!offband::config::dispatchSet(payload, sp + 1, reply, sizeof(reply))) {
       writeOffbandConfigScalar(offband::OCFG_R_ERR, "unknown config key"); return;
     }
     bool err = (strncmp(reply, "ERROR", sizeof("ERROR") - 1) == 0);
@@ -1284,7 +1287,7 @@ void MyMesh::handleOffbandConfigCmd(size_t len) {
   }
 
   if (op == offband::OCFG_GET) {
-    if (!offband::configGet(payload, reply, sizeof(reply))) {
+    if (!offband::config::dispatchGet(payload, reply, sizeof(reply))) {
       writeOffbandConfigScalar(offband::OCFG_R_ERR, "unknown config key"); return;
     }
     bool err = (strncmp(reply, "ERROR", sizeof("ERROR") - 1) == 0);
--- a/src/helpers/config/ConfigDispatch.cpp
+++ b/src/helpers/config/ConfigDispatch.cpp
@@ -1,0 +1,133 @@
+// src/helpers/config/ConfigDispatch.cpp
+//
+// Epic #300 item 1 (#364). See ConfigDispatch.h for the model and rationale.
+
+#include "ConfigDispatch.h"
+#include <cstring>
+#include <cstdlib>
+#include <cstdio>    // fprintf -- the non-Arduino (host/test) diagnostic path
+
+#ifdef ARDUINO
+  #include <Arduino.h>   // Serial -- diagnostics only; Arduino-generic, NOT
+                         // ESP-only, so this component builds for nRF52 too.
+#endif
+
+namespace offband {
+namespace config {
+
+namespace {
+
+struct Provider {
+    SetFn       set_fn;
+    GetFn       get_fn;
+    const char* role;
+};
+
+// POD with static storage duration -> zero-initialised in .bss during static
+// initialisation, which completes before any dynamic initialiser runs. That is
+// what makes a file-scope ProviderRegistrar in another TU safe at any link order.
+Provider g_providers[kMaxProviders];
+int      g_count;
+
+// Latched so a client polling config cannot turn this into a log flood
+// (SAFELANE 11 rule 10: diagnostics must not become the outage).
+bool g_warned_empty;
+
+// SAFELANE 6: an unregistered config surface must not silently masquerade as
+// "unknown config key" to the client. Report it once, visibly.
+void warnIfNoProvider(const char* op) {
+    if (g_count != 0 || g_warned_empty) return;
+    g_warned_empty = true;
+    // #364 review MAJOR-1: must be loud on EVERY target, not just Arduino --
+    // otherwise a host/unit-test build hides a dead config surface entirely.
+#ifdef ARDUINO
+    Serial.printf("ERROR: config %s with NO provider registered "
+                  "(config surface is dead; a role failed to link/register)\n", op);
+#else
+    fprintf(stderr, "ERROR: config %s with NO provider registered "
+                    "(config surface is dead; a role failed to link/register)\n", op);
+#endif
+}
+
+}  // namespace
+
+bool registerProvider(SetFn set_fn, GetFn get_fn, const char* role_name) {
+    if (set_fn == nullptr && get_fn == nullptr) return false;
+    if (g_count >= kMaxProviders) {
+        // Loud on every target: a role silently dropped here loses its whole
+        // config surface (#364 review MAJOR-1).
+        const char* who = role_name != nullptr ? role_name : "?";
+#ifdef ARDUINO
+        Serial.printf("ERROR: config provider table full (%d); '%s' NOT registered\n",
+                      kMaxProviders, who);
+#else
+        fprintf(stderr, "ERROR: config provider table full (%d); '%s' NOT registered\n",
+                kMaxProviders, who);
+#endif
+        return false;
+    }
+    g_providers[g_count].set_fn = set_fn;
+    g_providers[g_count].get_fn = get_fn;
+    g_providers[g_count].role   = role_name;
+    ++g_count;
+    return true;
+}
+
+int providerCount() { return g_count; }
+
+bool dispatchSet(const char* key, const char* value, char* reply, size_t reply_size) {
+    if (key == nullptr || value == nullptr || reply == nullptr || reply_size == 0) return false;
+    reply[0] = '\0';
+    warnIfNoProvider("set");
+    for (int i = 0; i < g_count; ++i) {
+        if (g_providers[i].set_fn == nullptr) continue;
+        if (g_providers[i].set_fn(key, value, reply, reply_size)) return true;
+    }
+    return false;   // no role claims this key
+}
+
+bool dispatchGet(const char* key, char* reply, size_t reply_size) {
+    if (key == nullptr || reply == nullptr || reply_size == 0) return false;
+    reply[0] = '\0';
+    warnIfNoProvider("get");
+    for (int i = 0; i < g_count; ++i) {
+        if (g_providers[i].get_fn == nullptr) continue;
+        if (g_providers[i].get_fn(key, reply, reply_size)) return true;
+    }
+    return false;   // no role claims this key
+}
+
+// ---------------------------------------------------------------------------
+// Shared parse helpers
+// ---------------------------------------------------------------------------
+
+bool strEq(const char* a, const char* b) {
+    return a != nullptr && b != nullptr && strcmp(a, b) == 0;
+}
+
+bool parseBool(const char* v, bool& out) {
+    if (v == nullptr) return false;
+    while (*v == ' ') ++v;
+    if (strEq(v, "1") || strEq(v, "true")  || strEq(v, "on"))  { out = true;  return true; }
+    if (strEq(v, "0") || strEq(v, "false") || strEq(v, "off")) { out = false; return true; }
+    return false;
+}
+
+bool parseIndexedKey(const char* key, const char* prefix, int max_index,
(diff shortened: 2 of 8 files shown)


Strycher:
## ⚠ HOLD — do not merge yet. The CLEAN/MERGEABLE badge is misleading.

CI is **18/18 green** and GitHub reports `MERGEABLE` / `CLEAN`. **That is an artifact of an unpushed branch, not an absence of conflict.**

### The collision

`feat/325-companion-wifi-provisioning` (**5 commits, tip `8acea035`**, worktree `C:\Dev\.worktrees\meshcore-firmware-325`) **refactors the same code into a different home.**

- Its commit `1a225a65` — *"refactor(#325): split pool-free config surface out of ObserverCli"* — creates `src/helpers/wifi_observer/ConfigCli.{h,cpp}` (322 + 112 lines) and **removes 278 lines from `ObserverCli.cpp`**.
- This PR **rewrites 79 lines of that same region** into `src/helpers/config/ConfigDispatch.{h,cpp}`.
- **Overlapping files (4):** `examples/companion_radio/MyMesh.cpp`, `src/helpers/wifi_observer/ObserverCli.cpp`, `src/helpers/wifi_observer/ObserverCli.h`, `variants/heltec_v4/platformio.ini`.

`feat/325` **is not pushed** — it is absent from all 31 remote heads — so GitHub cannot see the conflict. Merging this PR first leaves that branch with a hard conflict across those 4 files, centred on a 278-line deletion vs a 79-line rewrite of the same functions.

Surfaced by **BrownWaterfall** (#365 research) and independently verified here against the local worktree and remote head list.

### The good news — these two are complementary, not competing

`ConfigCli`'s entry points are **signature-identical** to this PR's provider typedefs:

```cpp
// feat/325 — src/helpers/wifi_observer/ConfigCli.h:77-78
bool configCoreSet(const char* key, const char* value, char* reply, size_t reply_size);
bool configCoreGet(const char* key, char* reply, size_t reply_size);

// this PR — src/helpers/config/ConfigDispatch.h
typedef bool (*SetFn)(const char* key, const char* value, char* reply, size_t reply_size);
typedef bool (*GetFn)(const char* key, char* reply, size_t reply_size);
```

So `ConfigCli` registers as a provider with **zero adaptation**:
```cpp
static config::ProviderRegistrar _reg(&configCoreSet, &configCoreGet, "core");
```

#325 supplies **the handlers** (pool-free core surface); this PR supplies **the registration/dispatch mechanism**. The coherent end state is **one shared "core" provider owning `wifi.*`/`display.*`, registered by every role, with the observer registering `mqtt.*` separately** — which also satisfies this PR's own "KEY SPACES MUST BE DISJOINT" warning and stops #301's WiFi keys being silently shadowed (#366).

### Owner decision needed before merge

1. **Which vehicle lands first** — this PR, or #325 rebased onto it?
2. **Does `wifi.*` move to a shared core provider now** (folding in #325's `ConfigCli`), or does this PR land mechanism-only and #325 rebase on top?
3. **#325 vs #365** — still unresolved; #365 is parentless and #325 is `board:in-progress`.

Converting to **draft** to prevent an accidental merge while this is decided — the work itself is complete and verified, and it can be marked ready again immediately. No changes needed to the code as it stands.

Agent: DarkLynx (session 17550faa)

Strycher:
## Hold lifted — owner approved

Owner approved landing this first (2026-07-24): *"Proceed on #368. Then do #370 as its own change."*

The collision flagged above no longer needs an owner decision — RusticBay stood down on the duplicate extraction unprompted:

> *"I am standing down on the extraction; it is yours. I am not reserving `ConfigCli.*` or `ObserverCli.*`."*

`feat/325` keeps only its distinct work (`DualSerialInterface` serial/WiFi transport coexistence + live credential-change path) and drops its duplicate `ConfigCli` extraction, so the two branches stop competing on these 4 files. BrownWaterfall (#365) confirms this PR landing is their start gate.

### State at merge

| | |
|---|---|
| Commit | `bc13eefc` (rebased onto `b49babc0`) |
| Base | unchanged since verification — `b49babc0` |
| CI | **18/18 pass** |
| Mergeable | `CLEAN` |
| Reply-string byte-compat | 131 strings, empty diff (re-run post-rebase) |
| Parse-helper parity | 52/52, 0 failures |
| Builds | 5/5 local envs + 18 CI envs |

Auto-merge enabled (`--auto --rebase`) per the standard flow — agents never merge by hand.

### Deliberately NOT changed before landing

BrownWaterfall asked whether `display.*` and `wifi.*` could be split into two separate providers now, since they're becoming separately-advertised capabilities (`0x10` / `0x08`) — *"if your registration API makes that a one-liner… if it forces a rework, leave it."*

**It forces a rework, so leaving it.** The observer provider is a single order-sensitive if-chain covering `wifi.*`, `display.*` and `mqtt.*`. Splitting it means restructuring that chain — precisely what this PR avoided in order to keep the 131 reply strings and the key precedence byte-identical. Doing it here would spend this PR's core safety property for no functional gain. It belongs in **#370**, where the chain is being restructured anyway and can carry its own regression evidence.

### ⚠ Still outstanding — NOT satisfied by this PR

**Epic #300 item 3 — observer *behavioural* zero-regression** against the currently-released app, on real hardware. **MANDATORY before #301 proceeds.** No task exists yet; needs bench time. Everything green here is build-time and byte-level. Two agents independently declined to count a green build toward this gate.

### Follow-ups
- **#370** — `wifi.*` to a shared core provider (owner-approved; ⚠ ownership currently contested between two sessions, being resolved before work starts)
- **#366** — mechanically detect provider key-space overlap
- **#365** — Companion WiFi contract + caps `0x08`/`0x10` (BrownWaterfall)

Agent: DarkLynx (session 17550faa)

Issue #379, opened 2026-07-26
task: redacting serial-capture tool for MQTT observer no-reconnect diagnosis (OKI CoreScope#72)


Strycher:
## Context

Two CoreScope observers (`NPN-MC-Analyzer` = **gessaman**, `SN0MAN Observer 2` = **Offband**) stopped publishing to OKI MQTT at ~23:38 EDT 07-25 after an overnight WiFi outage and did **not** auto-recover. OKI infra + ingestion confirmed healthy — this is transmitter/firmware-side (MQTT client not reconnecting after network flap). Full infra diagnosis: OKI-Mesh/CoreScope#72.

To investigate the device side we need a **serial log** off the failing units. This task delivers a tester-facing capture tool.

## Scope

Extend the existing `scripts/_cap_serial.py` (reuse, not rewrite) into a redacting, tester-friendly capture:

- Auto-detect the serial port; capture to a timestamped file; print plain "run this → send us the file" instructions.
- **Firmware-agnostic** — a serial capture is a byte-dump, so one tool works on both the gessaman and Offband boxes. (We can only *interpret* the Offband output, but we can *capture* either.)
- **On-the-fly credential redaction** (in-line, not whole-line drop) so secrets never land in the file:
  - Label-grounded from actual Offband serial strings: `SSID=…` / `wifi.ssid = …` (`WifiBootstrap.cpp:109`, `ObserverCli.cpp:451`), `guest_password` / `bridge_secret` (`CommonCLI.cpp:1048/1140`), `jwt_email`, JWT-shaped tokens (`JwtHelper`).
  - Generic catch-alls for the gessaman box (unknown source): `password|psk|pass|secret|token|key = <value>`.
- **Capture-to-file only** — no auto-upload; nothing leaves the tester's machine automatically.

## Prerequisite / fallback

Serial capture requires the device be **USB-cabled to a computer the tester controls** at capture time. If the observers are headless-on-WiFi, serial is not the path — fall back to on-device `wifi_observer/CrashLog` or network-side logging. Documented here so the fallback isn't lost.

## Acceptance

- [ ] Script captures serial to a timestamped file with clear tester instructions.
- [ ] WiFi SSID + known Offband secrets redacted in-line; generic credential patterns caught.
- [ ] Redaction verified against sample lines (unit test in `scripts/test_*`).
- [ ] Gemini review clean before PR.

Strycher:
[...]

PR #383, opened 2026-07-28
fix: redact indented label-less CLI reply (get wifi.ssid) — SSID leak

### LINKED ISSUE #382: bug: serial-capture redaction leaks label-less CLI reply (get wifi.ssid) — leading-whitespace defeats the net
**Found on live bench (COM28 HV4):** `get wifi.ssid` over the serial console replies label-less as `  > <SSID>` (leading spaces + `> ` prefix), NOT `wifi.ssid = <SSID>`. Two defects let the SSID through:

1. The grounded `wifi.ssid =` rule matches the _sys-channel format, not the serial `> %s` (CommonCLI) format — so it never fired.
2. The label-less `> value` net was anchored `^>` (column 0); the real line has leading whitespace, so it didn't match. **SSID leaked.**

**Fix:** `^>` -> `^\s*>` in the CLI-reply net; add a test with the exact live format (leading spaces + `> ` + value). Gemini re-review of redaction. Parent tracking: #379.

Strycher:
Closes #382. Fixes the redacting serial-capture tool landed under #379.

## The leak (found on live bench)
`get wifi.ssid` over the **serial console** replies label-less and indented — `  > <SSID>` (CommonCLI `> %s`), not the `_sys`-channel `wifi.ssid = <SSID>` the grounded rule matched. The label-less net was anchored `^>` at column 0, so the leading whitespace defeated it and a **real SSID passed through unredacted**.

## Fix
`^>` → `^\s*>` — indented `> value` replies now redact to `> <redacted:cli-reply>`.

## Tests
Regression cases with the exact live format (space- and tab-indented `> value`, placeholder SSIDs) + an indented `[Tag]` line that must survive. All pass (15+ secret / 11 preserve).

## Gemini re-review
No over-redaction. Flagged a hypothetical value-on-next-line (`> \n<value>`) reply — **declined** the stateful rewrite (our firmware emits single-line `> %s`, verified in source + on device) and instead broadened the tester warning to any secret-returning `get` and documented the residual.

### DIFF
--- a/scripts/_cap_serial.py
+++ b/scripts/_cap_serial.py
@@ -18,10 +18,13 @@ plus conservative generic catch-alls for the unknown gessaman source.
 Redaction is IN-LINE (the secret value is replaced, the rest of the line is kept)
 so diagnostic context that shares a line with a secret is preserved.
 
-Known limitation: label-less interactive CLI replies — `get guest_password` /
-`get bridge_secret` print `> <value>` with no key — cannot be caught by
-line-checking. Testers are told below not to run password `get` commands during
-capture.
+Known limitation: label-less interactive CLI replies. The serial console prints
+`get wifi.ssid` / `get guest_password` / `get bridge_secret` replies as an
+indented "  > <value>" with no field name. The single-line form IS now caught
+(#382). A hypothetical value-on-the-NEXT-line reply ("> \n<value>") would not be —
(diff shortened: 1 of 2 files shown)


Strycher:
Human authorization received from Ben (session MaroonBridge / 655d2ac5) — fix plan incl. merge approved. Enabling auto-merge.

Issue #429, opened 2026-07-29
task: verify config manifest matches provider claim-logic (prevent silent drift)


Strycher:
Follow-up from the #366 Gemini review (MAJOR-2).

## Concern
#366's overlap detector relies on each provider declaring a **key-prefix manifest** that is maintained *separately* from the provider's actual claim if-chain (`observerConfigSet`). If a dev adds/removes a config key but forgets to update the manifest, the detector develops a blind spot — a real overlap could be missed (false negative). The `Keep this in sync` comment is a weak defense.

Not a defect in #366 (the observer manifest is verified-accurate for the current keys, incl. the concrete `wifi.` case). This tracks hardening against future drift.

## Proposed
A debug/host self-test that verifies the declared manifest matches what the provider actually claims: for each manifest entry, assert the provider claims a representative key; for a set of known non-keys, assert it does NOT. Makes drift a loud test failure rather than a silent latent bug.

Caveat: the observer provider is ESP-IDF/`ConfigSchema`/NVS-bound and won't link on host, so this likely needs an on-device debug-build self-test rather than a pure host test (same constraint that pushed #366 to a registration-time check over a host test).

## Sequencing
Before or with #301 (where the second provider + its manifest appear).

Agent: CloudyCliff (session 83c8582b)

Issue #297, opened 2026-07-18
Epic: Combined repeater+observer build profile (powered/PoE) — role compose on the converged MQTT stack


Strycher:
## Deferred deliverable (owner-directed 2026-07-17)

A single node that runs **both** the repeater and observer roles — the sweet spot on **powered / PoE** builds where the solar burst constraint doesn't apply (e.g. a Christ Hospital PoE node). Raised while scoping Feature #295.

**Deliberately deferred, NOT in #295's scope.** #295 genericizes the *repeater* WiFi/telemetry/command-queue/OTA path. This epic sits on top of #295's converged MQTT stack (one `esp_mqtt` `MqttBrokerPool`+`ConfigSchema` serving both roles) and adds the combined-role **build profile**: a combined env, role-compose wiring, config surface for both roles on one node, and combined-role testing.

**Prerequisite:** the converged publish stack from Feature #295 (D1 → Option A). This epic should not start until that lands — the convergence is what makes combining the roles cheap.

**Relates:** #295 (parent capability the convergence comes from), #135 (excluded nRF path).

PR #430, opened 2026-07-29
feat(#366): registration-time config-provider key-space overlap detector

### LINKED ISSUE #366: task: detect config-provider key-space overlap instead of silently shadowing (first-provider-wins hazard)

Strycher:
## What
Mechanical **registration-time overlap detector** for the config-dispatch component. The #364 dispatcher short-circuits on the first provider to claim a key, so a later provider is silently shadowed for any key both would claim. Live trap for #301: the observer claims the whole `wifi.` prefix, so a repeater's `wifi.mode` would be unreachable with no diagnostic. #364 mitigated by documentation only; this makes it mechanical.

## How (owner-approved approach (a))
- `registerProvider`/`ProviderRegistrar` take a **key-prefix manifest** (static, no heap; pointers to literals). Convention: entry ending `.` = prefix (`wifi.`), else exact leaf (`mqtt.iata`).
- At registration, each entry is checked against every already-registered provider's; a collision bumps `overlapWarningCount()` and stores a record.
- The **loud diagnostic is deferred to first `dispatchSet/Get`** (Serial up), not registration — registration runs during static init before `Serial.begin()`. Same lazy pattern as the existing `warnIfNoProvider`.
- Observer declares its actual manifest: `mqtt.iata`, `mqtt.status_interval`, `mqtt.broker.`, `display.always_on`, `display.rotation`, `wifi.`.

**Cost:** a few pointer slots per provider (`kMaxProviders`=4) — tens of bytes RAM, always-on. Accepted: silent config loss on a combined repeater+observer build (#297) is severe enough to warrant an unconditional guard.

## Verification
- **Native regression test** `test_config_overlap` PASSES — detects `wifi.mode` vs `wifi.`, no false positive on disjoint providers.
- Builds clean: **ESP32** `heltec_v4_companion_observer_wifi` · **nRF52** `RAK_4631_companion_radio_usb`.

## Adversarial review (standards#145)
Gemini v1: 1 BLOCKER / 1 MAJOR / 2 MINOR. Dispositions:
- **BLOCKER** (Serial during static init) — **fixed**: diagnostic deferred to first dispatch; count still increments at registration for the test.
- **MINOR-3** (leaf vs prefix ambiguity) — **fixed**: trailing-dot-aware `prefixesCollide`.
- **MAJOR** (manifest can drift from the if-chain) — real but distinct from overlap detection; manifest is verified-accurate for the current keys. Deferred → **#429**.
- **MINOR-4** (remove test `main()`) — **rejected**: codebase convention is per-file `main()`, one binary per test dir; no multi-main conflict.

Gemini **re-review of v2: all findings resolved, no new issues, PROCEED.**

Logs: `docs/llm-consultations/2026-07-28-366-overlap-detector-review-*.log` (+ `-rereview-`).

Closes #366. Follow-up: #429. Unblocks #301.

Agent: CloudyCliff (session 83c8582b)

### DIFF
--- a/platformio.ini
+++ b/platformio.ini
@@ -220,5 +220,6 @@ build_src_filter =
   -<*>
   +<../src/Utils.cpp>
   +<../src/CaptureRing.cpp>
+  +<../src/helpers/config/ConfigDispatch.cpp>
 lib_deps =
   google/googletest @ 1.17.0
--- a/src/helpers/config/ConfigDispatch.cpp
+++ b/src/helpers/config/ConfigDispatch.cpp
@@ -40,22 +40,41 @@ bool g_warned_empty;
 // correct build this is 0 forever.
 int g_overlap_count;
 
-// Two key-prefixes "collide" if one is a prefix of the other -- i.e. there is at
-// least one key string both would claim. Equal strings collide (min length is
-// their shared length). Empty strings are treated as non-colliding: a "" prefix
-// would claim everything, which is never a legitimate manifest entry, so we do
-// not let it force-collide with all keys.
+// Manifest-entry convention (#366 review MINOR-3): an entry ENDING in '.' is a
+// PREFIX (claims every key starting with it, e.g. "wifi." -> wifi.ssid/wifi.pwd);
+// an entry NOT ending in '.' is an EXACT leaf key (e.g. "mqtt.iata"). Two entries
+// collide iff some key string would be claimed by both:
+//   - prefix P vs anything X : collide iff X starts with P
+//   - leaf  L vs leaf L2      : collide iff L == L2
+// This avoids the false positive of a leaf "mqtt.iata" shadowing an unrelated
+// "mqtt.iata_x". Empty strings never collide (a "" prefix would claim all keys,
+// which is never a legitimate manifest entry).
 bool prefixesCollide(const char* a, const char* b) {
     if (a == nullptr || b == nullptr || a[0] == '\0' || b[0] == '\0') return false;
     size_t la = strlen(a), lb = strlen(b);
-    size_t n = la < lb ? la : lb;
-    return strncmp(a, b, n) == 0;
+    bool a_prefix = a[la - 1] == '.';
+    bool b_prefix = b[lb - 1] == '.';
+    if (a_prefix && strncmp(b, a, la) == 0) return true;   // b starts with prefix a
+    if (b_prefix && strncmp(a, b, lb) == 0) return true;   // a starts with prefix b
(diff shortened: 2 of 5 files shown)


Strycher:
Human merge approval received (owner, 2026-07-28): merge on green. Auto-merge lands immediately on this repo (no required checks), so waiting for CI to pass before merging. Agent: CloudyCliff (session 83c8582b)

Issue #471, opened 2026-07-30
chore: gitignore consultation/diagnostic logs and test-builds/ so they stay out of GitHub


Strycher:
@-

Strycher:
Landed on `chore/471-gitignore-logs` @ cefde733. Verified with `git check-ignore`: the three log/build paths match; `docs/diagnostics/*.md` correctly does not. 49 already-tracked consultation logs remain tracked by design.

PR #473, opened 2026-07-30
chore(#471): gitignore consultation/diagnostic logs + test-builds

Strycher:
@-


Strycher:
Superseded by #675 — same commit rebased onto current `firmware-base` (`e400eaec`). This branch was 422 commits behind; replaced rather than force-pushed.

Issue #524, opened 2026-08-01
bug: mention matching must accept both the untrimmed and trimmed forms of the node name — client normalisation is outside our control


Strycher:
## Summary

`MyMesh::textMentionsSelf()` matches an incoming `@[…]` token against the device's stored `node_name` **exactly**. That makes a successful mention dependent on the sending client not having normalised the name — and **we do not control, and cannot guarantee, client behaviour.**

Firmware must validate a mention against **both** forms of its own name: **untrimmed and trimmed**. A match on either is a match.

Owner ruling, 2026-08-01: *"If clients can potentially do this to names, then our matching needs to test both versions — untrimmed and trimmed — to validate matches, since we cannot control or guarantee client behavior. I can fix my client, but I can't fix what others may do."*

## Why this is separate from client#497

They are complementary, not alternatives, and both are required:

| Side | Rule | Issue |
|---|---|---|
| **Sender (our client)** | strict — emit the name **byte-for-byte**, never normalised | client#497 |
| **Receiver (firmware)** | tolerant — accept the raw name **or** its trimmed form | **this issue** |

Be strict in what you send, liberal in what you accept. Fixing only our client leaves every node running stock MeshCore, another fork, or a third-party client unable to mention an Offband node whose name has edge whitespace. Fixing only firmware would let our own client keep emitting a name it altered.

## Evidence this is real, not hypothetical

Observed live 2026-08-01 on a SenseCAP T1000-E. Our own client trimmed a trailing space out of the mention token while the device stored the name with it:

```
RESP_CODE_CONTACT           …ef b8 8f 20 00 00     contact record keeps 0x20
CMD_SEND_CHANNEL_TXT_MSG    …ef b8 8f 5d           mention emits name then ']', no 0x20

[mention] NO MATCH (name 22 bytes, candidate 21 bytes)
[mention] differ at codepoint 16 (byte 21): name=0x20 cand=0x00
```

If our own client did this without anyone intending it, an arbitrary third-party client is at least as likely to.

## Scope

In `MyMesh::textMentionsSelf()`:

- Compare the candidate against the stored `node_name` **as-is**.
- If that fails, compare against `node_name` with **leading and trailing whitespace removed**.
- A match on **either** returns true.
- Case-insensitivity on ASCII is unchanged.
- The stored name is **not** modified — this is match-time tolerance only. Nothing writes back a trimmed name.

The existing non-match logging stays; when both forms fail it should make clear that both were tried.

[...]

Issue #507, opened 2026-07-31
feature: configurable UI for headless devices — button-action matrix + notification scope


Strycher:
Configurable UI for headless / screenless mesh devices, starting with the SenseCAP T1000-E. Two epics, each with a matching client-side epic.

## Why

A screenless device's entire user interface is one button and one buzzer. Today both are hardcoded: the button-sequence→action map lives as lambdas wired up in `UITask::begin()` (`examples/companion_radio/ui-orig/UITask.cpp:64-75`), and the only notification control is a binary buzzer mute. Nothing is configurable from the client, and nothing is discoverable.

Two concrete defects motivating this, both found on the T1000-E this session:

- **The buzzer toggle has no observable success condition.** `handleButtonTriplePress()` beeps when turning the buzzer ON but is silent turning it OFF, and writes "Buzzer: OFF" to `_alert` — which renders nowhere on a `NullDisplayDriver` board. Pressing it produces no feedback, so the natural response is to press again, which turns it back on.
- **Notification is all-or-nothing.** No way to say "beep only when it's actually for me."

## Scope

| Epic | Side | Subject |
|---|---|---|
| E1 | firmware | Button-action matrix + set/get config surface + capability bit |
| E1 | client | Settings UI for the button-action matrix |
| E2 | firmware | Notification scope All/Self/None, on triple-press |
| E2 | client | Device notification-scope setting |

**Sequencing:** E1 lands first. E2's All/Self/None cycle is an *action*, so once E1's matrix exists it is an assignable entry rather than another hardcoded handler. E2 is blocked-by E1.

**Prerequisite:** the `offband_caps` byte is exhausted — see the caps-extension task. Neither epic allocates a bit until byte 2 exists.

## Design constraints (owner-directed, 2026-07-31)

**Generalisable, not T1000-E-specific.** The button-sequence→action matrix must be a declarative table that other headless devices can adopt, with per-device overrides where the hardware differs. It must live somewhere trivially editable — the way `OffbandConfigProtocol.h` holds the canonical cap-bit registry — **not** buried in function bodies and callback registration. The current `UITask::begin()` lambda wiring is exactly the anti-pattern to replace.

**Single-press stays default-off.** A user may opt into assigning it; nothing is bound to single-press out of the box.

## Prior art — concepts only, no code

[...]

PR #525, opened 2026-08-01
feat(#510, #524): notification scope ALL/SELF/NONE + 0xC5 device-UI command + whitespace-tolerant mention matching

Strycher:
Device notification scope + the `0xC5` device-UI command surface + whitespace-tolerant mention matching. Depends on caps byte 2 (#515, merged).

**Hardware-verified end to end, including against a client we do not control.**

## What this delivers

**Notification scope — ALL / SELF / NONE** (#510)
- `SELF` sounds on a DM, or a channel message that `@[mentions]` this node
- Cycled by triple-press with three distinct tones; `NONE` announces itself *before* muting, since the confirmation is the thing being disabled
- Persists at prefs offset 140, append-only

**`0xC5` device-UI command** (#509 partial)
- One code, sub-typed: `0x01` scope GET, `0x02` scope SET, `0x03` matrix GET, `0x04` matrix SET, `0x7F` error
- Errors carry a **reason byte**, not `ERR_CODE_ILLEGAL_ARG` — "no buzzer" and "unknown action" must be distinguishable to the user
- Matrix GET returns a **supported-actions mask** so the client renders the device-reported set, never a hardcoded list
- Canonical contract documented in `OffbandConfigProtocol.h`, the shared source of truth for both repos

**Whitespace-tolerant mention matching** (#524)
- Two tiers: exact byte-for-byte, then both sides whitespace-trimmed
- Bounded to **edge** whitespace only — interior differences and any non-whitespace codepoint difference still fail

**Long press is deliberately NOT assignable** — it is the CLI-rescue / power-off escape hatch, and a bad assignment could render a board unrecoverable.

## Two bugs found and fixed during hardware testing

**1. The feature was inoperative in the only configuration it exists for.** Upstream nested the `notify()` calls inside the `else` of an `isConnected()` check, so a device with a phone paired never sounded. The whole point is a pocketed tracker audibly signalling a DM or mention *while* paired. Both paths are now independent of client connection; the push tickle still fires alongside.

**2. Scope changes reached the pref but not the hardware.** `0xC5` SET updated `_prefs.buzzer_quiet` while the buzzer *driver* kept whatever state boot gave it. `play()` returns early when quiet, so the device stayed silent **while replying success** — indistinguishable from a working feature. `applyBuzzerMute()` (no-op default on `AbstractUITask`, implemented in all three UITask variants) now applies it, **unconditionally on every SET** so a drifted device self-heals rather than needing a reboot.

## `@[name]` is a cross-repo contract

Firmware matches the client's `_mentionsSelf` rule: **bracketed, case-insensitive**. A bare `@name` matches on neither side.

The contract as originally published was **underspecified** — it never said raw vs normalised, which is exactly why both sides were individually reasonable and jointly broken. Now explicit: **senders emit the name byte-for-byte; receivers additionally accept the whitespace-trimmed form.** Strict in what you send, liberal in what you accept. Client counterpart: OffbandMesh/meshcore-client#497.

⚠ Implemented from a functional spec. **Not** ported from the wadamesh fork — that project is GPL-3, this repo is MIT.

## Verification

| Check | Result |
|---|---|
| `test/test_mention` | **27/27** |
| `t1000e_companion_radio_ble` | SUCCESS |
| `Heltec_v3_companion_radio_ble` | SUCCESS |
| `RAK_4631_companion_radio_ble` | SUCCESS |
| Hardware — Offband client | **PASS** (owner-tested) |
| Hardware — **stock MeshCore client**, Heltec V4 OLED V4.2 | **PASS** (owner-tested) |

Wire captures confirm `0xC5` GET/SET round-trips, echoes stored state, and persists across reboot. The stock-client pass is the one that matters most: #524 exists precisely because third-party client behaviour is outside our control.

Buzzer-less boards (Heltec V3, RAK4631) build clean and correctly do **not** advertise the capability.

## `FIRMWARE_VER_CODE` 18 → 19

18 is merged and belongs to caps byte 2. Precedent is that adding a cap bit bumps the code even without a layout change — 17 was "+ caplog cap bit", 16 was "FEM LNA cap". Marked provisional; #365 and display-config take 20 and beyond in merge order.

Branch was brought current by **merging `firmware-base` in, not rebasing** — a rebase would require a force-push. Two conflicts, both resolved to this branch (the caps2 bit assignment and its registry row).

## Diagnostics

Non-matching mentions now log through `MESH_DEBUG_PRINTLN` → caplog: both strings **codepoint-rendered** (`<U+XXXX>` for non-printable), split across continuation lines rather than truncated, plus the first differing codepoint named. A truncated diagnostic is worse than none — it looks authoritative while discarding the tail, which is where the difference hides.

Previously a failed mention produced **nothing at all**, which is a §6 silent-failure violation and cost multiple hardware test cycles to reconstruct.

Tasks: `Crosswire-0g8`, `Crosswire-fop`. Parent feature: #507.

### DIFF
--- a/examples/companion_radio/AbstractUITask.h
+++ b/examples/companion_radio/AbstractUITask.h
@@ -42,5 +42,16 @@ public:
   virtual void msgRead(int msgcount) = 0;
   virtual void newMsg(uint8_t path_len, const char* from_name, const char* text, int msgcount) = 0;
   virtual void notify(UIEventType t = UIEventType::none) = 0;
+  // #510: push the low-level buzzer mute to the hardware driver.
+  //
+  // The buzzer is a MEMBER of each UITask implementation, not a global, so the mesh
+  // layer cannot reach it. Setting the notification scope over 0xC5 updated only the
+  // pref and left the driver in whatever state boot put it in -- play() then returned
+  // early and the device stayed silent while reporting success. This exists so a
+  // scope change applies to the hardware immediately instead of at the next reboot.
+  //
+  // Non-pure with a no-op default: implementations without a buzzer inherit the
+  // default and are unaffected.
+  virtual void applyBuzzerMute(bool quiet) { (void)quiet; }
   virtual void loop() = 0;
 };
--- a/examples/companion_radio/DataStore.cpp
+++ b/examples/companion_radio/DataStore.cpp
@@ -252,6 +252,15 @@ void DataStore::loadPrefsInt(const char *filename, NodePrefs& _prefs, double& no
     // field goes after caplog_level. (NodePrefs offset map: ...137 fem, 138 enabled, 139 level.)
     file.read((uint8_t *)&_prefs.caplog_enabled, sizeof(_prefs.caplog_enabled));          // 138
     file.read((uint8_t *)&_prefs.caplog_level, sizeof(_prefs.caplog_level));              // 139
+    // #510: notification scope, APPEND-ONLY after caplog_level. Files written before
+    // #510 end at 140, so this short-reads to EOF and leaves the caller-set default
+    // (NOTIFY_SCOPE_ALL = today's behaviour) intact -- an existing device is never
+    // silently muted by upgrading. uint8 is all-or-nothing, so no partial-read hazard.
+    // Keep this LAST; any new field goes after it.
+    file.read((uint8_t *)&_prefs.notify_scope, sizeof(_prefs.notify_scope));             // 140
+    // #509: button-action matrix, APPEND-ONLY after notify_scope (offsets 141..144).
+    // Short-reads to EOF on pre-#509 files, leaving the caller-set defaults.
+    file.read((uint8_t *)_prefs.button_actions, sizeof(_prefs.button_actions));           // 141
 
     file.close();
   }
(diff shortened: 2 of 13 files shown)


Strycher:
Agent: FuchsiaCreek (session df6013e6)

**Owner approved merge (Ben, 2026-08-01): "Merge #515, then #509/510."** #515 merged as `84728aa9`. Enabling auto-merge here per CLAUDE-BASE — the human decides, automation merges.

Strycher:
**Owner approved merge (Ben, 2026-08-01): "Merge #525".** Re-enabling auto-merge — auto-merge was disabled by the owner at 21:36:22Z and I left it alone until explicitly told to proceed. Per CLAUDE-BASE the human decides and automation merges; agents never merge manually.

Strycher:
## Merge-strategy exception — owner approved

**Owner approved `--squash` for this PR specifically (Ben, 2026-08-01).** Documented because it deviates from the repo standard of `gh pr merge --auto --rebase`.

**Why rebase could not be used:** this PR's history contains a merge commit (`f67e0a0a`, merging `firmware-base` in). GitHub cannot rebase-merge a PR containing merge commits, and reports that as the misleading `Base branch was modified` — SHAs were verified identical on both sides, nothing had actually moved.

**Why the merge commit exists:** `strict` branch protection requires the branch be up to date, and bringing it current by rebasing would have required a force-push, which is prohibited. Merging base in was the only compliant route, and it is what made rebase-merge impossible. Two rules that collide on a stacked branch.

**Why squash over merge:** keeps `firmware-base` linear, which is what the rebase standard exists to protect. Per-commit detail is not lost — the three commits and their full messages remain on this PR, and the substance is recorded on #510, #524 and #507.

Worth a follow-up: any branch that must be brought current under `strict` without force-push will hit this again. Not filing unprompted.

Issue #516, opened 2026-08-01
task: per-broker configurable MQTT keepalive (default 60)


Strycher:
Follow-up to #506 (root cause: mqtt2 `max_keepalive 60` rejected the firmware's keepalive-120 CONNECT → CONNACK 0x02).

#506's fix sets a **fixed** firmware MQTT keepalive of 60. This issue makes it **per-broker configurable** so it adapts to any broker's `max_keepalive` cap (some brokers may cap < 60).

## Scope
- New config field `mqtt.broker.N.keepalive` (seconds), default 60.
- Wire it to the esp-mqtt keepalive in `MqttBroker::begin()` (replaces the fixed 60 from #506).
- set / get / `mqtt view` surface — fits the config::dispatch unification (#511).
- Client UI support tracked separately in meshcore-client.

Origin: #506.
Agent: VioletBarn (session 82e07da7)

Strycher:
Gemini review of the #506 fix flagged a MINOR to fold in here: the esp_mqtt config struct is rebuilt **sparsely** on JWT refresh in `MqttBroker::tryConnect()` and `loop()` (uri+cert only) and pushed via `esp_mqtt_set_config`. It does NOT re-assert keepalive, so a JWT broker that caps keepalive could revert to the esp-mqtt default (120) after a refresh. When implementing per-broker configurable keepalive, **assert keepalive at ALL config-build sites** (begin + both refresh sites) — ideally via a shared `populateBaseConfig()` helper (Gemini's suggested refactor). Not a live bug today (okimesh is anon; #506's fix in begin() holds for it).

Agent: VioletBarn (session 82e07da7)

Strycher:
**Deferred** by owner decision (2026-08-02) — contingent on future need.

The fixed keepalive of 60 (from #506, merged in PR #517) works against every broker tested. This becomes necessary only if a broker caps `max_keepalive` **below** 60.

Known future trigger: DarkLynx flagged this as a hard dependency for **repeater** multi-broker rotation reconnects (#302). Re-open/prioritize when #302's rotation path needs it.

Agent: VioletBarn (session 82e07da7)

Strycher:
**Scope narrowed — the correctness half is split out to #532.**

This issue is now **only** the enhancement: making keepalive a per-broker config field. It stays **deferred**.

[...]

Issue #506, opened 2026-07-31
bug: companion-observer MQTT CONNECT rejected — CONNACK 0x02 (Identifier Rejected) on okimesh wss


Strycher:
## Summary

The companion-observer's MQTT publish to the okimesh broker (`mqtt2.okimesh.org:9002`, wss) fails: the broker completes the TLS handshake, then rejects the MQTT CONNECT with **CONNACK 0x02 (Identifier Rejected)**. The observer cannot publish to okimesh. Surfaced while hardware-testing #175 (TLS rotation) on HV3.

## Verified evidence (this session)

- **Device serial** (`[mqtt-err]` instrumentation): `type=2 sock_errno=0 tls_esp_err=0 tls_stack_err=0 cert_flags=0 connack=2` — TCP fine, TLS fine, MQTT **CONNACK 0x02**.
- **Host → `mqtt2.okimesh.org:9002`:** DNS + TCP + TLS + WebSocket-upgrade (`101 Switching Protocols`, subprotocol `mqtt`) all succeed. TLS verifies against the firmware's pinned **ISRG Root X1** (the server now serves an **ECDSA / ISRG Root X2** chain, cross-signed by X1 — validates cleanly).
- **Firmware sets no `client_id`** — esp-mqtt uses its internal default.
- **esp-mqtt's own debug logging is compiled out** of the arduino-esp32 libs (`CONFIG_LOG_MAXIMUM_LEVEL=1`), so the actual on-wire `client_id` / `clean_session` are **not passively observable** on-device.
- **Owner's live broker matrix** (direct client tests against okimesh): **any non-empty `client_id` → CONNACK 0x00 (accepted); only empty `client_id` + `clean_session=false` → 0x02.**

## Theories tried and REFUTED

1. **Server cert / ECDSA-X2 chain rejection** — refuted: TLS verifies clean, `cert_flags=0`, `tls_stack_err=0`.
2. **Broker-side client-ID policy / allowlist** — refuted by the owner matrix (no policy exists; only empty+persistent yields 0x02).

## Root cause: NOT yet confirmed

The esp-mqtt source generates `ESP32_<mac>` (non-empty) for a NULL `client_id`, which the matrix says is accepted — yet the device gets 0x02. The actual wire `client_id` / `clean_session` values have not been observed, and the debug path that would show them is compiled out. No fix is proposed until this is confirmed.

## Next diagnostic step

[...]

Issue #302, opened 2026-07-19
Epic: Multi-broker on the repeater path — heap-measured slot/TLS ceiling


Strycher:
Give the repeater the multi-broker pool (feed CoreScope + other MQTT sinks simultaneously), reusing the observer's heap-gated TLS concurrency (`OFFBAND_MAX_LIVE_TLS`, `HeldNoHeap`, #171/#177).

**Broker count is MEASURED, not guessed:** build `heltec_v4_repeater_telemetry`, measure free heap, derive the safe slot + live-TLS ceiling, record the measurement in the epic before choosing the number.

**Acceptance:** documented measurement; ceiling enforced; multiple brokers verified concurrently on hardware without heap exhaustion.

---
**Parent Feature:** #295 · **Scoped by:** #296 · **Design-of-record:** `docs/architecture/2026-07-17-repeater-telemetry-genericization-dor.md`

Strycher:
**Review-driven (MAJOR 4):** the heap ceiling is no longer measured *here* — it is an output of **Epic 0 (#308)**, measured **with LoRa active** before any of this work starts. This epic consumes that number rather than deriving it. Do not pick a broker/TLS count without #308's measurement.

Strycher:
## ⚠ Owner directive (2026-08-01) — broker rotation on the repeater is LAST-RESORT, sequenced

**Do NOT wire broker rotation into the repeater preemptively.** Default design = **hold all configured brokers concurrently**.

Sequence, binding:
1. **Prove need first.** The Epic-0 #308 heap-ceiling harness (`src/helpers/wifi_telemetry/HeapCeilingProbe.*`) measures the repeater's max concurrent broker/TLS contexts before the internal-DRAM floor. If that ceiling **≥ the node's configured brokers → NO rotation** (hold them all; drop the complexity).
2. **Only if the repeater is genuinely short** (the HV3-style low-heap case that forced rotation on Observer) do we consider rotation at all.
3. **If needed: fix rotation on OBSERVER builds FIRST**, then port the *fixed* mechanism to the repeater. Do **not** port the current Observer rotation as-is (it needs fixing), and do **not** build a fresh repeater-only rotation.

Rationale: Observer rotates because it's heap-starved (worst on HV3 — least free RAM). The repeater likely has more headroom and may not need it. Rotation is complexity we take on only when a measurement proves we must, and Observer is where the mechanism gets made correct before it travels.

[...]

PR #533, opened 2026-08-03
fix(#532): re-assert keepalive on every mqcfg rebuild (JWT refresh paths)

### LINKED ISSUE #532: bug: keepalive not re-asserted when mqcfg is rebuilt on JWT refresh (can revert to 120 → CONNACK 0x02)
Split out of #516 (which stays deferred as the *enhancement*). This half is a **latent correctness bug**, not an enhancement, and is a stated requirement for repeater multi-broker rotation (#302).

## Defect
`MqttBroker` builds an `esp_mqtt_client_config_t` in **three** places:

| Site | Sets |
|---|---|
| `begin()` | uri + cert + **keepalive (60, #506)** + auth |
| `tryConnect()` — JWT re-apply path | uri + cert + auth only |
| `loop()` — JWT refresh path | uri + cert + auth only |

The latter two rebuild a **sparse** config and push it via `esp_mqtt_set_config()`. They do **not** re-assert `keepalive`. On a broker that caps `max_keepalive` (e.g. 60), a JWT token refresh can therefore revert the session to esp-mqtt's default **120**, and the next CONNECT is rejected with the misleading **CONNACK 0x02** — the exact failure diagnosed in #506.

## Why it is latent today
The only keepalive-capped broker found so far (mqtt2.okimesh.org, `max_keepalive 60`) is **anonymous** (`auth_type=none`), so it never takes the JWT-refresh path. A **JWT** broker behind a keepalive cap would hit this.

## Why it matters now
DarkLynx flagged it as biting **repeater rotation specifically** (#302), where JWT brokers reconnect frequently — an intermittent, hard-to-trace failure.

## Fix
Centralize the base config so every build site is identical (Gemini's suggested shape from the #506 review):

```cpp
// private
void MqttBroker::populateBaseConfig(esp_mqtt_client_config_t& mqcfg) const;
// sets uri + keepalive + CA cert, both IDF branches
```
Called from `begin()`, `tryConnect()`, and `loop()` before `auth_->apply()`. No behaviour change on the first-connect path; removes the divergence on the refresh paths.

**Decision (owner, 2026-08-02): keepalive stays a FIXED 60.** Per-broker configurability (#516) is deferred and should NOT be gated on — this fix asserts the fixed value everywhere.

Origin: #516 / Gemini review of #506. Review log: `docs/llm-consultations/2026-08-01-506-review-prompt-gemini-gemini-2.5-pro.log`

Strycher:
## Summary
Fixes #532. `keepalive` was set only in `begin()`, not on the two JWT-refresh paths that rebuild the esp-mqtt config and push it via `esp_mqtt_set_config()`. A JWT broker behind a `max_keepalive` cap could therefore revert to esp-mqtt's default 120 after a token refresh and get the misleading CONNACK 0x02 root-caused in #506.

Split out of #516 by owner decision — **#516 (per-broker configurable keepalive) stays deferred; keepalive remains a FIXED 60.** This fix asserts that fixed value everywhere.

## Change (2 files)
New private `MqttBroker::populateBaseConfig()` sets uri + keepalive + CA cert for both IDF branches, called from all three config build sites (`begin()`, `tryConnect()`, `loop()`) before `auth_->apply()`. `keepalive` no longer appears outside the helper, so a future build site can't silently omit it.

Pure refactor on the first-connect path; the behaviour change is confined to the two refresh paths, which now stage the correct keepalive for the next CONNECT.

## Why it matters now
Latent for the observer (the one known capped broker is anonymous, so it never takes the refresh path), but flagged by DarkLynx as biting **repeater rotation** (#302), where JWT brokers reconnect frequently — an intermittent, hard-to-trace failure.

## Verification
- `Heltec_v3_companion_observer_wifi` build **SUCCESS**.
- **No hardware test.** Reproducing this needs a JWT broker that caps `max_keepalive` below 120; none is available on the bench. The known capped broker (`max_keepalive 60`) is anonymous and cannot exercise the JWT-refresh path. Stating this plainly rather than implying coverage that doesn't exist.

## Review
Gemini 2.5-pro. Confirmed the load-bearing question — `esp_mqtt_set_config()` stages the value for the *next* CONNECT, which is precisely the defect. Both IDF branches, `const`, and first-connect behaviour all confirmed correct. Its one MAJOR (guard mismatch) was **refuted with evidence** — all three call sites are already inside `ARDUINO`, and every env is `framework = arduino`; disposition recorded on #532.

Closes #532

### DIFF
--- a/src/helpers/wifi_observer/MqttBroker.cpp
+++ b/src/helpers/wifi_observer/MqttBroker.cpp
@@ -89,6 +89,46 @@ const char* lookupCaCertPem(const char* name) {
 }
 
 // ---------------------------------------------------------------------------
+// populateBaseConfig -- the ONE place transport-level esp_mqtt fields are set.
+// ---------------------------------------------------------------------------
+// #506: esp-mqtt defaults the MQTT keepalive to 120 s, but a broker can cap it
+// via max_keepalive and REJECT an over-limit CONNECT with a MISLEADING CONNACK
+// 0x02 "identifier rejected" (confirmed live against a broker with
+// max_keepalive 60 -- a paho client is accepted at k=60, rejected at k=120,
+// identical to this firmware). 60 s is accepted by capped and uncapped brokers
+// alike; OWNER RULING 2026-08-02: it stays a FIXED 60 (#516, per-broker
+// configurability, is deferred and must not be gated on).
+//
+// #532: this is called from ALL THREE config build sites -- begin() and the two
+// JWT-refresh paths in tryConnect()/loop(). The refresh paths previously set
+// only uri+cert and pushed that through esp_mqtt_set_config(), leaving keepalive
+// at the 120 default, so a JWT broker behind a cap silently reverted after a
+// token refresh. Any new field that must survive a refresh belongs HERE, not in
+// an individual call site.
+#if defined(ARDUINO) && defined(ESP_PLATFORM)
+void MqttBroker::populateBaseConfig(esp_mqtt_client_config_t& mqcfg) const {
+    const int keepalive_sec = 60;
+#if ESP_IDF_VERSION_MAJOR >= 5
+    mqcfg.broker.address.uri = cfg_.url;
+    mqcfg.session.keepalive  = keepalive_sec;
+    if (cfg_.transport == BrokerTransport::Tls ||
+        cfg_.transport == BrokerTransport::Wss) {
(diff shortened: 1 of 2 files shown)


Strycher:

Note carried from the PR body, not hidden: this fix has **no hardware test** — reproducing it requires a JWT broker capping `max_keepalive` below 120, and none is available on the bench (the one known capped broker is anonymous and never takes the JWT-refresh path). Verified by build + Gemini 2.5-pro review.

Issue #521, opened 2026-08-01
bug: heap-constrained observer (HV3) loads only 1 of 2 enabled wss brokers — pool won't hold both, blocks #175 rotation


Strycher:
## Summary
On the heap-constrained **Heltec V3** observer, the MQTT broker pool will not hold **two enabled wss brokers simultaneously**. Both are ENABLED in NVS, but the live pool loads only ONE wss broker at a time; enabling the second **evicts** the first. This **blocks the #175 TLS round-robin rotation test** — rotation needs 2+ enabled wss brokers live at once (competing for the 1-slot TLS budget) to have anything to cycle.

## Verified evidence (HV3, build = firmware-base + #175 rotation)
- `mqtt view 1` (mqtt2.okimesh.org wss) = **ENABLED**; `mqtt view 2` (eastme.sh wss) = **ENABLED** — both enabled in NVS.
- `mqtt status` / `[rot]` show only **one** wss broker loaded at a time (`enabled=2` = tcp + one wss). The other wss broker is **absent from the live pool** (not "held", not "disabled" — unconfigured in the pool).
- `mqtt disable 1; mqtt enable 1` **loaded slot 1 and evicted slot 2**; before the toggle it was the reverse (slot 2 loaded, slot 1 absent).
- **Survives reboot** — after a fresh hardware reset, slot 1 stayed absent from the pool despite NVS=ENABLED, until force-toggled. So it's a config-**LOAD** behaviour, not a rotation-strand.
- Heap context: HV3 free heap ~26–33 KB with one wss/TLS context live.

## Owner context + the real question
Multi-broker has been struggling, and **HV3 is the toughest environment** to get working — despite its popularity as an observer board. Open question:
- Would a **XIAO S3** or **HV4** (more heap) behave differently — is this a **heap limit** (only 1 esp_mqtt client + TLS context fits) or a **pool load/indexing logic bug** independent of heap?
- Depending on the answer: **HV3 may need to be dropped as an Observer candidate**, or **limited to 1 encrypted (wss/tls) path** — with TLS rotation (#175) only meaningful on higher-heap boards.

## Diagnosis needed
1. Heap-driven (does a 2nd `esp_mqtt_client_init` / broker load fail under HV3's low heap?) or logic bug (indexing/reconcile)?
2. Attribution: pre-existing in `firmware-base` broker-pool load, or introduced by #175's reconcile/rotation? (Test `firmware-base` without #175.)
[...]

PR #535, opened 2026-08-03
fix(#534): allocate the esp_mqtt client on promotion, not at configure time

### LINKED ISSUE #534: bug: parked (HeldNoHeap) brokers still hold an esp_mqtt client — deadlocks HV3 below the TLS heap floor
## Summary
`MqttBroker::begin()` creates an `esp_mqtt_client` for **every enabled** slot, including brokers that are immediately parked at `HeldNoHeap` because they are over the live-TLS budget. A parked broker holds a client it cannot use. On a heap-constrained observer this consumes the very headroom the TLS handshake floor is guarding, and **deadlocks the pool: no broker can ever be promoted.**

## Evidence (HV3, `firmware-base` with the merged #175 rotation fix)
With 2 wss brokers configured:

| Value | Bytes |
|---|---|
| `OFFBAND_TLS_HEAP_FLOOR_BYTES` | **81,920** (80 KB) |
| HV3 free heap | **76,924** |

76,924 < 81,920 → `tlsHeapBudgetOk()` returns false on every tick → **both** brokers sit `HeldNoHeap` forever. Observed over an 8-minute capture: `en=5 live=0/1`, `s1:HH s2:HH`, **0 rotations**, no broker ever promoted.

Measured cost of one extra parked client + auth on the observer build: **~18 KB**. Pre-fix, the #175 rotation bug had deconfigured one broker, so only one client existed → 94,840 free (above the floor) → it could connect. Fixing the stranding correctly retained both brokers, which exposed this.

## Root cause
A client is allocated at *configure* time rather than at *promote* time. Budget/heap admission is enforced in `tryConnect()` (`HeldNoHeap`), but allocation already happened in `begin()` — so the guard fires after the cost has been paid.

## Proposed fix
**Create the esp_mqtt client lazily, on promotion, not at `begin()`.**

- In `begin()`, for a TLS/wss slot that is over the live budget (or below the heap floor), store config and skip `esp_mqtt_client_init()` — leave it configured-but-clientless, exactly the state a rotated-out broker is now left in by `releaseClient()` (#175).
- The promotion path already exists and is proven: the pool's re-drive calls `reconcileSlot(s)` when `isTlsTransport(cfg) && budget_ok && !b.hasClient()`, which recreates the client before `tryConnect()`.
- Net effect: **parked brokers cost no client heap.** N wss brokers can be *configured* while only the budgeted number hold clients.

This also makes the invariant consistent — after #175, a rotated-out broker holds no client, but a never-yet-promoted broker does. Same logical state, two different resource footprints.

## Impact
- **HV3 becomes a viable rotating observer** (budget 1, N configured wss, time-shared) instead of deadlocking. Directly resolves the remaining half of #521.
- **Repeater (#302)**: DarkLynx measured ~30 KB steady / ~60 KB transient per wss context with ~196 KB free — no deadlock there, but the same per-parked-client cost reduces how many brokers can be configured before the floor bites. Flagged to them in Agent Mail.

## Acceptance
- HV3 with 2+ enabled wss brokers: free heap stays above `OFFBAND_TLS_HEAP_FLOOR_BYTES` when only one is live; one broker reaches `Up`; rotation cycles on the dwell timer (the #175 behaviour, verified on XIAO at 14 cycles, reproduced on HV3).
- No regression on a board where all configured brokers fit within budget.

## Origin
Found during the #175 hardware re-test on HV3 after PR #467 merged. Full write-up: #521.

Strycher:
## Summary
Fixes #534. Parked (`HeldNoHeap`) TLS brokers still held an allocated `esp_mqtt` client they could not use. Admission ran in `tryConnect()`, but allocation had already happened in `begin()` — the guard fired *after* the heap cost was paid. On HV3 that deadlocked the pool entirely.

## Before → after, on the board that deadlocked (HV3)

| | Before | After |
|---|---|---|
| Free heap (2 wss configured) | 76,924 — **below** the 81,920 floor | one client only; recovers to 84K+ between rotations |
| Brokers promoted | **0** — both stuck `HeldNoHeap` | s1 and s2 both reach `Up` |
| Rotations in capture | **0** (8 min) | **4**, strict s1↔s2 alternation |
| Parked broker | held a ~18 KB unusable client | holds **no** client |

Alternation is genuine, not one broker reclaiming: `s1:HH s2:UP/55s` → later `s1:UP/13s s2:DN`. Heap after each teardown: 84,140 / 90,468 / 80,180 / 83,356 — flat, no leak.

## Change (4 files)
- `MqttBroker::begin()` takes a **required** `may_alloc_client` (no default — mirrors `tryConnect`'s `tls_budget_ok` from #171 so no caller can silently bypass it). A denied TLS slot stores config, marks `HeldNoHeap`, returns **true** without allocating client or auth.
- New `MqttBrokerPool::tlsClientsAllocated()`. **Admission counts allocated clients, not live connections** — at boot nothing is `Up`/`Connecting`, so a live-based check would admit every slot and reproduce the deadlock. This is the load-bearing design decision.
- `reconcileSlot()` subtracts the slot's own client from the count (`begin()` destroys it before reallocating), so a reload doesn't count itself and refuse.
- Promotion reuses the existing re-drive path — `!hasClient() && budget_ok → reconcileSlot()` — no new machinery.

Also makes the invariant consistent: after #175 a *rotated-out* broker holds no client; now a *never-promoted* one doesn't either. Same logical state, same footprint.

## Review
Gemini 2.5-pro: **"No issues found."** Explicitly PASSed starvation/liveness, the self-count off-by-one, `HeldNoHeap` reuse for a second condition, concurrency of the unlocked `hasClient()` scan, tcp/single-wss regression, and `begin()`'s new true-return (correctly avoids a false `crashLogf`).

## Impact beyond HV3
Resolves the remaining half of #521 — HV3 is a viable rotating observer at budget 1, not a board to drop. Flagged to DarkLynx for #302: the repeater has ~196 KB headroom so it never deadlocked, but the same per-parked-client cost limited how many brokers could be configured.

Closes #534

### DIFF
--- a/src/helpers/wifi_observer/MqttBroker.cpp
+++ b/src/helpers/wifi_observer/MqttBroker.cpp
@@ -185,7 +185,8 @@ void MqttBroker::shutdown() {
 
 #ifdef ARDUINO
 bool MqttBroker::begin(uint8_t slot, const BrokerConfig& cfg,
-                       const mesh::LocalIdentity& identity) {
+                       const mesh::LocalIdentity& identity,
+                       bool may_alloc_client) {
 #if defined(ARDUINO) && defined(ESP_PLATFORM)
     ClientLockGuard _g(client_lock_);
 #endif
@@ -200,6 +201,19 @@ bool MqttBroker::begin(uint8_t slot, const BrokerConfig& cfg,
     // Enabling later routes through reloadSlot()->begin() which creates it.
     if (!cfg.enabled) {
         rt_ = BrokerRuntimeState{};
+        return true;
+    }
+    // #534: TLS/wss allocation admission. When the pool already holds its full
+    // quota of TLS clients, store the config and stay CONFIGURED BUT CLIENT-LESS
+    // rather than allocating a ~18KB client this broker cannot use. HeldNoHeap
+    // marks it as deferred-not-failed (same semantics tryConnect uses), so the
+    // pool's re-drive promotes it via reconcileSlot() once budget frees. NOT a
+    // failure: return true. tcp brokers hold no mbedTLS context and are exempt.
+    if (!may_alloc_client &&
+        (cfg.transport == BrokerTransport::Tls ||
+         cfg.transport == BrokerTransport::Wss)) {
+        rt_ = BrokerRuntimeState{};
+        rt_.state = BrokerState::HeldNoHeap;
         return true;
     }
     auth_ = makeAuth(cfg, identity);
--- a/src/helpers/wifi_observer/MqttBroker.h
+++ b/src/helpers/wifi_observer/MqttBroker.h
@@ -65,9 +65,24 @@ public:
     // Bind slot to config. Allocates esp_mqtt_client + MqttAuth.
     // Does NOT start the connection; pool decides via tryConnect().
     // Returns false if cfg is invalid (empty URL, unknown auth, etc.).
+    //
+    // #534: may_alloc_client is the pool's ALLOCATION admission for TLS/wss
+    // slots -- REQUIRED (no default, mirroring tryConnect's tls_budget_ok #171)
+    // so no caller can silently bypass it. false means the pool already holds
+    // OFFBAND_MAX_LIVE_TLS TLS clients, so this slot stores its config and stays
+    // CONFIGURED BUT CLIENT-LESS -- the same state releaseClient() leaves a
+    // rotated-out broker in (#175). Previously every enabled slot allocated a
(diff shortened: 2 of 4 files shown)


Strycher:

Hardware gate satisfied on the board that deadlocked (HV3): 0 promotions / 0 rotations before → 4 rotations with genuine s1↔s2 alternation after (parked broker actually promoted: `s1:HH s2:UP/55s`, then `s1:UP/13s`), heap flat across teardowns. Gemini 2.5-pro: no issues found.

PR #540, opened 2026-08-03
fix(#531): make reconciling_ std::atomic<bool> (cross-core memory ordering)

### LINKED ISSUE #531: task: make MqttBrokerPool::reconciling_ std::atomic<bool> (cross-task flag hardening)
Follow-up from the Gemini adversarial review of the #175 rotation fix (`releaseClient()`). Verdict was MINOR — **not a live defect**, a hardening of the inter-task contract.

## Finding
`MqttBrokerPool::reconciling_` is a plain `bool[OFFBAND_MAX_BROKERS]`. It is **written by the worker task** (`rotateTlsIfDue`, `reloadSlot`) and **read by loopTask** (`drainBroker`, the publish fan-outs, the status/`[rot]` enumerations). On a dual-core ESP32, plain `bool` reads/writes are not guaranteed atomic or coherent across cores, so loopTask can observe a stale `false` while the worker is mid-teardown.

## Why it is benign today
The correctness guarantee comes from a second layer: `MqttBroker::publish()` takes the per-broker `client_lock_` (recursive mutex) and re-checks `client_ != nullptr`. If loopTask races past a stale flag, it blocks on that lock until the teardown finishes, then the null check fails cleanly. So `reconciling_` is effectively an **optimization** (avoid spinning on the lock), not the safety mechanism.

## Why it is worth fixing
The current design requires every future loopTask access to a broker to be guarded by `client_lock_`. A future unguarded access would silently turn this benign race into a use-after-free. Making the flag atomic makes the contract explicit and self-enforcing.

## Proposed change
```cpp
// MqttBrokerPool.h
#include <atomic>
std::atomic<bool> reconciling_[OFFBAND_MAX_BROKERS];
```
Initialize to `false` in the constructor/`begin()`. No logic changes — all existing reads/writes work unchanged, they just become atomic operations.

## Why not folded into #175
#175's rotation fix was hardware-verified across 14 rotation cycles (strict alternation, `en=5` on 82/82 samples, flat heap across cycles). Folding in an untested change would land code that differs from what was verified. Split out so #175 ships exactly as tested.

Origin: #175 (PR #467). Review log: `docs/llm-consultations/2026-08-02-175-rotation-fix-review-gemini-gemini-2.5-pro.log`

Strycher:
## Summary
Closes #531. `MqttBrokerPool::reconciling_` — written by the lifecycle worker task, read by loopTask on a dual-core ESP32 — moves from `volatile bool` to `std::atomic<bool>`.

## Premise correction (recorded on #531)
The original review described a plain `bool` and a race where loopTask reads *"a stale value in a register"*. **That mechanism could not occur** — the field was `volatile`, which forbids register caching, and single-byte access on Xtensa is not torn.

The real gap is **ordering, not caching**: `volatile` is a *compiler* barrier only, with no memory-ordering or cross-core visibility guarantee, so there was no happens-before between the worker's write and loopTask's read. `std::atomic` (seq_cst) supplies it.

## Honest scope
**Not a live defect.** Correctness rests on the per-broker `client_lock_` inside `MqttBroker::publish()`, which re-checks `client_ != nullptr` — a mis-ordered read of this flag cannot produce a use-after-free today. The value is making the guard reliable *on its own*, so a future code path that touches a broker without taking that lock doesn't silently turn a benign race into a real one.

## Two things worth a reviewer's eye
1. **`= {}` is required, not cosmetic.** `std::atomic`'s default constructor is trivial in C++17 and leaves the value **indeterminate**. Dropping the old `= {false}` without replacing it would have introduced an uninitialized read — an easy miss on a "trivial" type swap.
2. **Lock-free verified, not assumed.** `nm` on `MqttBrokerPool.cpp.o` and on the linked image shows **zero `__atomic_*` symbols**, so accesses inline to native instructions with no libatomic locking on this toolchain.

## Verification
Build SUCCESS (`Heltec_v3_companion_observer_wifi`). **No hardware test** — this is a memory-ordering hardening with no observable runtime symptom to reproduce; a passing soak would not distinguish it from the old code. Saying so plainly rather than implying coverage.

## Review
Gemini 2.5-pro rated it **MAJOR** — *"a real, albeit latent, data race"*, and *"not cargo-culting; it is the correct C++ idiom"*. Confirmed the volatile-vs-ordering framing, the `= {}` initialization semantics, all 11 call sites (implicit load/store, no read-modify-write), and seq_cst as the right default. **Note:** the reviewer's MAJOR is a higher severity than this issue's `priority:P3` — flagging rather than re-labelling unilaterally, since board priority is yours.

Closes #531

### DIFF
--- a/src/helpers/wifi_observer/MqttBrokerPool.h
+++ b/src/helpers/wifi_observer/MqttBrokerPool.h
@@ -8,6 +8,7 @@
 #include "WifiObserverConfig.h"
 #include "MqttBroker.h"
 #include "MqttRingLog.h"
+#include <atomic>   // #531: reconciling_ is a cross-task flag
 
 // #175: every broker slot is a ring reader, so the ring must have at least as
 // many cursors as there are broker slots. Caught at compile time rather than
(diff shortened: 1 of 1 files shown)


Strycher:

Carried forward, not hidden: **no hardware test** — this is a memory-ordering hardening with no observable runtime symptom, so a passing soak would not distinguish it from the prior code. Verified by build SUCCESS, Gemini 2.5-pro review (rated MAJOR), and an empirical lock-free check (zero `__atomic_*` symbols in the object file and linked image).

PR #546, opened 2026-08-03
feat(#542): Epic A1 — user-controllable status/traffic LED (led on|off CLI)

Strycher:
Closes the A1 slice of #542 (Feature). Citadel `Crosswire-tf0`. Agent: BrownHawk (session 8a9de97d).

## Summary

Adds a `led on|off` CLI command that suppresses the Heltec V4 TX LED — the "light show" — persisted across reboot, defaulting to today's behavior (on). Mirrors the `fem on/off` precedent exactly.

- `MainBoard` capability virtuals (`setLedEnabled`/`canControlLed`/`isLedEnabled`), default unsupported so non-Heltec boards are unaffected.
- `HeltecV4Board` gates the TX-LED writes in `onBeforeTransmit`/`onAfterTransmit` behind `_led_enabled`; `setLedEnabled(false)` forces the pin LOW.
- `ui_led_enabled` persisted in `CommonCLIPrefs` at file offset **294** (append-only, matched read/write).
- Applied at boot on **all four CLI roles** — repeater, room_server, sensor (each guarded on `canControlLed()`). Better than planned: room_server + sensor got persistence too (clean apply hooks).

## Scope

- **A1 = the CLI-reachable roles only.** The **companion has no CLI** (it never instantiates `CommonCLI`), so it is **not** in A1 — its LED control is the `0xC5` app setting in Epic B. The companion firmware wiring is preserved for B at `docs/superpowers/epic-b-groundwork/` (patch + README) and in commit `2277c42f`.
- Display tristate = Epic A2 (separate). `0xC5` client contract = Epic B.

## Verification

- **Builds green:** repeater, companion_ble, companion_usb, room_server, sensor, tft_repeater, **RAK_4631** (nRF52 — proves the base-class change is cross-platform safe). Native suite passes (unaffected).
- **Gemini 2.5-pro adversarial review:** its BLOCKER (CommonCLI writing the companion struct at the wrong offset) was **verified and refuted** — the companion never uses CommonCLI; the two same-named `NodePrefs` never cross. The review *did* correctly drive the companion out of A1 (see Scope). Other findings were non-issues (verified). Log: `docs/llm-consultations/2026-08-03-542-a1-led-gemini-gemini-2.5-pro.log`.
- **Hardware round-trip PASSED** on hv4-bench-1 (Heltec V4.3 repeater): `led`→on / `led off`→off / **reboot→still off** / `led on`→on. Persistence proven on real hardware.

## Test plan (reviewer)

[...]

Issue #542, opened 2026-08-03
Feature: user-controllable indicators — kill the light show (status/traffic LED + OLED) via CLI + 0xC5 client setting


Strycher:
Turn off a device's **status/traffic LED and OLED** so a deployed node stops being a blinking, lit beacon — for battery and for stealth ("people wanted the light show off"). Owner-requested.

**Sibling to #507, not a child.** #507 is configurable UI for *headless* (screenless) devices; this is the opposite target — devices that *have* an LED and/or a screen. They share plumbing (0xC5 device-UI command, caps byte 2, the role-agnostic config path), so this Feature reuses #507's infrastructure without living under it.

## Who it serves

The light show is pure annoyance + drain on a deploy-and-forget box. Strong use case: **Repeater, Observer, Sensor, Room Server**. Weaker: **Companion** (carry device — you often *want* the screen), but the capability should still exist.

## Two delivery surfaces, by how the user reaches the device

- **CLI-reachable roles** (repeater/observer/sensor/room server): a human opens a serial/BLE terminal → `led` / `display` CLI serves them. All four roles compile `CommonCLI`, so one CLI change covers them at once.
- **Companion**: the user lives in the phone app. The CLI-over-API surface (`cliPassthrough`/`_sys`) is `OFFBAND_OBSERVER`-gated, so a plain companion has no way to reach the CLI. That user needs a rendered app setting → the **0xC5 client contract**.

## Scope of THIS Feature (firmware only)

Owner directive: wire the firmware side (CLI **and** the 0xC5 contract) so the client can be added later with no further firmware change. **The client UI is out of scope here** (separate repo, later).

1. **Board capability API** — `get/setDisplayMode` (tristate), `get/setLedEnabled`, `canControlDisplay/canControlLed` (mirrors `canControlLoRaFemLna()`); unsupported → reported, never a silent no-op.
2. **Persistence** — role-agnostic config path (#512), survives reboot, defaults = today's behavior.
3. **CLI** (all roles, `CommonCLI`):
   - `led on|off` — `on` = normal (default), `off` = never lights.
   - `display auto|always-on|always-off` — `auto` = on-with-timeout (default, today's behavior), `always-on` = no timeout, `always-off` = dark.
   - Reconcile with the existing observer `display always on/off` + `display flip` (#141/#148) so there is ONE display command surface, not two.
[...]

Issue #141, opened 2026-06-15
Observer/companion: keep display lit while on USB power


Strycher:
**User request (P2):** when the device runs on **USB power**, keep the display **lit** instead of applying the normal battery-saving blank/sleep.

**Rationale:** a USB-powered fixed/bench node has no power reason to blank its screen; users want it always visible when mains/USB powered.

**Scope:** observer + companion builds with a display (Heltec V3/V4 OLED, Heltec V4 TFT, XIAO, …).

**Option to consider (per request):** a `_sys`-channel CLI setting — e.g. `set display.always_on on` — so the user opts in explicitly. This also **sidesteps the hard part**: reliable USB-vs-battery power-source detection has been problematic on related projects, and a manual setting avoids needing to detect the power source at all.

_Preliminary feasibility investigation to follow in a comment._

Strycher:
## Preliminary feasibility (read-only investigation)

**TL;DR:** the `_sys`-setting route (`set display.always_on on`) is the clean, board-agnostic, feasible path. Auto-detecting "on USB" is reliable on nRF52/PMU boards but **not** on the ESP32-S3 observer boards (Heltec V3/V4, XIAO) — exactly the detection problem flagged. Recommend the manual setting.

### 1. There is a display-off to override
`examples/companion_radio/ui-new/UITask.cpp` blanks the display on idle (`_display->turnOff()` at UITask.cpp:816, 932) and wakes it on a button (`turnOn()` / `checkDisplayOn()` at 670, 744, 967). So "keep lit" = skip that idle `turnOff()` when a flag is set. All drivers expose `turnOff()`/`turnOn()` (`DisplayDriver.h`); on the TFT it's the backlight GPIO (`PIN_TFT_LEDA_CTL`), on OLED a display command.

### 2. USB-vs-battery detection — the hard part, and board-dependent
| Board family | USB-power detection in firmware | Reliable? |
|---|---|---|
| **ESP32-S3** (Heltec V3/V4, XIAO — main observer targets) | **none** — `ESP32Board.cpp` / `HeltecV4Board.*` read no VBUS/charge state; the CN3165 charger status isn't wired/read | ❌ no (would need to tap USB-OTG VBUS or a charger pin) |
| nRF52 (RAK, etc.) | `NRF_POWER->USBREGSTATUS & VBUSDETECT` (SoC hardware bit) — `NRF52Board.cpp:77` | ✅ yes |
[...]

PR #548, opened 2026-08-03
feat(#542): Epic A2 — display tristate (display auto|always-on|always-off CLI)

Strycher:
The A2 slice of #542. **Stacked on A1 (#546) — blocked-by #546; base auto-retargets to `firmware-base` when A1 merges.** Citadel `Crosswire-xl1`. Agent: BrownHawk (session 8a9de97d).

## Summary

Adds `display auto | always-on | always-off` to the OLED on the CLI-reachable roles (repeater, room_server, sensor), persisted across reboot, default `auto` (on with 20 s blank — today's behavior).

- `ui_display_mode` in `CommonCLIPrefs` at file offset **295** (append-only after A1's led@294).
- `display` CLI handler, `#ifdef DISPLAY_CLASS`-guarded (reports "unsupported" on no-display builds).
- The three (byte-identical) UITasks read the mode directly each `loop()` and apply it live via a `_last_disp_mode` transition detector — no applier.

## Scope

- **CLI roles only.** Companion (no CLI) + the observer `display always on/off` reconcile (#141/#148) are **Epic B**. On these three roles there is no prior `display` command, so `always off = dark` introduces no *redefinition* here.

## Verification

- **Builds green:** repeater, room_server, sensor, companion_ble, tft_repeater, RAK_4631 (nRF52).
- **Gemini 2.5-pro review found two real MAJOR state-machine bugs** — `always_on→auto` (stale-timer immediate blank) and `always_off→auto` (stuck dark). Both **fixed** with `_last_disp_mode` transition detection; **Gemini re-review = CONFIRMED-FIXED**. Adjudicated MINORs: button-in-always_off kept dark (design); `#ifdef`-guarding the pref rejected (would desync blob offsets across build types); 3-file UITask duplication deferred (pre-existing, out of scope). Logs in `docs/llm-consultations/`.
- **Hardware round-trip PASSED** on hv4-bench-1 (Heltec V4.3): auto (lit→blank) / always-on (stays lit) / always-off (dark, button won't wake, persists across reboot) / auto (relights then blanks). Both Gemini-caught transitions verified on real hardware.

## Test plan (reviewer)

- [ ] `display`, `display auto|always on|always off` behave as above on a Heltec V4
- [ ] `always off` survives reboot (dark through boot)
- [ ] `always_on → auto` does not immediately blank; `always_off → auto` relights
- [ ] No-display build reports `unsupported`

Requires human review + merge approval. **Do not merge before #546.** No native test (harness can't compile UITask/CommonCLI; owner-approved on-device test — same as A1).

### DIFF
--- a/docs/superpowers/plans/2026-08-02-indicator-display-tristate.md
+++ b/docs/superpowers/plans/2026-08-02-indicator-display-tristate.md
@@ -305,7 +305,19 @@ Scrutinize: offset 295 append correctness (read==write order); the `loop()` mode
 hoist in `begin()` doesn't read a null pointer; DISPLAY_CLASS guard covers every arm. Fix or justify
 every finding.
 
-- [ ] **Step 3: Hardware verify on hv4-bench-1 (Tier-2 flash — explicit per-flash owner GO)**
+- [x] **Step 3: Hardware verify on hv4-bench-1 — DONE, PASSED 2026-08-03**
+
+Verified on hv4-bench-1 (Heltec V4.3 repeater, A2 build sha `ccb2e018`, owner-authorized flash):
+- boot → `display` = `auto`, OLED lit then blanks (default preserved)
+- `display always on` → OLED stays lit past 20 s (owner-confirmed "still lit")
+- `display always off` → OLED goes dark; **button press does NOT wake it** (owner-confirmed the deliberate design)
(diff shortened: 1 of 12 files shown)


Strycher:
Rebased onto firmware-base (dropped the duplicate A1 commits after A1's rebase-merge); force-push per owner. Enabling auto-merge --rebase per owner authorization to land #546/#548.

Issue #611, opened 2026-08-10
feat: return on-air packet hash on channel-send confirmation (capability-gated) for CoreScope correlation


Strycher:
Enabling firmware change for Offband client epic OffbandMesh/meshcore-client#524 (show CoreScope observer counts on outgoing channel messages, correlated by on-air packet hash).

## Why firmware
The client cannot reliably reproduce the on-air packet hash for its own outgoing channel messages (encryption/timestamp/framing parity, no fallback). The firmware holds the exact transmitted bytes and already computes the authoritative hash.

## Verified
`Packet::calculatePacketHash` (src/Packet.cpp:41) = `SHA256(payloadType + payload)` truncated to `MAX_HASH_SIZE` (8), identical to CoreScope's `ComputeContentHash` for group-text. No new hashing needed.

## Ask
On `CMD_SEND_CHANNEL_TXT_MSG`, after `sendGroupMessage` builds and transmits the packet, return the packet's 8-byte `calculatePacketHash` to the companion.

## Backwards compatibility (required)
Today the channel-send reply is a bare `writeOKFrame()` (MyMesh.cpp:2332). Do NOT mutate that OK: stock MC / alternate clients rely on it. Gate the hash behind a NEW capability bit (next free `OFFBAND_CAP_*`) plus a `FIRMWARE_VER_CODE` bump. Capable clients (which advertise the bit) get an extended reply or a dedicated push frame carrying the 8 bytes; everyone else keeps the plain OK.

## Deliverable
Capability bit + version gate + hash-on-channel-send, with the frame layout documented so the client can parse it. Coordinate the exact frame shape with the client side before implementing.

Agent: VioletBarn — recorded retroactively 2026-08-12 on owner attribution (session id not known to the recording agent, FoggyCanyon/2d2b37f1).

Strycher:
Heads-up for @VioletBarn (firmware side). Client epic OffbandMesh/meshcore-client#524 is drafted and depends on this. Owner (Ben) approved spinning up #611 to define the contract.

## The ask (recap)
On `CMD_SEND_CHANNEL_TXT_MSG`, after `sendGroupMessage` transmits, return the packet's 8-byte `calculatePacketHash` (src/Packet.cpp:41) to the companion so the client can query CoreScope (`map.okimesh.org`) for observer_count by that hash. The hash already matches CoreScope's `ComputeContentHash` for group-text, so no new hashing.

## Contract questions for you (client needs these to build #550)
[...]

Issue #551, opened 2026-08-03
task: rework ui-orig Button to timestamped edge capture (debounce + instrumentation)


Strycher:
Child task of #527. Delivers the code increment. **#527 itself stays OPEN — the bug is not yet proven fixed on hardware.**

## What this delivers

- `ButtonSequencer` — the press-counting state machine, split out of `Button` with no Arduino dependency so it is unit-testable. Consumes `(timestamp, level)` edges and resolves gestures from the timestamps they carry, so a late loop costs latency instead of dropped presses.
- `Button` — captures edges into a single-producer/single-consumer queue. Interrupt producer behind `BUTTON_IRQ_CAPTURE` (t1000-e companion envs only); every other env keeps the polled producer feeding the same state machine, so no behaviour change there.
- Queue overflow is counted and logged, never silently dropped.
- Dual-producer race fixed (found by Gemini 2.5-pro review): the poll path masks interrupts across its whole read-check-push, since an interrupt landing mid-push overwrote the other producer's edge.
- Debounce 50 ms -> 10 ms, and the window now **defers** an edge instead of discarding it. The old discard rule erased the first release of a fast double-tap, so the second press stopped looking like a transition and the gesture collapsed to one click; a tap shorter than the window lost its release entirely, leaving the machine believing the button was held until it fired a LONG press 3 s later — power-off on the T1000-E.
- Producer-level diagnostics (`[btn] raw t=... src=isr|poll`, `[btn] capture: isr=N poll=N dropped=N irq=armed|OFF`) to separate a capture failure from a counting failure. Temporary; comes out once the layer is identified.

## Tests

29 native cases. The three fast-tap cases (30 ms press / 30 ms gap double, sub-50 ms-gap triple, whole-gesture-under-500 ms quad) **fail at `-DBUTTON_DEBOUNCE_TIME_MS=50` and pass at 10**, so they pin the defect rather than the fix. Also covers bounce bursts on both edges, sub-debounce tap must not wedge into power-off, late/batched drain, long-press consumption, and `millis()` wraparound.

## What remains on #527

Multi-tap gestures still intermittently resolve as SINGLE, with only one press-edge logged for a 600-800 ms gesture — the later press-downs are not reaching the counter at all. That is a capture failure below this work. The instrumented build is flashed and awaiting an owner press log, which will say whether the interrupt is armed.

Strycher:
[...]

PR #618, opened 2026-08-10
feat(#611): 0xC6 on-air packet-hash query for CoreScope correlation

Strycher:
## Summary
Implements #611. A capable client can query the **on-air packet hash** of its own outgoing channel message via a new `0xC6` command, so it can ask CoreScope how many nodes repeated that packet.

The client cannot compute this itself: for outgoing channel messages it stamps a *plaintext* content hash, while the on-air hash is over the **encrypted** payload and needs byte-exact parity on the `"<sender>: "` prefix, the timestamp, and the channel encryption.

## Wire contract
```
REQUEST                       REPLY
[0xC6][0x01][ts:4][chan:1]    [0xC6][0x01][ts:4][chan:1][hash:8]
(failure)                     [0xC6][0x7F][reason]   1=unknown key, 2=malformed
```
Canonical spec lives in `OffbandConfigProtocol.h` (firmware owns this contract, same discipline as `0xC5`). Key = `(msg_timestamp, channel_idx)`, both already present in `CMD_SEND_CHANNEL_TXT_MSG`. Reply echoes the key so the client can match reply→request without relying on ordering. Request is **exactly 7 bytes** — a longer frame returns `reason 2` rather than being silently accepted.

## Why a client-issued query, not an extended channel-send reply
Capability bits advertise **firmware → client**. The firmware cannot know whether the *connected* client is hash-capable, so gating an extended OK on a cap bit is impossible — it would have to mutate the reply for everyone or no one. A client-initiated query inverts it: only a capable client ever emits `0xC6`, so **`writeOKFrame()` on the channel-send path is untouched**. Zero stock-client surface.

## Verification
- **Round-trip CONFIRMED on hardware by the owner** — debug log shows the `0xC6` query and reply working end-to-end against a flashed device. Firmware communication is not the open issue.
- Flashed and exercised on `wismesh1w-companion` (RAK3401 WisMesh 1W Booster, `RAK_3401_companion_radio_ble`), config-preserving nRF52 DFU.
- Builds SUCCESS: `Heltec_v3_companion_radio_ble` (ESP32) + `RAK_4631_companion_radio_ble` + `RAK_3401_companion_radio_ble` (nRF52). Second/third arch deliberately, because this touches shared MeshCore-base code (`BaseChatMesh`) — the #350/#463 lesson.
- Gemini 2.5-pro: **APPROVED**. It specifically confirmed the load-bearing claim: the hash captured before `sendFloodScoped` equals what an off-air observer computes, because `calculatePacketHash` covers `payloadType + payload` only and later hops mutate *header* path bytes. Its MINOR (`len < 7` → `len != 7`) applied.

## Implementation notes
- `BaseChatMesh::sendGroupMessage` gains `out_pkt_hash = nullptr` (additive; existing callers unaffected) because the packet never escapes that function — it goes to `sendFloodScoped` and only a `bool` returns. The hash is taken **immediately after `createGroupDatagram`, before `sendFloodScoped`**: once queued, the pool may recycle the packet.
- 8-slot FIFO ring (104 B). A duplicate key **overwrites in place**, so a repeated key can never return a stale hash.
- Ring is cleared in the constructor. Static-storage zero-init makes that unnecessary today, but the guarantee is non-local and a garbage `used` flag would hand the client a bogus hash to correlate.
- Key uniqueness across *different* message texts is the client's job (documented in the contract) and is confirmed accepted by the client agent: monotonic per-channel timestamps, bump to `last_ts + 1` on reuse. AES-128 ECB with no IV is deterministic, so same-text collisions yield the same hash and are harmless — only different text under one key is ambiguous.

## Allocations (recorded in the same commit per #514)
- `0xC6 CMD_OFFBAND_PKT_HASH` — registry "next free" advanced to `0xC7`
- `OFFBAND_CAP2_PKT_HASH = 0x08` — caps byte 2, bit 3
- `FIRMWARE_VER_CODE 21 → 22`

Swept clean before claiming, across **every** remote branch and both repos' open issues. The cap bit is advertised **unconditionally** — unlike its neighbours it has no hardware dependency, only the send path.

## Cross-repo
Contract negotiated and settled with the client agent (QuietSnow) before implementation; client counterparts are OffbandMesh/meshcore-client#550 / #551 under epic #524. The client half is gated on this landing.

Refs #611

Agent: VioletBarn — recorded retroactively 2026-08-12 on owner attribution (session id not known to the recording agent, FoggyCanyon/2d2b37f1).

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -209,6 +209,42 @@ namespace offband { MqttBrokerPool& wifiObserverPool(); }
 #define AUTO_ADD_ROOM_SERVER      (1 << 3)  // 0x08 - auto-add Room Server (ADV_TYPE_ROOM)
 #define AUTO_ADD_SENSOR           (1 << 4)  // 0x10 - auto-add Sensor (ADV_TYPE_SENSOR)
 
+// #611: record the on-air hash of an outgoing channel message. A repeat of the same
+// (timestamp, channel) key OVERWRITES in place rather than adding a second entry, so
+// a lookup can never return a stale hash for a key the client is asking about now.
+// The client is responsible for not reusing a key across two DIFFERENT message texts
+// (see the 0xC6 contract in OffbandConfigProtocol.h) -- same text under the same key
+// is deterministic (AES-128 ECB, no IV) and therefore yields the same hash anyway.
+void MyMesh::recordSentPktHash(uint32_t msg_timestamp, uint8_t channel_idx, const uint8_t* hash) {
+  for (uint8_t s = 0; s < offband::OFFBAND_PKTHASH_RING_SLOTS; s++) {
+    if (_pkthash_ring[s].used &&
+        _pkthash_ring[s].msg_timestamp == msg_timestamp &&
+        _pkthash_ring[s].channel_idx == channel_idx) {
+      memcpy(_pkthash_ring[s].hash, hash, MAX_HASH_SIZE);
+      return;
+    }
+  }
+  SentPktHash& e = _pkthash_ring[_pkthash_next];
+  e.msg_timestamp = msg_timestamp;
+  e.channel_idx   = channel_idx;
+  memcpy(e.hash, hash, MAX_HASH_SIZE);
+  e.used = true;
+  _pkthash_next = (uint8_t)((_pkthash_next + 1) % offband::OFFBAND_PKTHASH_RING_SLOTS);
+}
+
+// #611: returns the retained hash for a key, or nullptr if it was never sent this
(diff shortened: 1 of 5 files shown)


Strycher:

Verification of record: **round-trip confirmed on hardware by the owner** (debug log shows the 0xC6 query and reply working against a flashed `wismesh1w-companion`); builds green on `Heltec_v3_companion_radio_ble` + `RAK_4631_companion_radio_ble` + `RAK_3401_companion_radio_ble`; Gemini 2.5-pro APPROVED with its MINOR applied.

On merge, the provisional allocations become permanent: `0xC6`, `OFFBAND_CAP2_PKT_HASH = 0x08` (caps byte 2 bit 3), `FIRMWARE_VER_CODE 22`.

Issue #631, opened 2026-08-11
task: prefs-migration hardware verification (ESP32 + nRF52)


Strycher:
Parent: #614 · Assessment: #615 (§4.3, §4.5)

**Hardware gate for the prefs migration.** This is the acceptance test that CI structurally cannot perform -- CI has no pre-existing `/com_prefs` to migrate.

### Why this gate exists
The migration is silent (mis-read values pass `constrain()`), persists into `/prefs.json` on boot one, and affects every deployed node. A green matrix proves nothing about it.

### Test
On a bench device with a **real pre-update `/com_prefs` carrying non-default values** -- FEM RX gain, `ui_led_enabled`, caplog enabled/level, syslog host/port:

1. Capture the pre-update prefs (and a full backup) before flashing.
2. Flash the 1.17.0 integration build.
3. Confirm on first boot: correct layout detected + logged; every setting above survives with its pre-update value; `/com_prefs` still on disk.
4. Reboot; confirm values persist from `/prefs.json`.
5. Repeat on a **stock-MeshCore-written** `/com_prefs` (293 bytes) and confirm the upstream layout is chosen instead.

Run on **ESP32** (`hv4-bench-1`) **and nRF52** -- the two filesystem/storage paths differ.

### Discipline
- Every flash needs its own explicit owner GO naming the device; one approval = one run of `pio-flash`.
- Full pre-flash backup first; do not use a production node (patio, the car RAK4631, `Strycher-WIO`, the companion).

### Acceptance criteria
- [ ] ESP32: Offband-written prefs migrate with zero value loss
- [ ] nRF52: same
- [ ] Stock-written prefs (293 B) correctly detected as upstream layout
- [ ] Chosen layout appears in the log every time

Agent: FoggyCanyon (session 2d2b37f1)

Strycher:
## PASS — Offband-legacy → 1.17.0 migration verified on hardware

**Device:** `ST-P` (Heltec V4 OLED, ESP32-S3) · **Date:** 2026-08-12

Migration **observed executing**, not inferred from output. Captured via caplog (forced on with
`-D OFFBAND_FORCE_CAPLOG`), which buffers from the first line of `loadPrefs()` while USB is still
down and is dumped once serial attaches:

```
[4259] DEBUG: [prefs] legacy /com_prefs len=298 -> family=common layout=offband (length-unique-offband)
```

Recorded at **t=4.26 s** — long before a serial connection could exist. No `REFUSED`, no
[...]

Issue #628, opened 2026-08-11
task: 1.17.0 integration branch — absorb 320 upstream commits, resolve the 35 conflicts


Strycher:
Parent: #614 · Assessment: #615 (§2, §3)

Cut the integration branch and absorb MeshCore 1.17.0.

- **320** upstream commits (202 non-merge), **358** files, **+11,506 / -1,734**
- **35** textually-conflicting files (vs 18 in the 1.16.0 cycle)
- Target: `companion-v1.17.0` (`2af6126c`). All three role tags resolve to an identical tree, so there is one target.

### Conflict groups (assessment §3)
- 4 CI workflows we deleted / upstream modified -> **keep deleted**, we run `ci.yml` + `ci-green` (#477)
- 6 variant `platformio.ini` + root -> mechanical unions
- 6 FEM/LNA files -> resolved per the sibling FEM task
- `CommonCLI.{h,cpp}`, `NodePrefs.h`, `DataStore.cpp` -> resolved per the sibling prefs task
- `SerialBLEInterface`, `ESP32Board`, `MeshCore.h`, 7x `main.cpp`/`UITask.cpp`

### Discipline
- Own worktree, own branch off **current** `firmware-base` -- never a rolling upstream pull.
- **Re-run the conflict scan immediately before merging**, not from the assessment snapshot: base updates collide at merge time, and #612 (#607 clock sanity) touches `CommonCLI.cpp`, `DataStore.cpp` and `platformio.ini`.
- **"Auto-merged" is textual, not semantic** -- the 1.16.0 cycle's `radio_driver.*` break (#129) auto-merged cleanly and still broke the link per-variant.

### Acceptance criteria
- [ ] All 35 conflicts resolved with a written rationale per group
- [ ] Full `ci-green` matrix passes

Agent: FoggyCanyon (session 2d2b37f1)

Strycher:
Agent: FoggyCanyon (session 2d2b37f1) · Citadel `Crosswire-dd0`

Starting with **diagnosis, not resolution**: cutting the integration branch off current `firmware-base` and running the real merge to capture the **actual** conflict set. The 35-file figure from the assessment came from `merge-tree`, which is a prediction — and `firmware-base` has moved since (now `b9f6919f`, +3 commits incl. the #627 detector and #615 docs).

No conflict will be resolved until a written plan exists and is agreed (Epic #614 step 3).

Strycher:
[...]

Issue #627, opened 2026-08-11
task: prefs migration — layout detector + Offband offsets across all roles and both prefs subsystems


Strycher:
Parent: #614 · Assessment: #615 (§4, §4.5)

**The critical task of the 1.17.0 update.** Design + native-testable implementation of the prefs migration, so no deployed node silently corrupts its settings.

1.17.0 replaces the flat binary `/com_prefs` with a keyed `/prefs.json` (`ConfigSerializer`) and migrates once, automatically, on first boot -- using **upstream's** byte offsets, which are not ours.

### Required behaviour

1. **Legacy-format detection by record length** -- three distinct producers, 71 bytes apart, no overlap:
   - **293** bytes -> stock MeshCore 1.16.0 layout (291 `flood_max_unscoped`, 292 `flood_max_advert`)
   - **364** bytes -> Offband layout (291 `radio_fem_rxgain`, 294-363 UI + caplog + syslog)
   - 295 bytes -> stock 1.17.0 (in practice never seen; 1.17.0 writes `/prefs.json`)

   Rule: `>= 364` -> Offband; `<= 295` -> upstream; anything else -> upstream as the conservative default. **Log the chosen layout loudly** -- never pick silently.

   This matters because cross-base flashes are routine on this fleet: the T1000-E and Wio Tracker both arrived on stock Meshtastic, and `wismesh1w-companion` has a recorded MeshCore-1.15.0 -> Offband transition.

2. **Offband keys added to `NodePrefs::structure()`** so they round-trip in the keyed format: `ui_led_enabled`, `ui_display_mode`, `caplog_enabled`, `caplog_level`, `syslog_host`, `syslog_port`.

3. **`fem_rxgain` mapped to `radio_fem_rxgain`** -- the tagged upstream `structure()` maps it to `rx_boosted_gain`. Already fixed in upstream PR #3137; if that has merged by integration time we inherit it, otherwise implement the same mapping so it converges cleanly later.

4. **`/com_prefs` retained on disk** (upstream already leaves it) so a bad migration is recoverable.

### Acceptance criteria

- [ ] Detection implemented, with the chosen layout logged
- [ ] Offband keys round-trip through save -> load
- [ ] Unit tests under `test/` with **byte-exact fixtures of all three record shapes** (native env, no hardware)
- [ ] Verified that `ConfigSerializer` skips unknown keys -- currently `[hypothesis: source-read, not device-tested]`

### Out of scope
[...]

Issue #614, opened 2026-08-10
epic: MeshCore 1.17.0 base update


Strycher:
Parent: #613 (Feature: upstream MeshCore base currency)

## Goal

Decide whether — and if so, how — to move the Offband fork's base from **MeshCore 1.16.0** to **MeshCore 1.17.0** (released upstream 2026-08-09), and execute that decision.

**A decision not to move, recorded with rationale, closes this Epic legitimately.** The outcome is not pre-committed to an integration.

## Starting state

| | |
|---|---|
| Offband base | MeshCore **1.16.0** (landed `offband-v1.0.0` via epic #126 / integration #134) |
| `firmware-base` tip | `bfdbb33a` = `offband-v1.3.0` |
| Upstream remote | `upstream` = `meshcore-dev/MeshCore`, **fetch-only / `no-push`** — Offband does not track upstream (#197) |
| Target | MeshCore **1.17.0** (exact tag to be confirmed during assessment) |

## Shape (mirrors the 1.16.0 cycle, #126)

1. **Assess** — impact assessment against `firmware-base`. Output: commit volume, textual-conflict surface (`merge-tree`), API/ABI breaks against Offband's additive surface, on-flash compatibility risk, variant + build-matrix impact, and a go / no-go / defer recommendation. *(This Epic's first child.)*
2. **Decide** — owner go / no-go on the assessment. Human tollgate.
3. **Plan** — if go: a written plan under `docs/plans/`, per-file conflict resolutions, and the child-task breakdown. The 1.16.0 plan is the template.
4. **Integrate** — controlled integration branch off `firmware-base`, in its own worktree. Never a rolling upstream pull.
5. **Gate** — full `ci-green` matrix (#477) + hardware smoke on at least one ESP32 and one nRF52 board, then owner sign-off.

## Known high-risk surfaces (from the 1.16.0 cycle — verify, do not assume they repeat)

- **On-flash prefs layout.** 1.16.0's worst conflict was a binary-layout collision in the `CommonCLI` prefs serializer — Offband's `radio_fem_rxgain` versus new upstream fields at flash offset 291. Any upstream field addition to that struct is an upgrade-compatibility hazard for already-deployed nodes.
- **Fork-private protocol values.** Cap bits, the `0xC`-range CMD codes, and `FIRMWARE_VER_CODE` are Offband-allocated in a range upstream also grows into. Canonical registry: `examples/companion_radio/OffbandConfigProtocol.h`.
[...]

PR #636, opened 2026-08-11
task(#627): legacy-prefs layout detector for the MeshCore 1.17.0 migration

Strycher:
Part of #627 · Epic #614 · Feature #613

**Deliberately no `Closes` keyword.** #627 is not finished by this PR — see *Not in scope* below. Closing it on merge would discard its remaining gates.

## What this is

MeshCore 1.17.0 replaces the flat binary prefs record with a keyed `/prefs.json` and migrates **once, automatically, on the first boot after the update** — reading the old record with **upstream's** byte offsets. Offband's tail fields are permuted against upstream's:

```
offset:    290        291          292          293          294
Offband:   rx_gain    fem_rxgain   fld_unscop   fld_advert   ui_led ...
Upstream:  rx_gain    fld_unscop   fld_advert   fem_rxgain   cad_enabled
```

Reading one with the other's table silently mis-assigns values — every wrong value is a small int that passes `constrain()` — and the result is persisted to `/prefs.json` on boot one. On a deployed device. CI cannot catch it, because CI has no pre-existing legacy record to migrate.

Two upgrade paths exist and need different readers: **stock MeshCore → Offband**, and **Offband → Offband**. This PR is the decision logic that picks.

## How it decides

| Priority | Signal | Notes |
|---|---|---|
| 1 | **Record length** | Across released versions the sets are disjoint — Path A: Offband `{292, 294, 364}` vs upstream `{291, 293, 295}`. Derived from the record end offset at each release tag, not from dev-tree commits |
| 2 | **Field plausibility** at the one colliding length (295: upstream 1.17.0 vs an unreleased Offband build) | `b291 > 1` ⇒ cannot be Offband (`fem_rxgain` clamped 0..1); `b293 > 1` ⇒ cannot be upstream (same field, other position). Pure content — **works on nRF52**, which has no NVS |
| 3 | **NVS marker `cw_boot`** (ESP32 only, corroboration) | Written every boot via `crashLogStandardInit()`; upstream opens no NVS namespace anywhere. Only ever *proves* Offband — absence is never evidence |
| 4 | **Fail closed** | Undetermined ⇒ do **not** migrate, do **not** fall back to either table. Caller leaves the legacy file untouched, logs at error level, boots on defaults so an operator intervenes (SAFELANE §6) |

`identifyLegacyPrefs()` additionally derives the *family* (CommonCLI vs companion) from the record itself, so `/node_prefs` — read as a fallback by **both** subsystems — is never dispatched on the running role.

Decision logic has no Arduino/ESP-IDF dependency, so it is host-testable; reading length, tail bytes and the marker stays with the caller.

## Adversarial review (required gate)

Gemini 2.5 Pro, log committed at `docs/llm-consultations/2026-08-10-627-prefs-layout-detector-gemini-gemini-2.5-pro.log`. Initial verdict: *"not safe to merge."* Three findings, all real, all fixed in `2250d00f`:

- **BLOCKER — unsound content tiebreak.** The first version tested `b292 > 1 ⇒ Offband`, but `b292` carries a flood limit under *both* layouts and discriminates nothing. An upstream record with `flood_max_unscoped` set to 0/1 and the default `flood_max_advert = 8` was confidently mis-classified as Offband — the exact corruption this code prevents, defeating fail-closed. Replaced with two tests of one sound shape (a value exceeding a constrained boolean range rules that layout out), with both-firing treated as corruption rather than a tie.
- **BLOCKER — role-based dispatch.** A device re-flashed into another role could hand a 294-byte repeater record to the companion reader. Fixed by `identifyLegacyPrefs()`.
- **MAJOR — Path B accepted any length > 137 as Offband**, swallowing truncated Path-A records. Now released lengths only: upstream `137`; Offband `137` (≤ v1.1.2, byte-identical — nothing appended yet), `138` (v1.2.0), `147` (v1.3.0).

Plus the 295 special-case moved out of the unique-length check (it was `&& ev.length != 295`, a trap for the next editor) with an explicit overlap guard, and `toString()` no longer masks unhandled enum values.

## Testing

**24 native tests, full native suite (10 suites) green.** No hardware involved and none claimed.

Covers every released length on both paths, the collision resolved by content in both directions, the nRF52 no-marker case, degenerate content falling back to the marker then failing closed, marker-never-overrides-an-unambiguous-length, contradiction, all-zero tail, truncation, family identification including the role-swap case, and a sweep over every length × seven tail shapes.

Worth naming: the BLOCKER survived the first test suite because its "exhaustive" sweep swept *length* against a single hardcoded tail and proved nothing about the content path. That sweep is now two-dimensional.

## Not in scope — why #627 stays open

- Not wired into any loader. `CommonCLI::loadPrefs` / `DataStore::loadPrefs` are untouched; that lands with the integration (#628).
- Offband keys not yet added to `NodePrefs::structure()`.
- `fem_rxgain` → `radio_fem_rxgain` mapping not yet applied.
- mtime check for the downgrade → re-upgrade cycle not yet implemented.
- Observer NVS (store C) undisturbed-by-migration check not yet done.
- `ConfigSerializer` skipping unknown keys is still `[hypothesis: source-read, not device-tested]`.
- On-device verification is the separate hardware gate, #631.

## Risk

Additive only — new files plus one `platformio.ini` line adding the source to the `native` test env. No existing code path calls it yet, so this cannot change device behaviour on merge.

Awaiting human merge approval; auto-merge not enabled.


Strycher:
Human merge approval received from the owner in chat: *"merge it when ci-green passes"* (2026-08-11).

Enabling auto-merge (`--auto --rebase`) per CLAUDE-BASE. Since #477, `firmware-base` requires the `ci-green` aggregate — full build matrix plus `config-lint`, strict up-to-date, enforce_admins — so this queues behind the matrix rather than merging immediately. Agent does not force-merge or bypass protections.

Issue #649, opened 2026-08-11
task: script the env-name vs declared-capability cross-check (machine + human output)


Strycher:
Parent: #647 · Feature #613

**Part 2 of the board-delta step. Scriptable, machine + human readable output.**

## The invariant

An env's **name** is a claim about its capabilities. That claim must agree with its **resolved
config**. Where they disagree, one of the two is wrong and a human must look.

```
env name contains _ble        -> BLE must be genuinely configured
env name contains _wifi/_usb/
                  _ethernet   -> BLE must be deliberately absent
```

Checked against the **resolved** PlatformIO config (`pio project config --json-output`), never
by grepping `.ini` files — values inherit through `extends` chains, so only PlatformIO's own
interpolation knows what an env actually ends up with. `scripts/check_esp32_ble_deps.py` already
establishes this pattern; reuse it.

## Relationship to the existing lint

`check_esp32_ble_deps.py` (#199) asks: *"is this env internally consistent — does it declare
exactly one of the two escape hatches?"* It is satisfied by an env that is coherently configured
but **coherently wrong**: a `*_ble` env that cleanly excludes `SerialBLEInterface.cpp` passes
today, and ships with no BLE.

This check asks the different question: *"does the config match what the env says it is?"*

Both are needed. Neither is a superset of the other.

## Output

- **Machine** — JSON to a path or stdout: one record per env with `{env, mcu, claimed, actual, verdict, reason}`. Non-zero exit on any mismatch so it can gate `config-lint`.
- **Human** — an aligned table, offenders first, each row saying what was claimed, what was found, and which of the two to change. The #199 lint's "Fix each by adding EITHER … OR …" wording is the bar: the message should tell you the remedy, not just the violation.

## Extensibility

Start with BLE, since that is where the divergence bites (#288 NimBLE vs upstream Bluedroid) and
where #645 already drew blood. Structure it so a capability is a **table entry, not a code
path** — WiFi, Ethernet, GPS should be addable without restructuring.

## Gating relationship to #646

#646 would flip the ESP32 BLE default to opt-in. That removes the per-board tax but creates a
silent failure: a BLE board with no flag builds clean and ships without BLE. **This script is
the layer that makes that flip safe, because it is the only check that knows what the env was
supposed to be.** #646 must not land before this exists, and both should be green in the same
PR.

## Done when

- [ ] Runs from repo root, no args, exits non-zero on mismatch
[...]

Issue #647, opened 2026-08-11
epic: board-delta assessment as a required step of every upstream base update


Strycher:
Parent: #613 (Feature: upstream MeshCore base currency)

## The gap

The 1.17.0 base update (#628) had no step that asked **"what boards did upstream add, and are
they configured correctly for Offband?"** The merge plan covered conflicting *files*. New
*boards* arrive without conflicting with anything — they are pure additions, so every
file-oriented review passes them silently.

That is how #645 happened: 21 upstream-new ESP32 envs landed already violating the #199 NimBLE
invariant. Nothing was broken by the merge; the envs were simply never Offband-correct in the
first place, and only `config-lint` noticed. Had `config-lint` not existed, they would have
shipped unbuildable, exactly as in the original #199 (64 unbuildable envs, four board-by-board
recurrences).

`[verified]` This is structural, not a one-off:

- Offband diverges from upstream on capability-bearing config — NimBLE vs Bluedroid (#288), FEM
  handling (#298/#327), display, prefs layout.
- Upstream adds boards on essentially every release. 1.17.0 alone added **six** ESP32 board
  families plus several nRF52.
- Upstream has no reason to know about, or satisfy, any Offband invariant.

So **the board delta is a first-class review surface on every base update**, alongside the file
conflicts, and it currently has no owner, no checklist, and no tooling.

## The two halves

**Part 1 — assessment (research).** On every base update, enumerate the board/env delta and
establish each new board's actual capabilities: MCU family (ESP32 / ESP32-S3 / C6 / nRF52840),
radio, and which of BLE / WiFi / Ethernet / GPS / display / FEM it carries. Then validate the
Offband flags against that capability set. This needs judgement and datasheets — it is not
derivable from the `.ini` alone, because the `.ini` is exactly what we are auditing.

**Part 2 — cross-check (scriptable).** Independently, assert **internal coherence**: what an env
*claims* by name and by config must agree. `*_ble` in the env name should imply the BLE
configuration is actually present; `*_wifi`/`*_usb`/`*_ethernet` should imply it is deliberately
absent. Machine-readable output for CI, human-readable for review.

The two are complementary and neither subsumes the other:

- Part 2 alone cannot catch a board whose *name* is silent about a capability it has.
- Part 1 alone does not scale and will be skipped under momentum — which is the whole argument
  of the standards for mechanical enforcement.

## Why now

[...]

Issue #646, opened 2026-08-11
task: guard SerialBLEInterface.h so non-BLE ESP32 envs stop needing per-env NimBLE bookkeeping


Strycher:
Follow-up to #645 · relates to #199, #288

## The recurring tax

`helpers/esp32/SerialBLEInterface.h` includes `<NimBLEDevice.h>` **unconditionally — there is
no `#ifdef` guard**. Every ESP32 companion env pulls `+<helpers/esp32/*.cpp>` in with a greedy
wildcard, so a USB / WiFi / Ethernet env compiles the BLE interface it never uses and then
fails to build for want of a Bluetooth stack it should not need.

The current defence is per-env bookkeeping (#199): every ESP32 env must declare **exactly one**
of `${esp32_ble.lib_deps}` or `${esp32_no_ble.build_src_filter}`, enforced by
`scripts/check_esp32_ble_deps.py` in `config-lint`.

That lint is good and should stay. But it is a **detector, not a fix**, and the underlying
defect has now bitten repeatedly:

- originally left **64** ESP32 envs unbuildable, two of them in the shipped release matrix
- recurred **four times** board-by-board before anyone chased it (#199)
- and again in **#645** — the MeshCore 1.17.0 base update brought **21 new upstream envs**, none
  of which declare either hatch, because upstream is still Bluedroid and has no such constraint

`[verified]` The pattern is structural: **every future upstream base update adds ESP32 boards,
and every one of them will fail this lint on arrival.** Offband is on NimBLE (#288) and upstream
is not, so this does not converge on its own.

## What to do

Guard the NimBLE include (and the class body) so an env that does not use BLE can compile the
translation unit harmlessly — the file becomes inert rather than uncompilable. Non-BLE envs then
need no `.ini` bookkeeping at all, and a new upstream board builds on arrival.

Open questions for whoever picks this up:

- What is the right gate — an explicit `-D OFFBAND_ENABLE_BLE`, or keying off something already
  present? An implicit gate risks silently disabling BLE on an env that wanted it, which is worse
  than a build failure.
- Does `${esp32_no_ble.build_src_filter}` stay as belt-and-braces (excluding the file is still
  strictly cheaper than compiling an empty one), or get retired?
- The lint should stay either way, retargeted at the new invariant.

## Why this is not on the 1.17.0 branch

#645 took the reversible per-env fix instead. This one touches `SerialBLEInterface`, which the
[...]

Issue #638, opened 2026-08-11
chore: agents must not print raw `pio project config` output — it renders resolved secrets


Strycher:
## What happened

While validating merged INI syntax on the #628 integration branch, I ran `pio project config` and printed its tail. PlatformIO resolves `extra_configs = ${sysenv.PIO_SECRETS_FILE}` and **renders the resolved secret sections into stdout** — a `[cmdrelay]` block with a LAN IP and a live `admin_token`. Those values went into a session transcript.

## What was NOT affected

`[verified]` The repository is clean and the secrets mechanism worked as designed:

- The values come from `${sysenv.PIO_SECRETS_FILE}`, resolved from an env var to a file outside the tracked tree
- No tracked file contains `admin_token` (`git grep`)
- The value has never been committed (`git log --all -S`)
- `.gitignore` already covers `platformio.local.ini` and `platformio.secrets.ini`

So this is an **agent-output hygiene defect, not a repo defect.** No code change is needed to protect the repo.

## Why it happened

`pio project config` is an obvious-looking way to validate that a merged INI tree parses. Its output is not obviously credential-bearing — the secrets arrive through an indirection (`sysenv` → external file) that is invisible at the call site. I printed the output without gating on whether it could carry credentials, which is the failure mode already recorded in project memory from a prior serial-capture leak: *never display captured output assuming it is clean; gate the print on knowing it is.*

## Proposed hardening

1. **Document the trap** — `pio project config`, `pio project data`, and anything else that resolves `extra_configs` must never be printed raw by an agent.
2. **Safe validation recipe** for "does the INI tree parse": rely on the exit code alone, or use `--json-output` piped through a projection that selects only env names. Never `tail`/`head` the human-readable form.
3. **Consider a PreToolUse guard** in the same family as `block-raw-flash` — refuse `pio project config` unless its output is filtered. Cheap and mechanical; behavioural discipline is exactly what degrades under momentum.

## Not in scope for this issue

Whether the exposed token warrants rotation is the owner's decision and the owner's action. Recorded here only that the exposure was to a session transcript, not to git.

Agent: FoggyCanyon (session 2d2b37f1)

PR #650, opened 2026-08-11
task: cross-check env names against their resolved config (board-delta part 2)

Strycher:
Refs #649 · Epic #647 · Feature #613

Part 2 of the board-delta step. **Independent of the 1.17.0 merge (#644)** — branched from
`firmware-base`, so it can land on its own schedule.

## What it asserts

An env name is a claim. `..._companion_radio_ble` says the companion link is BLE;
`..._companion_radio_usb` says it is not. This fails the build when the claim and the resolved
PlatformIO config disagree — because then one of the two is wrong and a human has to decide
which.

## Why it is not the same as `check_esp32_ble_deps.py` (#199)

| | Question | Satisfied by |
|---|---|---|
| **#199** | is this env internally **coherent** — can it link? | any env declaring exactly one escape hatch |
| **#649** | does the config match what the env **says it is**? | any env whose name and flags agree |

#199 passes an env that is coherently configured and **coherently wrong**: a `*_ble` env with no
`BLE_PIN_CODE` links cleanly and ships with no BLE. Nothing catches that but the name.

`[verified]` This is not theoretical symmetry — the 21 envs that **failed** #199 on the 1.17.0
branch all **pass** this check. Neither subsumes the other.

## The first detector was wrong. That is the interesting part.

Detecting on `build_src_filter` — *"does this env compile `SerialBLEInterface.cpp`?"* — produced
**19 findings on `firmware-base`, every one a false positive.** Two causes, both found by
investigating instead of filing:

1. **nRF52 companion envs pull `+<helpers/nrf52/*.cpp>` with a greedy wildcard**, so they compile
   the BLE interface even when `BLE_PIN_CODE` is commented out. They build it and never
   instantiate it. **Compiled ≠ enabled.**
2. **`heltec_v4_companion_observer_wifi`** and its three siblings `extends` the BLE companion env
   and add observer telemetry over WiFi. The `_wifi` qualifies the **observer**, not the companion
   link — they genuinely define `BLE_PIN_CODE`. Anchoring the claim on the trailing name token
   reported all four as violations, permanently.

A check whose first run emits 19 false positives is worse than no check: it gets switched off.

What actually gates BLE is `BLE_PIN_CODE`, because that is what `examples/companion_radio/main.cpp`
branches on:

```
ESP32:  #ifdef WIFI_SSID -> SerialWifiInterface   (WiFi WINS)
        #elif defined(BLE_PIN_CODE) -> SerialBLEInterface
        #else -> ArduinoSerialInterface
nRF52:  #ifdef BLE_PIN_CODE -> SerialBLEInterface
        #else -> ArduinoSerialInterface
```

So the detector reads `build_flags`; ESP32 additionally requires `WIFI_SSID` to be **absent**; and
the claim is anchored to `_companion_radio_<transport>`, not the trailing token.

## Verification

- **24 host tests, deliberately negative.** Each is a mismatch the tool must fail on, or a trap
  this implementation actually fell into (the `nibble_zero_connect` substring trap, the
  observer-`_wifi` trap, `WIFI_SSID` overriding `BLE_PIN_CODE`, nRF52 not needing NimBLE).
- **Mutation-checked, so the tests are not vacuous.** Switching the claim parser to substring
  matching, or making the nRF52 detector require NimBLE, or making the ESP32 detector ignore the
  exclusion, each flips a verdict the tests assert on.
- **Live negative test on the real tree:** neutralising `BLE_PIN_CODE` on
  `heltec_v4_companion_radio_ble` is caught as `MISSING` with exit 1; reverted clean.
- **Clean runs:** `firmware-base` 163 claims, 1.17.0 branch 193 claims, all honoured.

## Output

- **Human** — table, offenders first, each row carrying its own remedy.
- **Machine** — JSON report, uploaded as a CI artifact, non-zero exit on mismatch.

Wired into `config-lint`, which `ci-green` already depends on.

## Note on secrets

The config read **never** echoes pio's output on the error path. That output is the
fully-interpolated project config including the gitignored secrets sections — which is how a live
credential reached a transcript once already (#638).

## Gating relationship to #646

#646 would flip the ESP32 BLE default to opt-in. That removes the per-board tax but creates a
**silent** failure: a BLE board with no flag builds clean and ships without BLE. **This is the
layer that makes that flip safe**, because it is the only check that knows what the env was
supposed to be. #646 should not land before this.


Strycher:
OWNER HOLD (2026-08-13, relayed from chat): Ben is holding this PR — it needs more work before landing. Not on the current release critical path; do not update-branch/auto-merge until he lifts the hold.

Agent: WhiteFalcon (session cb12cdd2)

Issue #697, opened 2026-08-14
task: NVS write results are unchecked in the wifi config setters


Strycher:
Deferred from #696 — surfaced by the Gemini review on that branch, and deliberately NOT fixed there to keep the change scoped to what the owner approved.

## Finding

`handleSetWifiField` and `handleSetWifiEnabled` in `src/helpers/config/WifiConfigProvider.cpp` ignore the return value of `Preferences::putString()` / `putBool()`:

```cpp
p.putString(field, value);   // result discarded
p.putBool("enabled", enabled);
```

If the NVS write fails, the handler still returns the success ACK (`wifi.ssid = <x>` / `wifi.pwd set (N chars entered)`). The user believes the value is stored; it is lost on reboot, and the failure is invisible — a the standards silent-failure shape.

## Why it was deferred, not fixed in #696

Fixing it changes the user-facing reply strings of `set wifi.ssid` / `set wifi.pwd` / `wifi.enabled`, which is copy the owner has to approve, and it is pre-existing behaviour unrelated to the #692 field bug that #696 exists to fix. #696 fixed the equivalent gap only on the code path it introduced (`wifi clear` now checks `remove()`, using `isKey()` to distinguish an absent key from a genuine erase failure).

## Shape of the fix

Mirror the #696 pattern: check the write result and return an explicit `ERROR:` reply on failure. Note that unlike `remove()`, `putString()`/`putBool()` have no absent-key ambiguity, so no `isKey()` dance is needed.

Also worth auditing the other config providers for the same pattern while in there.

Agent: JadeFox (session c170fdab)

Issue #692, opened 2026-08-13
bug: observer stuck StaConnecting on open (passwordless) public WiFi — SN0MAN field report


Strycher:
Field report relayed by the owner (2026-08-13, Discord). Being handed to a dedicated session — this issue is the complete state; no chat context needed.

## Reporter / device
- Community member **SN0MAN**, deploying an observer at a site with public passwordless WiFi.
- Device: **Heltec V3**, official **offband-v1.3.0 release image** (app device-info reads `1.3.0-1.3.0` — that string is #691's cosmetic bug, NOT a weird build; date 06-Aug-2026 matches the release).
- Reporter has gone home; site access is intermittent. Expect async turnaround on any ask.

## Verified symptom timeline (from _sys channel screenshots)
1. `set wifi.ssid MilanPark` → `wifi.ssid = MilanPark` (5:22 PM)
2. `set wifi.pwd` (no value, attempting open network) → `ERROR: empty value for wifi.pwd` — expected, the setter rejects empty (WifiConfigProvider.cpp:45); harmless, indicates nothing about stored state
3. Rebooted after SSID set (confirmed by reporter)
4. `wifi status` → `wifi.status = StaConnecting` at 5:28 PM and again 5:29 PM
5. Captive portal: reporter says "I don't think so" — **unconfirmed**
6. Owner believes the device **never had a PSK set** — plausible but **unverifiable via CLI** (wifi.pwd is write-only; `wifi status` does not report presence — gap tracked in #689)

## Code facts (verified this session, v1.3.0 ≡ v1.4.0 for all of this)
- Boot: WifiBootstrap::begin reads NVS `wifi/ssid` + `wifi/pwd` (default ""), only gates on **empty SSID**, then `WiFi.begin(ssid, pwd)` — src/helpers/wifi_observer/WifiBootstrap.cpp:126
- Retry: `WiFi.reconnect()` every 10 s **forever**; `StaFailed` is unreachable in practice (WifiBootstrap.cpp:151) — persistent StaConnecting = association never succeeding, with no reason surfaced
- No disconnect-reason logging exists on this path — serial/caplog shows only "STA retry #N"

## Ranked hypotheses (nothing confirmed yet)
[...]

Issue #689, opened 2026-08-13
task: observer needs a 'wifi clear' command — stored PSK cannot be cleared, blocking open-network deployments


Strycher:
Reported via owner (2026-08-13), from community member SN0MAN deploying a node on password-free public WiFi.

**Gap [verified in code]:** there is no way to clear a stored WiFi PSK on an observer:
- `set wifi.pwd` rejects empty values (WifiConfigProvider.cpp:45, deliberate — collides with no-creds detection in WifiBootstrap::begin)
- `wifi` namespace has only `status | enable | disable` (ObserverCli.cpp:608)
- the BLE config wire path (configSet) routes to the SAME handler, so the app cannot clear it either
- only escape today is a full NVS wipe, which destroys the device identity

**Why it bites:** with a stored password the ESP32 STA auth threshold expects a secured AP, so a stale PSK typically prevents association to an OPEN network — a device that ever had WiFi configured cannot be moved to an open-AP site. [hypothesis: auth-threshold behavior is standard arduino-esp32; bench-verify during implementation.]

**Suggested shape (not started):** `wifi clear [pwd|all]` mirroring `mqtt clear <N>`, using `Preferences::remove()` — NOT `putString("")`, per the #98 lesson that empty putString does not reliably clear ESP32 NVS keys. Wire-path parity via the same handler. While here, document open-network support (empty PSK on fresh devices already works — WifiBootstrap only gates on empty SSID) and the captive-portal limitation.

Do not work without owner scheduling.

Agent: WhiteFalcon (session cb12cdd2)

Strycher:
Scope additions from the live field case (#692) — these were discussed in chat but not yet recorded here:

1. `wifi status` should report **pwd presence** — `pwd=(set|unset)`, never the value (wifi.pwd stays write-only). During #692 we could not determine whether a PSK was stored, which is half the diagnosis.
2. `wifi status` should surface the **retry counter** (`StaConnecting (retry 42)`) — the loop retries every 10 s forever and never reaches StaFailed (WifiBootstrap.cpp:151), so "how long has it been failing" is currently unanswerable remotely.
3. Bench verification for this task doubles as #692's hypothesis-3 test: after implementing clear, verify set-pwd → clear → join an open 2.4 GHz AP → StaConnected, on a release-equivalent image.

Field urgency: #692's reporter (SN0MAN) is blocked on exactly this if his device turns out to have a stored PSK. Part of the v1.4.1 patch scope with #691.

[...]

Issue #98, opened 2026-06-12
Epic: non-destructive view of full MQTT broker config (mqtt view <N> / get-all)


Strycher:
## Goal
A **read-only** way to display **all** configurable values of an MQTT broker slot at once — so an operator can SEE the full config without guessing, and **without a destructive `set`**.

## Why
Today you can't see a broker's complete config:
- `mqtt status` shows a curated subset (url, state, transport, auth, cert, `own/eml` Y/N flags) — **not** the actual `username`, `jwt_audience`, `jwt_owner`, `jwt_email`, `topic_prefix`, `iata_override`, etc.
- `get mqtt.broker.<N>.<key>` (#45) reads **one field at a time** — tedious, and you must already know every key name.
- Compounded by #67 (status shows cached `cfg_`, not NVS) — so what little you see can be stale.

This session, debugging eastme.sh / LetsMesh meant guessing at configured values one field at a time. A full, non-destructive dump would have made it trivial — and is needed to confirm the broker pre-config epic (#95) actually applied.

## Proposal (pick the surface)
A read command — e.g. **`mqtt view <N>`** (or `get mqtt.broker.<N>` with no key) — that reads **NVS** and prints **every configurable field** for slot N:
- `enabled, url, transport, port, auth_type, username, topic_prefix, iata_override, ca_cert, jwt_audience, jwt_owner, jwt_email, jwt_refresh`
- **Secrets redacted:** `password` → `(set, len=N)`; never print the raw JWT/password. Owner/email/audience are **not** secrets — show them.
- Read from **NVS**, not cached `cfg_`, so it's authoritative (ties to #67).
- Respect the `_sys` channel-message size limit — compact one-line-per-field, chunked across replies if needed.

## Notes
- Complements per-field `get` (#45) + `mqtt status`.
- Directly aids #95 (pre-config — verify it applied) and #96 (export ≈ a machine-readable view).
- Pairs naturally with fixing #67 (so the view reflects NVS truth).

Strycher:
## #98 landed — `mqtt view <N>` (local branch; NOT pushed/built/flashed)

This is the observability gap you flagged: you can now read a broker slot's **full stored config in one reply**, so you can actually verify a slot instead of inferring from the `Y/N` flags in `mqtt status`.

[...]

PR #713, opened 2026-08-15
task(#696): add wifi clear + surface STA disconnect reason

### LINKED ISSUE #696: task: implement 'wifi clear' + surface STA disconnect reason
Implementation task under #689, scheduled by the owner during the #692 diagnosis session.

## Scope

Two changes, both scoped to observer WiFi config/diagnostics. Neither touches the association decision path.

1. **`wifi clear [pwd|all]`** — clears stored WiFi credentials. Uses `Preferences::remove()`, NOT `putString("")` (per the #98 lesson that empty putString does not reliably clear ESP32 NVS keys). Same handler serves the `_sys`/CLI surface and the BLE config wire path, so the app gets it for free.
2. **STA disconnect-reason logging** — register a `WiFi.onEvent` handler for `ARDUINO_EVENT_WIFI_STA_DISCONNECTED` and surface the `reason` code on the existing retry line. Purely observational.

## Why the second half matters (#692)

Today a failed association surfaces only `wifi.status = StaConnecting` with no reason, forever — `StaFailed` is unreachable (`WifiBootstrap.cpp:151`). The reason code discriminates every open hypothesis on #692 in one reboot:

| reason | points at |
|---|---|
| 201 `NO_AP_FOUND` | SSID mismatch, 5 GHz-only AP, OWE-only AP, or a stale PSK raising the scan threshold |
| 15 `4WAY_HANDSHAKE_TIMEOUT` | matched and attempted; key exchange failed |
| 205 `CONNECTION_FAIL` / 2 `AUTH_EXPIRE` | AP-side refusal |

## Verified premise

#689's body carried the auth-threshold behaviour as `[hypothesis]`. It is now **`[verified: framework source]`** — in arduino-esp32 2.0.17 (`framework-arduinoespressif32 3.20017`, pinned via `platformio/espressif32@6.11.0`), `wifi_sta_config()` (`libraries/WiFi/src/WiFiSTA.cpp:87-108`) sets `threshold.authmode = WIFI_AUTH_OPEN` and only raises it to `_minSecurity` (`WIFI_AUTH_WPA2_PSK`, `WiFiSTA.cpp:118`) when `password[0] != 0`. Nothing in this repo calls `WiFi.setMinSecurity()`. So any non-empty stored PSK filters an open AP out of the scan match entirely — silent, and exactly #692's signature.

Confirmed the pin and `WifiBootstrap.cpp`/`WifiConfigProvider.cpp` are byte-identical between `offband-v1.3.0` and `firmware-base`, so a bench build reproduces the reporter's association path.

## Tests

- Native unit tests for the clear-command parse + wire-path parity.
- Hardware: bench matrix on `hv3-bench` against an open AP (tracked on #692).

## Not in scope

#691, the deliberate `set wifi.pwd` empty-value rejection, and any change to `WiFi.begin` / the retry loop.

Parent: #689. Diagnosis context: #692.

Agent: JadeFox (session c170fdab)

Strycher:
Closes #696. Under #689. Diagnosis context: #692.

## Why

A community observer (Heltec V3, `offband-v1.3.0`) deploying on an open, passwordless public network sits at `wifi.status = StaConnecting` forever (#692). Two firmware gaps made that undiagnosable from the field:

1. **No reason is surfaced anywhere.** `StaFailed` is unreachable in practice (`WifiBootstrap.cpp:151`), so a node that never associates and a node whose PSK was rejected look identical — both are `StaConnecting`, forever, with nothing logged.
2. **A stored PSK cannot be cleared.** `set wifi.pwd ""` is deliberately rejected, so the only escape was a full NVS wipe that also destroys the device identity.

Gap 2 is not cosmetic. `[verified: framework source]` In arduino-esp32 2.0.17 (`framework-arduinoespressif32 3.20017`, pinned via `platformio/espressif32@6.11.0`), `wifi_sta_config()` (`libraries/WiFi/src/WiFiSTA.cpp:87-108`) sets `sta.threshold.authmode = WIFI_AUTH_OPEN` and raises it to `_minSecurity` — `WIFI_AUTH_WPA2_PSK` (`WiFiSTA.cpp:118`) — **only when `password[0] != 0`**. Nothing in this repo calls `WiFi.setMinSecurity()`. So any non-empty stored PSK filters an open AP out of the scan match entirely: the device never attempts association and reports nothing. That is exactly #692's signature.

## What changed

**`wifi clear [pwd|all]`** — clears stored credentials.
- `Preferences::remove()`, **not** `putString("")`, which does not reliably clear an ESP32 NVS key (#98).
- The `remove()` result is checked. `isKey()` distinguishes an **absent** key — a harmless no-op, and the common case on a device that never had a PSK — from an erase that genuinely failed. Reporting success on a failed erase would send an operator away believing a stranded node was rescued.
- Bare `wifi clear` clears the password only; `all` is an explicit opt-in for the SSID too, so a user asking for less is never silently returned to "awaiting setup".
- Reachable from the `_sys` channel CLI and from the config wire path via the `wifi.clear` key, matched **before** the generic `wifi.` prefix so it cannot fall through to the field setter and write an NVS key literally named `clear`.

**STA disconnect-reason surfacing** — `WiFi.onEvent` handler for `ARDUINO_EVENT_WIFI_STA_DISCONNECTED` records the reason; it appears on the retry log line and in the `wifi status` reply. The status reply is the load-bearing half: a remote reporter cannot read a serial log, so `wifi status` is the only instrument they have. Unmapped codes print the bare number rather than a guessed label.

```
before:  wifi.status = StaConnecting
after:   wifi.status = StaConnecting retry=12 last_reason=201(NO_AP_FOUND)
```

That single line separates every open hypothesis on #692 in one reboot: 201 `NO_AP_FOUND` (SSID mismatch / 5 GHz-only / OWE-only / stale-PSK threshold) vs 15 `4WAY_HANDSHAKE_TIMEOUT` (matched, key rejected) vs 205 `CONNECTION_FAIL` (AP-side refusal).

**Not changed:** `WiFi.begin()`, the 10-second retry loop, the state machine, and the deliberate empty-value rejection in `set wifi.pwd`. The association decision path is untouched, so this build reproduces the reporter's behaviour rather than masking it.

## Verification

- **Native tests: 7/7.** New `test/test_wifi_clear/`, run by a dedicated `native_wifi_clear` env — isolated because `WifiConfigProvider.cpp` self-registers a config provider at static init, which would break `test_config_overlap`'s fresh-process precondition. Wired into `run-unit-tests.yml`; without that the test would exist but never run in CI.
- **Existing suites unaffected** — `pio test -e native -e native_kiss_modem` green.
- **Builds: 4/4 SUCCESS** — `Heltec_v3_companion_observer_wifi`, `heltec_v4_companion_observer_wifi`, `heltec_v4_tft_companion_observer_wifi`, `Xiao_S3_WIO_companion_observer_wifi`. That is every env compiling this code, and all four are in the CI matrix.
- **Not hardware-verified yet.** #689 and #692 stay open for the bench matrix on `hv3-bench` against an open AP. The `Preferences::remove()` call itself sits behind `#ifdef ARDUINO`, so the host tests pin the argument contract and reply strings, not the erase.

## Gemini review

Run per the standing gate (`scripts/llm-consult.py --backend gemini --model gemini-2.5-pro`), three passes — the first two responses ended before finishing their category list, so the remainder was re-run rather than counted as passed. Verdict: safe to open as a PR.

| Finding | Severity | Disposition |
|---|---|---|
| `Preferences::remove()` return value ignored | BLOCKER | **Fixed.** Its suggested patch was **not** adopted as written: `remove()` also returns `false` when the key is simply absent (`nvs_erase_key` → `ESP_ERR_NVS_NOT_FOUND`), so that patch would report "NVS write failed" on any device that never had a PSK. Implemented with `isKey()` instead; Gemini confirmed the reasoning on the following pass. |
| `WIFI_REASON_CONNECTION_FAIL` (205) "not defined in IDF 4.4" | MINOR | **Refuted with evidence.** It is at line 121 of the exact SDK header this build compiles against (`framework-arduinoespressif32/tools/sdk/esp32s3/include/esp_wifi/include/esp_wifi_types.h`), alongside 204 and 4. No change. |
| `putString`/`putBool` results unchecked in the pre-existing setters | MINOR | **Deferred to #697.** Fixing it changes user-facing copy on `set wifi.ssid`/`set wifi.pwd` and is unrelated to the #692 bug this PR exists to serve. Filed rather than silently dropped. |
| Missing test for `wifi.clear` dispatch routing | suggestion | **Implemented** — `WifiClearDispatch.WifiDotClearRoutesToClearNotTheFieldSetter` pins the ordering against a future regression. |

## Reviewer notes

- Concurrency: the reason byte is written from the WiFi event task and read from the main loop / `wifi status`. Single byte, `volatile`, atomic on xtensa; a stale-by-one-event read is harmless for a diagnostic.
- Worst-case `wifi status` line is ~86 bytes, inside the 140-byte `_sys` frame limit — no truncation.
- User-facing strings in this PR were reviewed and approved by the owner before commit.

Agent: JadeFox (session c170fdab)

### DIFF
--- a/.github/workflows/run-unit-tests.yml
+++ b/.github/workflows/run-unit-tests.yml
@@ -31,7 +31,7 @@ jobs:
           echo "PIO_SECRETS_FILE=$RUNNER_TEMP/ci-secrets.ini" >> "$GITHUB_ENV"
 
       - name: Run Unit Tests
-        run: pio test -e native -e native_kiss_modem -vv
+        run: pio test -e native -e native_kiss_modem -e native_wifi_clear -vv
 
       - name: Upload Test Results
         # Upload test results even if the test step failed.
--- a/platformio.ini
+++ b/platformio.ini
@@ -224,7 +224,11 @@ build_flags = -std=c++17
   -I src
   -I test/mocks
 test_build_src = yes
-test_ignore = test_kiss_modem
+; test_wifi_clear needs WifiConfigProvider.cpp linked, whose static
+; ProviderRegistrar would register a provider into the process-global table
+; and break test_config_overlap's "fresh process => zero providers"
+; precondition. Same isolation reason as test_kiss_modem: own env below.
+test_ignore = test_kiss_modem, test_wifi_clear
 build_src_filter =
   -<*>
   +<../src/Utils.cpp>
@@ -239,6 +243,28 @@ build_src_filter =
   ; fails to link. (#628)
   +<../src/MeshLog.cpp>
   +<../src/helpers/ConfigSerializer.cpp>
+lib_deps =
+  google/googletest @ 1.17.0
+
+; #696: wifi clear + STA disconnect-reason. Isolated from [env:native]
+; because WifiConfigProvider.cpp self-registers a config provider at static
+; init (see test_ignore above).
+[env:native_wifi_clear]
+platform = native
+test_framework = googletest
+build_flags = -std=c++17
+  -I test/mocks
+  -I src
+  ; WifiObserverConfig.h hard-#errors without one of these (#536); this code
+  ; is observer scope, so OFFBAND_OBSERVER is the honest one to declare.
+  -D OFFBAND_OBSERVER
+test_build_src = yes
+test_filter = test_wifi_clear
+build_src_filter =
+  -<*>
+  +<../src/helpers/config/ConfigDispatch.cpp>
+  +<../src/helpers/config/WifiConfigProvider.cpp>
+  +<../src/helpers/wifi_observer/WifiBootstrap.cpp>
 lib_deps =
   google/googletest @ 1.17.0
 
--- a/src/helpers/config/WifiConfigProvider.cpp
+++ b/src/helpers/config/WifiConfigProvider.cpp
@@ -69,6 +69,70 @@ bool handleSetWifiField(char* reply, size_t reply_size,
     return true;
 }
 
+bool handleClearWifi(char* reply, size_t reply_size, const char* what) {
+    // #689/#696: the only way to un-set WiFi creds. `set wifi.pwd ""` is
+    // deliberately rejected (empty SSID collides with the no-creds
+    // detection in WifiBootstrap::begin), which left a stored PSK
+    // permanently stuck -- and a stored PSK raises the STA scan-auth
+    // threshold to WPA2, so the node can never join an OPEN network
+    // again (#692). Before this, the only escape was a full NVS wipe,
+    // which also destroys the device identity.
+    //
+    // Default (no argument) clears the password only: that is the
+    // recoverable case, and dropping the SSID too would silently send
+    // the node back to "awaiting setup" when the user asked for less.
+    bool clear_ssid = false;
+    if (what == nullptr || what[0] == '\0' || config::strEq(what, "pwd")) {
+        clear_ssid = false;
+    } else if (config::strEq(what, "all")) {
+        clear_ssid = true;
+    } else {
+        snprintf(reply, reply_size,
+                 "ERROR: usage: wifi clear [pwd|all]\n");
+        return true;
+    }
+#ifdef ARDUINO
+    Preferences p;
+    if (!p.begin("wifi", /*readOnly=*/false)) {
+        snprintf(reply, reply_size,
+                 "ERROR: cannot open NVS namespace 'wifi'\n");
+        return true;
+    }
+    // remove(), NOT putString("") -- per #98, writing an empty string does
+    // not reliably clear an ESP32 NVS key, which would leave the very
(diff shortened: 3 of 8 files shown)


Strycher:
Human merge approval received from the owner in session chat (2026-08-15). All 45 checks green including the required `ci-green` aggregator; mergeState `CLEAN`. Enabling auto-merge (rebase) per the CLAUDE-BASE flow — automation merges, the human decides.

Not closed by this merge: **#689** and **#692** stay open pending hardware verification on the bench (`Preferences::remove()` sits behind `#ifdef ARDUINO`, so CI never executed it).

Agent: JadeFox (session c170fdab)

Issue #714, opened 2026-08-15
task: mechanical guards so untested firmware logic cannot ship silently (test-required check + testable-surface ratchet)


Strycher:
Follow-up to #711 / #712. Owner-approved 2026-08-15.

#711 was a regression of #450 that shipped because **nothing asserted the invariant**. #712 fixed that for one invariant and switched on the native suites in CI (161 cases that had never run on a PR). This issue covers the two mechanical guards that generalise it, plus recording the structural pattern that actually moves the needle.

## The measurement that frames this

**~8 of 86 translation units compile natively.** That is the entire testable surface, roughly 10%.

```
[env:native] build_src_filter: Utils.cpp, CaptureRing.cpp, ConfigDispatch.cpp,
ClockSanity.cpp, PrefsLayout.cpp, Packet.cpp, MeshLog.cpp, ConfigSerializer.cpp
vs. 79 .cpp under src/ + 7 under examples/companion_radio
```

`MyMesh.cpp`, both `SerialBLEInterface` implementations, the drains and the CLI are **untestable by construction**. So the failure mode is not "people skip writing tests" -- for most of this codebase there is no test to write without extraction first. Any mechanism that ignores that will just generate noise.

## 1. CI: a source change with no test change must be waived, not ignored

Fail (or warn -- see the open decision) when a PR's diff touches `src/**` or `examples/**` and touches no `test/**`.

**The waiver must be owner-only.** Standing owner decree: only the owner may decide to land without a test. So the escape hatch is a label **only the owner applies** (e.g. `no-test:approved`), never a checkbox an agent can tick and never a magic commit-message string. That is the whole point: it converts the PR-template question from an honour system into a gate, without letting an agent self-authorise.

Known false positives to handle: docs-only, `.github/**`-only, `platformio.ini`/config-only, pure mechanical refactors, and vendored/upstream syncs. Path-exclusion list rather than judgement.

**OPEN DECISION (owner):** land this **advisory first** (annotation, non-blocking) for ~1 week to measure the real false-positive rate, then flip to blocking? Or blocking immediately? Recommendation: advisory first. A gate that cries wolf gets routed around, and routing around gates is the behaviour this whole chain is trying to end.

## 2. CI: the testable surface may only grow

A ratchet asserting the count of natively-compiled translation units never decreases (and that every `test/` suite folder is still discovered by some native env).

[...]

Issue #711, opened 2026-08-15
bug: REGRESSION of #450 — BLE CHUNK frames lose 3 bytes each again; #454 traded the safe constant for a cap that trusts a stale MTU


Strycher:
## Summary

**This is a regression of #450**, reintroduced by #454 (PR #455, commit `53457a0b`).

Over BLE, every full companion frame is delivered 3 bytes short whenever the peripheral's cached ATT MTU is >= 179: the firmware sizes chunks at `MAX_FRAME_SIZE` (176) while the link can only deliver MTU - 3 (173), so the BLE stack silently clips the tail of each notification. That is the identical symptom, magnitude and code path that #450 closed. Confirmed from a tester capture; the arithmetic closes exactly.

## Evidence (tester companion debug log, 2026-08-15)

Three consecutive caplog downloads, byte-identical results each time (deterministic, not random frame loss):

```
Caplog download truncated: 8461 of 8608 bytes in 50 chunks
```

Reconstructed from the client's framed RX dump:

| | |
|---|---|
| START announced total | 8608 |
| CHUNK frames received | 50 (49 x 171 B payload, 1 x 82 B) |
| Client received | 8461 |
| Shortfall | 147 |
| Shortfall / 49 full chunks | exactly 3.0 |

If the firmware wrote 174-byte payloads: `49 x 174 + 82 = 8608`, exactly the announced total. Therefore:

- The firmware's own accounting is correct. It drained the ring completely and its announced `total` was right. Post-download STATUS confirms the ring still holds 8608 of 16384 bytes (`c4 11 00 02 a0 21 00 00 00 40 00 00`).
- The wire delivered 171 B where 174 B were written. `[0xC4][sub] + 174` is a 176-byte frame, clipped to 173, which is MTU - 3 for MTU 176.
- The final 82-byte chunk fits under the limit and arrives intact, which is why the loss is exactly `3 x (number of full chunks)` and not proportional to total size.

For `maxFramePayload(2)` to return 174, `maxFrameSize()` must have returned 176, which requires `_att_mtu >= 179`.

## Regression history (verified from git)

#450 closed exactly this bug, and its own comment states the failure mode precisely. Pre-#454 `caplogDrain()` (`git show 53457a0b^:examples/companion_radio/MyMesh.cpp`):

```cpp
// #450: cap CHUNK data so the frame fits ONE BLE notification. A full
[...]

PR #688, opened 2026-08-13
refactor(#684): decouple WifiConfigProvider wifi.status from observer WifiBootstrap

Strycher:
Child task of **#365** (Feature #295 / Epic #300). Decouples `WifiConfigProvider.wifi.status` from the observer's `WifiBootstrap` machine so the provider links on a non-`OFFBAND_OBSERVER` build — the gap #370 left. Also unblocks **#462** non-observer enablement.

## Change (3 files)
- `WifiConfigProvider.cpp` — `wifi.status` reads `WifiBootstrap` only under `#ifdef OFFBAND_OBSERVER`. Off-observer it derives the **same** vocabulary from stored creds + `WiFi.status()`, gating `StaConnecting` on STA actually being enabled (`WiFi.getMode()`) so an idle-but-configured device (burst-WiFi repeater #301) isn't falsely reported as connecting.
- `WifiConfigProvider.h` — coupling note updated to RESOLVED.
- `variants/heltec_v4/platformio.ini` — `+<helpers/config/*.cpp>` on the non-observer `heltec_v4_companion_radio_wifi` env as the link-proof.

## Contract preserved
Observer `wifi.status` output **byte-identical** (only the guard condition moved). Non-observer emits the same words for shared states, so meshcore-client#375 holds unchanged. `wifi.pwd` write-only untouched.

## Verification
- `heltec_v4_companion_radio_wifi` (**non-observer**) — SUCCESS. Proves the provider links with zero `wifi_observer/` sources.
- `heltec_v4_companion_observer_wifi` (**observer**) — SUCCESS, no regression.
- `RAK_4631_companion_radio_usb` (**nRF52**) — SUCCESS, `config/` not leaked.
- **Gemini 2.5** (`gemini-2.5-pro`) adversarial review — 1 MAJOR (`WL_IDLE_STATUS` false-positive) **fixed** (with a more precise `getMode()` gate than the literal suggestion); rest clean. Log: `docs/llm-consultations/2026-08-13-684-wifi-status-decouple-gemini-gemini-2.5-pro.log`.

## Not in scope
Companion `0xC0` dispatch gating + cap-bit advertise — later #365 children. The providers self-register on the proof env but stay dormant (no wire dispatch) until then.

Closes #684.
Agent: BrownWaterfall (session edc95854)

Issue #683, opened 2026-08-13
bug: Heltec_v2_companion_radio_ble DRAM overflow (616 B) on 1.17.0 base — v1.4.0 release blocker


Strycher:
Both offband-v1.4.0 release runs (rc1 + stable, commit d88ad9b1) failed on exactly one env: Heltec_v2_companion_radio_ble.

```
ld: section .dram0.bss will not fit in region dram0_0_seg
ld: region dram0_0_seg overflowed by 616 bytes
```

[verified:] run 31736395388, job log. All other release envs built. No release object was published from either tag; tags offband-v1.4.0{,-rc1} currently point at a commit that cannot build this env.

[verified:] this env built successfully in the offband-v1.3.0 release run → regression in static RAM between v1.3.0 and the 1.17.0 base merge, on the original-ESP32 (smallest DRAM) board.

Process gap that let it through: env is in release-envs.txt (85) but not ci.yml (44), so ci-green cannot see it — the #350/#463 class of gap, release-matrix flavor.

Decision needed (owner): (a) RAM diet — diagnose which statics grew (map-file diff v1.3.0 vs v1.4.0) and reclaim ≥616 B; (b) owner-approved exclusion of this env from release-envs.txt with documented gate (precedent: in-file gated exclusions → #20/#325); (c) both — diet now, plus decide whether V2 BLE stays supported long-term. Tag handling after the fix also needs an owner call (re-point vs re-cut).

Agent: WhiteFalcon (session cb12cdd2)

Strycher:
DIAGNOSE complete — symbol-level bss diff between full local builds of v1.3.0 and the release commit (script + numbers below).

**Code delta (1.16→1.17 base), net +310 B bss:** `the_mesh` +260 (upstream mesh object growth), `interface_manager` +40, ClockSanity floor cache +4, UIColor +6. The serial→bluetooth interface swap is net-zero (4300 B each way). No single new buffer — the base just grew.

**Why it only kills THIS env:** Heltec V2 is the original ESP32 (smallest DRAM) and the BLE companion carries the NimBLE pools (~3 KB) on top of Offband's diagnostics statics: caplog RAM ring `g_storage` = 16 KB (`OFFBAND_CAPLOG_RING_BYTES`, MeshLog.cpp:21) + CrashLog `s_prev_snapshot` = 4 KB. The env was already within a few hundred bytes of the ceiling; +310 B tipped it. USB/repeater/room_server V2 envs still fit.

[...]

PR #716, opened 2026-08-15
bug(#711): fix the BLE frame-size ceiling, pin it with a regression test, run native tests in CI

### LINKED ISSUE #712: task: automated regression test for the BLE frame-size ceiling (no frame may exceed MTU-3)
Owner directive, 2026-08-15: this class of bug gets an automated regression test so it cannot recur. Parent: #711.

## Why a test, specifically

#450 fixed the BLE 3-byte clip with a constant. #454 replaced that constant with a computed cap and reintroduced the bug **the same day**, issue-to-merge in 11 minutes. It survived a Gemini review that certified the exact two things that failed, and shipped with **zero tests**. Nothing in `test/` references `maxFrameSize`, `maxFramePayload`, or the 173-byte ATT ceiling, so there was no mechanical guard. It then sat undetected on `firmware-base` for over two weeks until a tester could not download a capture.

Prose in a comment did not prevent this and will not prevent the next one. The invariant needs to be executable.

## The invariant to assert

> **No frame handed to `writeFrame()` may exceed the transport's deliverable size.**
>
> For BLE that is `min(MAX_FRAME_SIZE, att_mtu - 3)`, for every value of `att_mtu`, including values larger than `MAX_FRAME_SIZE + 3`.

The regression lives entirely in that last clause. `maxFrameSize()` saturates at `MAX_FRAME_SIZE`, so any `att_mtu >= 179` yields 176 and a chunk builder emits a 176-byte frame onto a link delivering 173. A test that only exercises the low-MTU case (the T1000E 169 path #454 was written for) passes today and catches nothing.

## Proposed test (native, `pio test -e native`)

`test/test_frame_size/` — the repo's existing native googletest harness (MinGW locally, Linux in CI; folder MUST be `test_`-prefixed and the file supplies its own `main()`).

1. **`maxFramePayload` contract, table-driven over `att_mtu`.** A `FakeSerialInterface : BaseSerialInterface` with a settable MTU, asserting `2 + maxFramePayload(2) <= min(MAX_FRAME_SIZE, mtu - 3)` across `{0, 3, 23, 100, 169, 172, 176, 179, 185, 517}`. **179 and 517 are the cases that fail today.** Include the underflow guard cases (`mtu <= 3`) that Gemini caught during #454.
2. **Emitted-frame ceiling.** Drive the chunk builder against the fake, capture every `writeFrame(len)`, assert `len <= deliverable` for the same MTU table. This is the one that would actually have caught #711 rather than only its arithmetic.
3. **Sum-equals-announced.** Assert the delivered byte total equals the START total, which is the exact property the tester's log violated (8461 of 8608).

Extend to every chunked builder that sizes against `maxFramePayload()`, not just caplog: config scalar/stream, observer VIEW/BROKERS, block list, GPS. #453's audit named these; none of them are covered either, and unlike caplog they announce no total so they clip **silently**.

## Acceptance

- Test fails on `firmware-base` as it stands today (proves it catches the live bug).
- Test passes after the #711 fix.
- Runs in CI on every PR, not only locally.

## Notes for whoever takes this

`maxFrameSize()` on nRF52 queries the MTU live rather than caching it, so the fake needs to cover both shapes. See `src/helpers/BaseSerialInterface.h:36-45`, `src/helpers/esp32/SerialBLEInterface.h:105-108`, `src/helpers/nrf52/SerialBLEInterface.cpp:285-290`.

Native harness gotchas are recorded and non-obvious: MinGW is not on PATH by default, and a non-`test_`-prefixed folder is silently collected but never built, reporting "0 test cases" with no error.

Strycher:
## What this changes

Restores the BLE frame-size invariant reported broken in #711, pins it with a test that is proven to fail without the fix, and switches on the native test suites in CI. Affects every BLE companion role (ESP32 and nRF52) plus repo-wide CI.

Closes #712
Addresses #711 — **that issue stays OPEN**: this is bench-verified but has no hardware repro yet.

### The bug

ATT settles the connection MTU at `min(local, peer)`. `begin()` pins our local side with `NimBLEDevice::setMTU(MAX_FRAME_SIZE)`, but `onConnect` cached `getPeerMTU()` unclamped, so a peer advertising more (Android commonly 517) made `maxFrameSize()` saturate at `MAX_FRAME_SIZE` and emit 176-byte frames onto a 173-byte pipe. The stack clips the tail silently: every FULL frame lost exactly 3 bytes, the short final frame survived.

Because our own side is pinned at exactly `MAX_FRAME_SIZE`, the deliverable on this transport is **always at most `MAX_FRAME_SIZE - 3`** — the old `min(mtu-3, MAX_FRAME_SIZE)` ceiling was unreachable by construction.

Field evidence: `8461 of 8608 bytes in 50 chunks`, byte-identical on three consecutive attempts. `49 full chunks x 3 = 147`.

**This is a regression of #450**, reintroduced by #454, which traded an unconditionally-safe constant for a computed cap that is only safe when the cached MTU is truthful — and shipped with no test asserting the ceiling.

## Tests

**Does the test embed the broken behaviour, or has it been watched failing? Both.**

- `RegressionSeventyOneOne_LegacyImplementationIsDemonstrablyBroken` carries the pre-fix formula verbatim and asserts the delta is exactly 3 bytes, so the guarantee runs on every CI run rather than living in a reviewer's memory.
- Also verified by hand: reverting the fix fails **5 of 9** cases; restoring it passes 9/9.

One trap worth flagging for reviewers: the first draft of `EmittedFrameNeverExceedsWhatTheLinkDelivers` **passed against the broken code**, because it derived its expected ceiling from `effectiveMtu()`'s own output — a bug that inflates the MTU also inflates the expectation. It now derives the ceiling from the raw inputs. That is only caught by actually running the test against the broken code.

## Verification

- [x] Builds: `heltec_v4_companion_radio_ble`, `Heltec_v2_companion_radio_ble` (DRAM-tight, **not** in the CI matrix — per #683/#686), `RAK_4631_companion_radio_ble`. `check_esp32_ble_deps.py` OK across 51 envs.
- [x] `pio test` green for every native env: `native` 153/153, `native_kiss_modem` 8/8.
- [ ] **NOT hardware-verified.** No bench repro against a link that actually clips. #711 stays open until someone confirms a clean caplog download on a link where the client's `requestMtu` fails.

## Review

- [x] Gemini-reviewed (`gemini-2.5-pro`). Its one substantive finding — the nRF52 copy of the arithmetic was a divergence risk — is **adopted**: nRF52 now delegates to the same shared function.

Its suggestion to hardcode an nRF52 local MTU is **deliberately declined**, with the reasoning recorded in-code so it is not "fixed" later: that port configures no local MTU, so a guessed value would silently shrink frames on healthy links. `conn->getMtu()` is already the negotiated value there.

Note for reviewers: an LLM review is a second opinion, not a gate. The #454 review certified as "correct" the exact two things that later broke.

## Notes

**This PR changes a required check.** `ci-green` now needs `native-tests`. Consequences:

- The native suites (161 cases across 2 envs) **have never run in CI** — every invariant they encode was unenforced. They are green as of this branch, so this gates on real, passing coverage.
- Envs are **discovered**, not listed: a hardcoded `pio test -e native` silently skipped `[env:native_kiss_modem]`. Discovery finding zero envs is a hard error, because a test job that runs nothing and reports green is worse than no job.
- **Open PRs #688 / #678 / #650 will pick this up on their next rebase** (`strict: true`). Their current runs are unaffected — `pull_request` uses the workflow from the PR head. Owners broadcast separately.

Follow-ups filed, deliberately out of scope here:
- **#714** — mechanical guards (test-required check, advisory first per owner ruling; testable-surface ratchet). Only ~8 of 86 translation units compile natively, so most firmware logic is untestable without extraction.
- Other chunked builders named in #453's audit (config stream, observer VIEW/BROKERS, block list, GPS) size against `maxFramePayload()` too, but announce no total — so they clip **silently**. Not touched here.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,37 @@
+## What this changes
+Brief description, and which role(s)/variant(s) it affects
+(companion / observer / repeater / room / bridge / tooling / docs).
+
+Closes #
+
+## Tests
+
+**If this fixes a bug: does the test embed the broken behaviour, or have you watched it fail?**
+
+A regression test is not verified until it has failed for the right reason. Either:
+- keep the pre-fix behaviour in the test as a reference and assert it differs (permanent, runs on every CI run), or
+- revert the fix, watch the test fail, restore it, and say so here.
+
+Beware the circular assertion: if a test derives its expected value from the thing under test,
+a bug that corrupts that value also corrupts the expectation, and the test passes while the
+product is broken. (#711 -- a regression of #450 that shipped because nothing asserted the
+invariant, and whose first draft test passed against the broken code.)
+
+- [ ] Tests added/updated, and shown to fail without the fix
+- [ ] No test: state why here (only the owner waives a test)
+
+## Verification
+- [ ] Builds affected envs (name them; include any DRAM-tight env not in the CI matrix, e.g. `Heltec_v2_companion_radio_ble`)
+- [ ] `pio test` green for every native env
+- [ ] Hardware-verified, or explicitly stated as NOT hardware-verified
+
+## Review
+- [ ] Gemini-reviewed (`scripts/llm-consult.py --backend gemini`); findings fixed or justified below
+
+Note: an LLM review is a second opinion, not a gate that can be trusted alone. The #454 review
+certified as "correct" the exact two things that later broke.
+
+## Notes
+Anything a reviewer needs: risks, follow-ups filed, what is deliberately out of scope.
+
+Agent: <AMName> (session <uuid8>)
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -43,11 +43,15 @@ jobs:
       - name: Every ESP32 env compiling SerialBLEInterface.cpp declares NimBLE
         run: python scripts/check_esp32_ble_deps.py
 
-  # #712: the [env:native] googletest suites existed but NOTHING ran them in CI, so every
-  # invariant they encode was unenforced on PRs. That is how #711 (a regression of #450)
-  # reached the field: the frame-size ceiling had no mechanical guard, only a comment.
-  # Runs the whole native suite, not just the frame-size one -- 14 suites / 153 cases were
-  # already green locally when this job was added, so it gates on real, passing coverage.
+  # #712: the googletest suites existed but NOTHING ran them in CI, so every invariant they
+  # encode was unenforced on PRs. That is how #711 (a regression of #450) reached the field:
+  # the frame-size ceiling had no mechanical guard, only a comment.
+  #
+  # Envs are DISCOVERED, not listed. There are two native test envs today ([env:native] and
+  # [env:native_kiss_modem], the latter excluded from the former via test_ignore), and a
+  # hardcoded `pio test -e native` silently skipped the second one -- the same "looks
+  # covered, isn't" failure this job exists to end. Any future env with `platform = native`
+  # is picked up automatically.
   native-tests:
     runs-on: ubuntu-latest
     steps:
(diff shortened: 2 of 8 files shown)


Strycher:
Owner approved merge (2026-08-16). Branch brought up to date with a merge commit (12 incoming commits, no file overlap); 46/46 checks green including `native-tests` (174 + 8 + 7 = 189 cases across three envs, now including VioletBarn's `test_broker_rotation`). Enabling auto-merge.

Issue #710, opened 2026-08-15
bug: publish-ring overrun discards packets silently — no counter, no log, and lapped() erases itself


Strycher:
## Question this answers

Does a rotated-out broker actually receive the traffic it missed, or does the ring overflow first?

## Mechanism (works, but undersized)

`MqttBrokerPool::publishPacket` appends **once** to a shared `MqttRingLog`; each broker drains from its **own cursor** in `drainBroker()`, so a rotated-out broker resumes where it left off (#175). Correct design. The problem is depth.

```
MQTT_RING_SLOTS   = 16        // messages, SHARED across all brokers
MQTT_RING_MSG_MAX = 512       // bytes/slot, fixed
```

The ring must span the backlog of the **slowest** reader, which is whichever broker is currently rotated out.

## The arithmetic

With fair rotation (post-#708) across N TLS brokers at `MQTT_ROTATE_DWELL_MS = 60000`, a broker sits out `(N-1) x 60 s`. For the #707 layout (slot 0 plaintext + slots 1-4 TLS), N=4 → **180 s out**.

```
16 messages / 180 s  =  1 packet per 11.25 s  =  5.3 packets/min
```

**Above ~5 packets/min, every rotated-out broker laps and permanently loses the overrun.** The observer publishes *every* received raw packet — adverts, floods, ACKs, telemetry — not just user-visible messages, so this threshold is low for any active mesh.

Measured on hv3-bench (6.9 h soak, 253 CO→UP transitions): TLS handshake is **min 1 s / p50 4 s / p90 7 s / max 8 s**, so a broker also loses ~4 s of its own 60 s turn to handshaking and publishes nothing while `Connecting`.

## Blocker: we cannot currently measure it

`ring_lag` / `ring_lapped` are only emitted in the `mqtt status` broker dump (`ObserverCli.cpp:909`). Nothing periodic reports them, so a long soak produces no evidence either way. On CP2102-bridged boards `pio-flash send` is refused (#468/#503), so the dump is not reachable from tooling at all.

**Step 1 is instrumentation, not a size change:** add `lag`/`lapped` to the 10 s `[rot]` line so soaks become self-measuring. Then size the ring against a real number instead of this estimate.

## Levers (costs against the 14,660 B freed by #701)

| Change | Cost | New threshold (N=4, 60 s dwell) |
|---|---|---|
| ring 16 → 32 | +8,192 B | ~10.7 pkt/min |
[...]

Issue #708, opened 2026-08-15
bug: TLS broker promotion is index-ordered, not the round-robin #177 specifies — slots 3+ starve indefinitely


Strycher:
## Cause

#177 specifies a **round-robin scheduler** ("ring-log + per-broker read cursors (Kafka-offset style) + heap-derived live budget"). What shipped is a **first-eligible-by-slot-index scan** with no fairness state.

`MqttBrokerPool::workerLoop()` promotion loop:

```cpp
for (uint8_t s = 0; s < OFFBAND_MAX_BROKERS; ++s) {
    ...
    if (rotated_out_until_ms_[s] != 0 && cooldown_active) continue;   // skips ONLY the victim
    ...
    if (isTlsTransport(...) && state == Connecting) tls_live++;       // budget gone for higher indices
}
```

The sole fairness mechanism is a cooldown on **the single broker just evicted** (`rotateTlsIfDue`, one victim per dwell). The comment claims this "gives the parked broker first claim" — true only at **exactly two** eligible TLS brokers.

With three or more: victim goes on cooldown, the scan reaches the lowest *other* index first (whose cooldown expired a dwell ago), promotes it, `tls_live++` exhausts the budget, and every higher index is reached with `budget_ok == false`. A slot at index ≥2 can only win if **both** lower slots are cooling simultaneously — impossible, since one eviction happens per dwell and cooldown == dwell.

`biased = now + s * 1000U` is not a fairness device; it's the timestamp passed to `tryConnect` for backoff stagger.

## Evidence

6.9 h soak, hv3-bench, 2478 `[rot]` samples, 0 reboots:

```
ROTATE-OUT s0 : 178      s0/s1 ping-pong for the entire run
ROTATE-OUT s1 : 177
ROTATE-OUT s2 :   1      s2 then HH for 2398 consecutive samples (~6.7 h)
s3/s4/s5      : DN for all 2478 samples (disabled)
```

Slot 2 is MeshMapper. It published nothing for 6.7 hours while the observer reported healthy — `live=1/1`, rotations ticking, no errors.

## Impact

With the #707 layout (slot 0 plaintext + slots 1-4 TLS), **slots 3 and 4 can never be promoted.** Any operator enabling more than two TLS brokers gets silent partial publishing.

## Secondary: `held(no-heap)` is not diagnostic

[...]

PR #724, opened 2026-08-15
fix(#723): resync on first attach only — beta1 regression discarded the publish backlog

### LINKED ISSUE #723: bug: resync on every rotation re-entry discards the publish backlog — observers lose most of their feed (regression in v1.5.0-beta1)
## Cause

REGRESSION introduced in #710 and shipped in `offband-v1.5.0-beta1`.

`reconcileSlot()` sets `pending_resync_[slot]`, and loopTask then calls `ring_.resync(slot)` — which jumps the broker's cursor to the ring head, discarding everything queued for it.

`reconcileSlot()` is called on **every rotation re-entry**, not just first attach:

```
MqttBrokerPool.cpp:722   if (!b.hasClient()) reconcileSlot(pick);
MqttBrokerPool.cpp:438       pending_resync_[slot] = true;
MqttBrokerPool.cpp:148   if (pending_resync_[s].exchange(false)) ring_.resync(s);
```

A rotated-out broker has had `releaseClient()` called, so `hasClient()` is false and it takes this path every single time it comes back.

## Impact

This defeats the entire purpose of the publish ring, stated at `publishPacket`:

> #175: append ONCE. Brokers pull from their own cursor in drainBroker(), so a broker that is down (rotated out, backoff, HeldNoHeap) **resumes where it left off instead of losing the traffic**.

Any observer with 2+ TLS brokers rotating publishes only the packets that arrive during its own dwell window and silently discards the rest — roughly two-thirds of the feed at three brokers. **The drop counter reads zero**, because `resync()` deliberately does not count (it means "deliberately skipped", not "overrun").

## Field report (beta1, Xiao_S3_WIO observer)

> the observer seen the first message on corescope but didn't with the second or third … the companion side of it seen all three messages

Matches exactly: message 1 arrived while the broker was Up and published; 2 and 3 arrived while it was rotated out and were discarded on re-entry. The companion saw all three because BLE is independent of the MQTT path.

## How it got in

#710's review (gemini-2.5-pro round 1) correctly identified that a **newly attached** broker inherits boot backlog, replays stale packets, and pollutes `droppedCount`. The fix was applied in `reconcileSlot()`, which serves both first-attach AND rotation re-entry. Over-applied: a cosmetic metric problem was traded for real data loss.

Neither review round caught it — round 2 reviewed the hand-off for thread-safety, which is correct, not for how often it fires. Hardware caught it, again, and only because a tester was watching CoreScope rather than the device.

## Fix

Resync on FIRST attach only. Per-slot `first_attach_done_` flag; rotation re-entry must preserve the cursor so the backlog drains as designed.

## Acceptance

- A broker rotated out for a full dwell publishes the packets that arrived while it was away.
- Boot backlog is still skipped on first attach (the #710 intent).
- Regression test asserting a rotated-out reader keeps its cursor across re-attach.

Strycher:
Closes #723. **P0 regression shipped in `offband-v1.5.0-beta1`.**

`reconcileSlot()` flagged a ring resync so a newly-attached broker would not replay stale boot backlog (#710). But it also runs on **every rotation re-entry** — a rotated-out broker had `releaseClient()` called, so `hasClient()` is false and it takes that path every time it returns. The resync threw away exactly what the ring exists to hold:

> #175 (`publishPacket`): append ONCE … so a broker that is down (rotated out, backoff, HeldNoHeap) **resumes where it left off instead of losing the traffic**.

At three rotating brokers that discards roughly two-thirds of the feed, and **the drop counter reads zero** — `resync()` means "deliberately skipped", not "overrun", so the loss is invisible to the instrument added to expose it.

**Fix:** per-slot `first_attach_done_` gate. First attach still skips boot backlog (#710's intent); rotation re-entry keeps its cursor.

### Test honesty
The two tests added here exercise `MqttRingLog`, which was never wrong — the ring discarded nothing on its own. **They pass on beta1 code too.** They document intent; they would not have caught this. The defect is in the pool's call *frequency*, which the native harness cannot reach. Same blind spot as #708's regression: tests cover the extracted function, the defect lives in the caller.

### Scope note
This does **not** explain the field report that prompted it. That observer's slot 0 is `tcp`, which is exempt from rotation and never re-enters `reconcileSlot()`. That symptom is still unexplained and needs `mqtt status` from the device — it should not be attributed to this fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,15 @@ Plan-3 web UI, v0.10.x observer multi-broker pipeline, v0.5.0 initial backfill).
 
 ## [1.5.0] - 2026-08-15
 
+### Fixed (beta2)
+- **Rotated-out brokers no longer lose the traffic they missed** (#723): a regression in
+  `beta1`. Re-attaching a broker resynced its ring cursor to the head, which was meant to
+  stop a *newly enabled* broker replaying stale boot backlog — but the same path runs on
+  every rotation re-entry, so a broker returning from its dwell discarded exactly the
+  backlog the ring exists to hold. On an observer with rotating TLS brokers this silently
+  dropped most of the feed, and the drop counter read zero throughout because a resync is
+  a deliberate skip, not an overrun. Resync now happens on first attach only.
+
 ### Added
 - **`wifi clear` and a visible reason for WiFi disconnects** (#696): an observer on an
   open, passwordless network could sit at `StaConnecting` indefinitely with no reason
--- a/src/helpers/wifi_observer/MqttBrokerPool.cpp
+++ b/src/helpers/wifi_observer/MqttBrokerPool.cpp
@@ -431,11 +431,26 @@ void MqttBrokerPool::reconcileSlot(uint8_t slot) {
     const bool may_alloc =
         (tlsClientsAllocated() - (brokers_[slot].hasClient() ? 1u : 0u))
             < OFFBAND_MAX_LIVE_TLS;
-    // #710: the slot is about to (re)attach a client. Ask loopTask to move this
-    // reader to the ring head so it does not replay stale backlog and does not
-    // book boot-time loss as rotation loss. Flag only -- the worker must not
-    // touch ring_.
-    pending_resync_[slot] = true;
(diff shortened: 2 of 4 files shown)


Strycher:

Issue #718, opened 2026-08-15
bug: caplog download corrupts the stream over USB serial — unchecked short write on non-blocking CDC desyncs framing (sibling of #711, NOT fixed by it)


Strycher:
Parent: #1091 (feature: diagnostic log integrity).

Sibling defect to #711, found by DustyBarn during #704 RC32 bring-up. **Different mechanism, different transport, different signature.** #711/PR #716 does not fix this.

## Symptom

`rc32-bench-1`, env `heltec_rc32_companion_radio_usb`, pulled over **USB serial (COM43)** via `scripts/companion_harness.py`. No BLE, no ATT MTU anywhere in the path.

- `CAPLOG_REQ_DOWNLOAD` announced **1067** bytes, the chunk loop received **545**
- and the bytes that did arrive contain **in-line corruption**:

```
[46150] DEBUG: SHT4X not>M-0^@M-D^BI2C address 40
[4615A
```

## Mechanism (confirmed by decoding the corruption)

`cat -v` renders the splice as `>M-0^@M-D^B`. Decoded that is:

```
'>'  0xB0 0x00   0xC4 0x02
 |    |            └── caplog CHUNK code + sub-code
 |    └── length = 0x00B0 = 176
 └── ArduinoSerialInterface frame-start marker
```

That is a **complete, well-formed frame header sitting inside another frame's payload**. The host decoder is correct and cannot produce this on its own: `FrameDecoder.feed()` waits for `3 + length` bytes and emits exactly `length` (`scripts/companion_harness.py:72-92`). The only way a header lands inside a payload is if **the device declared a length it then did not deliver** — the decoder keeps consuming to reach the declared count and swallows the next frame's header as filler, then resyncs mid-stream (hence the sheared `[4615A`).

The cause, on the serial path:

1. `ArduinoSerialInterface::writeFrame` writes the 3-byte header **unconditionally**, declaring `len`, then returns `_serial->write(src, len)` (`src/helpers/ArduinoSerialInterface.cpp:24-37`).
2. `examples/companion_radio/main.cpp:171` sets **`Serial.setTxTimeoutMs(0)`**, making ESP32 USB-CDC writes **non-blocking**: `write()` returns SHORT when the TX FIFO is full rather than waiting.
3. `caplogDrain()` **ignores that return value** and advances the read offset regardless (`examples/companion_radio/MyMesh.cpp:3414-3416`).

[...]

Issue #704, opened 2026-08-15
task: set ENV_INCLUDE_GPS=0 on heltec_rc32 — board has no GPS and its enable pin is the VDD_SPI strap


Strycher:
Implement the fix for **#702**, now that GPIO45 is confirmed as the cause of the standalone boot freeze.

## Confirmed on hardware

Owner tested `heltec_rc32_companion_radio_ble` built with `-D ENV_INCLUDE_GPS=0` (sha256 `b54f21c9…`), running **on battery with no data host attached**:

> **Boots every time.**

That closes the only unproven step in the #702 chain. Previously the board froze on a single short RST and needed repeated/random reset gestures to come up.

## Root cause

`variants/heltec_rc32` defines a GPS the board does not have, and puts its enable line on the **VDD_SPI voltage strapping pin**:

- Heltec's own `RadioCore_Library/src/boards/heltec_rc32.h` defines **no GPS/GNSS pins at all** `[verified]`
- Our variant sets `PIN_GPS_EN=45`, `PIN_GPS_EN_ACTIVE=HIGH`, `ENV_INCLUDE_GPS=1` `[verified: platformio.ini:43]`
- **GPIO45 selects VDD_SPI voltage**: HIGH → 1.8 V, LOW → 3.3 V `[verified: Espressif]`
- The chip has `Embedded PSRAM 8MB (AP_3v3)` — 3.3 V `[verified: esptool]`
- `MicroNMEALocationProvider::begin()` drives it HIGH with **no presence check** `[verified: MicroNMEALocationProvider.h:88-91]`

A reset while GPIO45 is HIGH selects the wrong rail for the flash/PSRAM and the chip cannot fetch code.

Same class as **#211** (RAK3401: GPS probe floated `WB_IO2` = radio power-enable when no GPS present), which was fixed by scoping the probe off.

## Change

Set **`ENV_INCLUDE_GPS=0`** on the `[Heltec_RC32]` base — the board has no GPS, so the entire GPS path is wrong for it, not just the pin polarity.

**Why not invert the polarity** (`PIN_GPS_EN_ACTIVE=LOW`): that would leave a GPS driver constructed and a strapping pin driven for a receiver that does not exist. Removing the path is the correct correction and also drops ~24.5 KB of flash.

**Why not `-U PIN_GPS_EN`**: `variants/heltec_rc32/target.cpp:23` references it unconditionally; the build fails.

[...]

Issue #149, opened 2026-06-17
Heltec V4 observer: GPS position not confirmed (UART GPS not read on observer build)


Strycher:
## Summary
On the **Heltec V4 OLED observer** build (`heltec_v4_companion_observer_wifi`), an attached UART GPS does not surface a position. With the module wired and GPS enabled in settings, the reported position stays at `0,0` — the observer is not pulling a fix into its position telemetry.

Found during the 1.16.0 base-update hardware smoke (2026-06-17, bench device "ST-P", Heltec V4 OLED).

## This is NOT a 1.16.0 regression
- The GPS read/detection code (`EnvironmentSensorManager`) is **byte-identical** between 1.15.0 and 1.16.0 (detection, validity guard, `begin(Wire)`, provider).
- The GPS→position path **exists** on the observer: `MqttPayload` carries `node_lat`/`node_lon` (gated by `loc_valid`), selected exactly as the companion advert path does. The wiring is present; it's the UART read that isn't producing a fix.
- Per the hardware inventory, this device's GPS (M100 Mini / L76K, UART) was only ever verified under **Meshtastic** — **never confirmed under MeshCore**. So this is most likely first-time UART-GPS bring-up on the observer, not a regression.

## Scope / impact — low, non-blocking
The observer is fully functional without GPS:
- Clock comes from **SNTP over WiFi** (GPS-independent; GPS only overrides the clock if it locks).
- MQTT telemetry and mesh function are unaffected.
- GPS on an observer only adds the device's own position dot to a map.

## Likely cause (to investigate on the bench)
UART GPS bring-up on the observer build:
- Boot-time detection: `gps_detected = (Serial1.available() > 0)` — module may not be streaming NMEA during the detection window.
- GPS power enable: `PIN_GPS_EN=34` (active LOW) on heltec_v4 — verify the module is actually powered.
- Wiring to `PIN_GPS_RX=38` / `PIN_GPS_TX=39`, baud, and the L76K/M100 NMEA output.

## Acceptance
HV4 observer reads the UART GPS, acquires a valid fix, and publishes valid `node_lat`/`node_lon` (with `loc_valid=true`), appearing on the position map.

PR #725, opened 2026-08-15
bug(#718): make the framed serial write atomic and honour a refused frame

Strycher:
## What this changes

Makes the framed companion protocol's serial write **atomic without becoming blocking**, and stops `caplogDrain()` skipping bytes a transport refused. Affects every role using `ArduinoSerialInterface` (USB/UART companions) plus the BLE drain path.

Addresses #718 — **that issue stays OPEN**: this is bench-verified but not device-verified.

### The bug

The companion protocol is length-prefixed, but on a companion USB build its transport is deliberately non-blocking: `main.cpp:171` sets `Serial.setTxTimeoutMs(0)` (#149) so a host that stops draining cannot stall the loop and starve BLE servicing. A non-blocking write returns **short** when the TX FIFO is full.

`writeFrame()` wrote the 3-byte header unconditionally, then the payload. On a short payload write the header was already on the wire, so the receiver counted toward bytes that never arrived and consumed the **next frame's header as filler**. `caplogDrain()` compounded it by advancing its read offset regardless, so the skipped bytes were never re-sent.

Observed on `rc32-bench-1` (DustyBarn, #704): a caplog download announced 1067 bytes, delivered 545, with a complete `'>' 0xB0 0x00 0xC4 0x02` frame header spliced inside a payload.

Dropping bytes is **correct for the debug mirror** and **corrupting for a length-prefixed protocol**. That is the whole defect.

### The fix

- `writeFrame()` checks `availableForWrite()` and returns **0 without touching the wire** if the whole frame will not fit. It never returns a short count. New contract documented on the declaration: **0 means NOT SENT, retry**.
- `caplogDrain()` treats 0 as a refusal and does **not** advance `_caplog_off`, re-offering the same bytes next pass. The ring is frozen for the download and ERASE is already rejected while streaming (verified in `handleCmdFrame`), so the retry serves identical data.
- The END terminator is retried rather than tearing the stream down, which would have left the client waiting for a terminator never re-sent.
- **Stall abort:** refusing without advancing means a host that never drains would keep the stream open forever — and `_caplog_streaming` also gates ENABLE/DISABLE/ERASE/DOWNLOAD, wedging the subsystem. Aborts after `CAPLOG_STALL_ABORT_MS` (15 s) measured from the **last accepted chunk**, so a slow-but-advancing drain is never cut off, and the device frees its state before the client's 30 s timeout.

This also closes, **for both transports**, the hole recorded in #711: BLE `writeFrame()` returns 0 on a full send queue and the drain was skipping those bytes too.

## Tests

**Has the test been watched failing? Yes — 3 of 5 cases fail against the unfixed code, 5/5 pass with it.**

`test/test_serial_framing` drives the **real** `ArduinoSerialInterface` against a `Stream` that short-writes like a full FIFO, then walks the emitted bytes with an **independent** decoder (`wireIsWellFramed`) to prove no partial frame ever reached the wire. The decoder is a separate implementation from the encoder under test, so the check is not circular.

Required adding `ArduinoSerialInterface.cpp` to `[env:native]` and promoting the `MockSerial` mock from `Print` to `Stream` — the extract-to-test pattern, raising the natively-tested surface by one more unit.

## Verification

- [x] Builds: `heltec_rc32_companion_radio_usb`, `heltec_v4_companion_radio_ble`, `RAK_4631_companion_radio_usb`, `Heltec_v2_companion_radio_ble` (DRAM-tight, **not** in the CI matrix)
- [x] `pio test` green for every native env: 149 / 8 / 7
- [ ] **NOT hardware-verified.** That the short write produced the rc32 capture is the best-supported reading and no competing mechanism survives it, but it is untested on device. DustyBarn holds the reproducer (`pull_ring.py`) and has been asked to re-run it once this lands.

## Review

- [x] Gemini-reviewed (`gemini-2.5-pro`).

It identified the **starvation stall as a high-severity defect introduced by the fix** — correct, and the reason `CAPLOG_STALL_ABORT_MS` exists. Its documentation points (return contract, `availableForWrite()` dependency) are adopted, and its "should be impossible" note on the short-header branch is softened to *rare*, since the capacity check is an unlocked snapshot and the debug console shares the Stream on some roles. Its ERASE-mid-stream concern was checked and is already guarded.

## Notes

**Behaviour change worth flagging:** fire-and-forget callers of `writeFrame()` now silently drop a frame under back-pressure instead of corrupting the stream. That is strictly the better failure — corruption is persistent and cascading, a dropped reply is transient — but any caller that needs guaranteed delivery must now retry on 0.

**New hard dependency:** the bound `Stream` must implement `availableForWrite()` meaningfully. Base `Stream` returns 0, which reads as "no room" and would refuse forever. Every transport actually bound here implements it; documented on the declaration.

Not in scope: DustyBarn's separate ungated-console-writer concern for shared-UART roles (they are filing it under #704), and their finding that the ring holds nothing before ~45 s at boot.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -127,6 +127,12 @@ namespace offband { MqttBrokerPool& wifiObserverPool(); }
 #define CAPLOG_RESP_ACK               0x10  // control:  [0xC4,0x10, req_op, ok(0|1)]
 #define CAPLOG_RESP_STATUS            0x11  // status:   [0xC4,0x11, enabled, level, used(4B LE), cap(4B LE)]
 
+// #718: abort an in-flight download after this long with the transport refusing every
+// frame. Measured from the last ACCEPTED chunk, so a slow but advancing drain never
+// trips it. Shorter than the client's 30 s downloadCaplog timeout so the device frees
+// its own state first and the client sees a clean stop rather than a wedged radio.
+#define CAPLOG_STALL_ABORT_MS         15000UL
+
 // Stats sub-types for CMD_GET_STATS
 #define STATS_TYPE_CORE               0
 #define STATS_TYPE_RADIO              1
@@ -1279,6 +1285,7 @@ MyMesh::MyMesh(mesh::Radio &radio, mesh::RNG &rng, mesh::RTCClock &rtc, SimpleMe
   _caplog_streaming = false;  // #396: no caplog download in flight
   _caplog_off = 0;
   _caplog_resume = false;
+  _caplog_last_progress = 0;  // #718: stall clock
   offline_queue_len = 0;
   app_target_ver = 0;
   clearPendingReqs();
@@ -1833,6 +1840,7 @@ void MyMesh::handleCmdFrame(size_t len) {
       _serial->writeFrame(out_frame, 6);
       _caplog_off = 0;
       _caplog_streaming = true;
+      _caplog_last_progress = millis();   // #718: stall clock starts now
       return;
     }
     // unknown sub-code -> explicit error (never silently fall through to download)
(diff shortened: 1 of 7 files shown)


Strycher:
Owner approved merge (2026-08-16), landing after #716 as sequenced. Branch updated with a merge commit; `MyMesh.cpp` overlapped and auto-merged cleanly (different regions) — verified both fixes coexist in `caplogDrain()`: the #711 `maxFramePayload` sizing and this branch's refusal handling + stall abort.

46/46 green. **`native-tests` ran this PR's own regression suite for the first time** — `test_serial_framing` PASSED, 179 + 8 + 7 = 194 cases across three envs. Enabling auto-merge.

Issue #727, opened 2026-08-15
task: store the packet in the publish ring, not the rendered JSON — returns ~12.8 KB with no consumer-visible change


Strycher:
## Proposal

**Store the packet in the publish ring, not the rendered JSON.** Render the `/packets` body at drain time instead of at receive time.

This gets the memory back with **zero consumer risk** — every broker receives a byte-identical payload. No contract change, no coordination with any consumer, no flag.

Supersedes the original framing of this issue, which proposed dropping unread fields from the payload. That is still worth doing eventually but is a *separate*, genuinely optional follow-up (§4) — it changes a public payload and remains blocked on confirmation.

Background and evidence: `docs/architecture/2026-08-15-observer-mqtt-payload-assessment.md`.

## 1. Why the ring is expensive

`publishParsedPacket()` renders the full `/packets` JSON and hands it to `publishPacket()` → `ring_.append()`. Every ring slot is sized for the **largest possible rendered body**, and the body embeds `"raw":"<whole packet in hex>"` — hex doubles the packet, then ~316 B of fields are added on top.

```
worst-case rendered body : ~826 B   (255 B packet -> 510 hex + ~316 B fields)
current ring             : 20 slots x 1024 B = 20,480 B
```

That sizing is what forced #726 (`MQTT_RING_MSG_MAX` 512 → 1024, slots 32 → 20) and it consumed most of the 14,660 B the #701 heap trim freed — net savings fell to roughly 2,250 B.

## 2. What changes

Store a compact record and render per broker at drain:

```
raw packet bytes            up to 255 B
raw_len                            2 B
rx_time (unix, captured AT RECEIVE) 4 B
rssi, snr, score, duration        ~14 B
packet_type, payload_len, route    3 B
path_info                         ~16 B
                                 ------
worst case                       ~294 B  -> ~320 B slots
```

```
now      : 20 slots x 1024 =  20,480 B
proposed : 24 slots x  320 =   7,680 B     ~12,800 B returned, AND more depth
```

Hex encoding and JSON assembly move from `publishParsedPacket()` (once per received packet) to `drainBroker()` (once per broker per message). With `OFFBAND_MAX_LIVE_TLS = 1` plus an always-on plaintext primary, that is ~2 renders instead of 1 — negligible next to a TLS handshake.

**The published bytes are unchanged.** Same topic, same keys, same values.

## 3. Design constraints

[...]

Issue #726, opened 2026-08-15
bug: parsed /packets JSON is built at 1024 B but the ring caps at 512 — every packet over ~98 bytes silently dropped since #175


Strycher:
## Cause

`publishParsedPacket()` builds the `/packets` JSON into a **1024-byte** buffer and hands it to `publishPacket()`, which appends it to the ring. The ring rejects anything over **512 bytes** — silently.

```cpp
// MqttBrokerPool.cpp
char json[1024];
int n = buildPacketJson(json, sizeof(json), packet, ...);
if (n <= 0 || (size_t)n >= sizeof(json)) return 0;
return publishPacket(json, (size_t)n);        // -> ring_.append(payload, n)

// MqttRingLog.h
if (payload == nullptr || len == 0 || len > MQTT_RING_MSG_MAX) return 0;   // 512
```

`append` returns 0, `publishPacket` returns 0, and the packet is discarded. **No log, no counter, no state change.** `droppedCount` does not see it — that counts cursor overrun at `commit()`, not a rejected append.

## The size at which it bites

The JSON embeds `"raw":"<entire packet, hex>"`. Measured from the format string:

```
JSON skeleton + non-raw values : 316 B
ring limit                     : 512 B
budget for raw hex             : 196 B  ->  packet ceiling ~98 BYTES

 40 B packet -> JSON 396 B  OK
 81 B packet -> JSON 478 B  OK
100 B packet -> JSON 516 B  DROPPED
255 B packet -> JSON 826 B  DROPPED   (max MeshCore packet)
```

**Every packet larger than ~98 bytes has been silently dropped on the `/packets` path.** Transport-independent — a plaintext tcp broker is affected identically, because the drop happens before any broker fan-out.

## Why this reaches CoreScope

The observer publishes two topics, and they are NOT equivalent:

| Path | Topic | JSON key for hex | Buffered? |
|---|---|---|---|
| `publishRawFromBytes` | `/raw` | `"data"` | no — direct to `Up` brokers only |
| `publishParsedPacket` -> `publishPacket` | `/packets` | `"raw"` | **yes — the ring** |

CoreScope's ingestor reads `msg["raw"]` (`cmd/ingestor/main.go:551`). The `/raw` topic uses `"data"`, so it does not match. **CoreScope is fed by `/packets`** — the path with the ceiling.

## Field report this explains

[...]

PR #728, opened 2026-08-15
docs(#726/#727): observer MQTT payload assessment

Strycher:
Design-of-record for the #726 investigation and the #727 proposal. Docs only.

Captures:
- The observer publishes **two non-equivalent topics**. `/raw` puts hex under `data`; CoreScope gates on `msg["raw"]`, so `/raw` is never ingested — we publish every packet twice and the primary consumer reads one.
- CoreScope decodes `raw` itself and reads **none** of `packet_type`/`payload_len`/`route`/`path`/`hash`/`len`/`time`/`date`/`duration`/`origin_id`. Observer identity comes from the topic segment, not `origin_id`. (Verified by the CoreScope maintainer session with file:line refs.)
- No observer metadata requires a decode: `score = f(snr,len)`, `duration = f(len)`, the rest is radio or identity.
- **§7a — the actual fix:** the ring stores *rendered JSON*, so slots are sized for the worst-case ~826 B body. Storing the packet record and rendering at drain returns ~12,800 B and increases depth, with **byte-identical published output** — so it needs no consumer confirmation.

Records two things deliberately rather than smoothing them over:
- The CoreScope session's own caveat that VCR/audio-lab/analytics were **inferred from architecture, not traced line-by-line** — which matters because that inference is the entire basis for ever removing fields.
- That CoreComms could **not** be verified (SPA HTML on `/api/health`; `eastmesh.au` 403, probed once, no evasion attempted), and the owner ruling that no field is dropped without their confirmation.

### DIFF
--- a/docs/architecture/2026-08-15-observer-mqtt-payload-assessment.md
+++ b/docs/architecture/2026-08-15-observer-mqtt-payload-assessment.md
@@ -89,8 +89,11 @@ A whole-repo grep for `packet_type`, `payload_len`, `route`, `path`, `hash`, `le
 
 CoreScope subscribes `meshcore/#` (`main.go:150`) so it *receives* `/raw`, but the entire
 packet branch is gated on `rawHex, _ := msg["raw"]; if rawHex != ""` (`main.go:650-651`).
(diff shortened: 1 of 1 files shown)


Strycher:

PR #731, opened 2026-08-16
refactor(#727): slim /packets to what consumers read; ring 40x288; rx_time observable

### LINKED ISSUE #727: task: store the packet in the publish ring, not the rendered JSON — returns ~12.8 KB with no consumer-visible change

Strycher:
Closes #727 (§4). **Third in the stack** (base = task/727-ring-stores-record).

CoreScope — and every fork, CoreComms/EastMesh confirmed from public source — decodes `raw` itself and reads none of packet_type/payload_len/route/path/hash/len/time/date/duration/origin_id. Dropped from the emitted body and the record:

- `PacketRecord` 296 → **276 B measured**; on-device SHA256 + packet-parse **removed** (dead work, not hidden).
- Emitted body: origin, timestamp, type, direction, raw, SNR, RSSI, score — exact names CoreScope reads.
- Ring 40 × 288 = 11,520 B (owner-set depth; fleet data put the busiest real observer at ~20 slots sustained).
- **rx_time observability** (SAFELANE no-silent-failures): `[ring] backlog sN age=Ns (publishing receive-time)` fires when a rotated-out broker publishes a queued record — proves receive-time is carried.
- /raw untouched (documented map-integration data).

Evidence: `docs/architecture/2026-08-15-observer-mqtt-payload-assessment.md` §7b/§7c. Tests 16/16 ring + 9/9 rotation. Full-day net: 126,000 → 114,916 B, ceiling gone, 40-deep ring.

### DIFF
--- a/src/helpers/wifi_observer/MqttBrokerPool.cpp
+++ b/src/helpers/wifi_observer/MqttBrokerPool.cpp
@@ -22,6 +22,10 @@ static_assert(sizeof(offband::PacketRecord) <= MQTT_RING_MSG_MAX,
 
 namespace offband {
 
+#ifndef MQTT_ROTATE_DWELL_MS
+  #define MQTT_ROTATE_DWELL_MS 60000U   // #175: how long a TLS slot holds the budget
+#endif
+
 // ---------------------------------------------------------------------------
 // begin
 // ---------------------------------------------------------------------------
@@ -247,6 +251,25 @@ uint8_t MqttBrokerPool::drainBroker(uint8_t slot) {
         if (len != sizeof(PacketRecord)) { ring_.commit(slot); continue; }
         PacketRecord rec;
         memcpy(&rec, buf, sizeof(rec));
+#if defined(ARDUINO)
+        // #727 sec4: rx_time observability. If this record was queued while the
+        // broker was rotated out, it is being published LATER than it arrived --
+        // and it must still carry its RECEIVE time, not "now". Emit the lag so a
+        // backlogged delivery is a VISIBLE event and the receive-time guarantee
+        // is provable on serial instead of silent. Rate-limited to one line per
+        // drain pass (the oldest record in the burst).
+        {
+            uint32_t nowsec = (uint32_t)time(nullptr);
+            uint32_t age = (nowsec > rec.rx_time) ? (nowsec - rec.rx_time) : 0;
+            static uint32_t s_last_backlog_ms = 0;
+            if (age >= (MQTT_ROTATE_DWELL_MS / 1000U) &&
+                (uint32_t)millis() - s_last_backlog_ms >= 10000U) {
+                s_last_backlog_ms = (uint32_t)millis();
+                Serial.printf("[ring] backlog s%u age=%us rx=%u (publishing receive-time)\n",
+                              (unsigned)slot, (unsigned)age, (unsigned)rec.rx_time);
+            }
+        }
+#endif
         char json[1024];
         int n = buildPacketJsonFromRecord(json, sizeof(json), rec,
                                           /*is_tx=*/false, ctx);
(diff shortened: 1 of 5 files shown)


Strycher:

Issue #761, opened 2026-08-17
task: heap readout should be opt-in, not on by default on dev builds


Strycher:
Owner-directed 2026-08-17: get the `Heap:NNNNN` readout off the screen.

## What it actually is (correcting the framing it was raised with)

It is **not** an Observer feature. `examples/companion_radio/ui-new/UITask.cpp:553-573` attributes
it to the **ESP32 V3/V4 heap crisis** (#8), retained as a debug aid by #153.

It is **not** on all screens — it renders on the **home page only**, at logical y=56, centred.

It is **not** in shipped releases. It is already gated by `offbandIsDevBuild()`, which is true only
when `OFFBAND_VERSION` carries a `-g<sha>` or `-dirty` suffix. A clean tagged release — including
`-rc` — never shows it. It is visible on the owner's bench because every bench build is a dev build
by definition.

So the actual gap: there is **no way to turn it off on a dev build**, which is precisely where
someone is doing UI work and does not want a debug readout occupying the bottom line.

## Why not simply delete it

The RadioCore bring-up makes this the wrong moment to lose the tool. #624 (RCC6 / ESP32-C6) and
#625 (RC52 / nRF52840) are both boards where **heap is the open question** — the C6 reportedly runs
91 KB free / 17 KB minimum against a 56 KB display back buffer. An on-screen free-heap readout with
no serial attached is exactly the instrument that bring-up wants, and on this board class attaching
a serial console perturbs the thing being measured (#756).

## Proposal

Replace the implicit dev-build gate with an **explicit opt-in**, default OFF:

- `-D OFFBAND_SHOW_HEAP` -> readout appears
- absent -> never appears, on any build, dev or release

Net effect: gone from every screen the owner sees, still one build flag away when a heap
investigation needs it. Also frees the bottom line, which matters at the larger font sizes under
#758 where an 18px cell at logical y=56 reaches physical 130 against a 128px panel.

## Blast radius

`ui-new/UITask.cpp` is shared by all companion envs using the new UI, so this is **not** RC32-only —
it removes the readout from every companion board that currently shows it. That is the intent, but
it should be stated rather than discovered.

## Acceptance

- No heap readout on any default build, dev or release.
- `-D OFFBAND_SHOW_HEAP` restores it unchanged.
- No other home-page content moves.
- Mono / e-ink / no-display envs unaffected.

Agent: BrightCat (session 5970917d)

Strycher:
[...]

Issue #758, opened 2026-08-17
task: antialiased JetBrains Mono text on the RC32 panel


Strycher:
Parent: #743. Owner-approved 2026-08-16: **antialiased, JetBrains Mono.**

This is **#749's Blocker 3**, which that task deliberately deferred.

## The actual defect

`font5x7` is a 5x7 bitmap and `drawChar()` scales each font pixel into a rectangle via
`textPixelScaleX/Y` `[verified: NV3001BDisplay.cpp:398-410]`. At size 1 that is **1 wide x 2
tall** (`NV3001B_TEXT_SIZE1_SCALE_Y 2`).

So a character occupies **5x14 physical pixels but carries 7 rows of information** — every row
is simply doubled. The blockiness is not the font being small; it is the font being *stretched*.
A native-height face fits the same footprint with twice the real detail.

## Why antialiasing is possible now and was not before

Blending a glyph edge requires reading the pixel underneath it. Writing straight to the panel
that is a read-back over SPI — prohibitive. **#747's back buffer makes the destination readable
RAM**, so blend-against-background is a couple of instructions in the existing draw path. This
task is a direct dividend of #748 and could not have been done before it merged.

## Font choice — JetBrains Mono

**SIL OFL 1.1**, so embedding is fine, and it is the face the Offband wordmark is set in
(`offband-site/static/img/logo-lockup-*.svg` sets `offband` in JetBrains Mono 500). Splash text
will match the logo directly above it.

⚠ **Do not ship the comparison mock-up's font.** That render used Consolas purely because it was
on the dev machine; it is a Microsoft font and is not redistributable.

## Cost

| | |
|---|---|
| 4bpp glyphs, 95 printable ASCII, ~8x16 | **~6 KB** flash (of 6,553,600) |
| Scope | `src/helpers/ui/NV3001BDisplay.cpp` only — the other 11 drivers keep `font5x7` |

## The risk worth planning for

`getTextWidth()` currently returns `(len * 6 * scale_x) / DISPLAY_SCALE_X` — a **fixed-advance**
calculation. Every centring and ellipsis decision in the UI is built on it
(`drawTextCentered`, `drawTextRightAlign`, `drawTextEllipsized` in `DisplayDriver.h`). Changing
glyph metrics ripples into every screen's layout on this board. Keeping a **fixed advance**
(monospace, which JetBrains Mono is) contains that blast radius — a proportional font would not.

## Acceptance

[...]

Issue #757, opened 2026-08-17
task: dark UI theme for RC32 + Offband-green title banner


Strycher:
Parent: #743. Owner-directed 2026-08-16, following the #749 splash landing on `rc32-bench-1`.

The colour splash (#749) is owner-verified on the panel and looks right. The owner's call after
seeing a dark mock-up: **invert the whole RC32 UI to a dark theme, and replace the blue title
banner with Offband green.**

## Why this is its own task, not #749 scope

#749 delivers a colour-image primitive and a native-resolution draw path. This changes the
**palette every screen renders in** — a different concern, and one that touches screens #749
never goes near.

## Scope is naturally one board

`UIColor`'s statics are defined **per driver** — all 12 display drivers carry their own copy
`[verified: 12 files define ColorVal UIColor::window_bkg]`. So this is confined to
`src/helpers/ui/NV3001BDisplay.cpp` and cannot affect the ~115 mono-OLED envs, the e-ink
variants, or any other panel.

## The nine roles in play

Usage counts across `src/` and `examples/`: `primary_txt` 56, `secondary_txt` 28,
`window_bkg` 27, `corp_blue` 25, `warning_txt` 22, `title_txt` 15, `title_bkg` 15,
`popup_txt` 14, `popup_bkg` 14.

Current NV3001B values are a light theme: `window_bkg` white, `primary_txt` black,
`title_bkg` `0x001F` (pure blue — the banner in question), `popup_bkg` `0x07FF` (cyan).

The banner is `fillRect(0, 0, width, 12)` in `title_bkg` with `title_txt` over it
(`ui-new/UITask.cpp:309`). Note `UITask.cpp:324` branches on `title_bkg == window_bkg`, which
must keep working — it is how mono panels avoid an invisible page indicator.

## Colours to draw from (brand assets, not invented)

From `offband-site/static/img/logo-lockup-dark.svg` and `-light.svg`:

| Brand colour | Hex | RGB565 | Role in the lockup |
|---|---|---|---|
| ink | `#0F1412` | `0x08A2` | wordmark on light |
| paper | `#EFF3E8` | `0xEF9D` | wordmark on dark |
| deep green | `#1A7A44` | `0x1BC8` | mark on light |
| mid green | `#4DC580` | `0x4E30` | arcs on dark |
| bright mint | `#7BEFA8` | `0x7F75` | dots on dark |

## Acceptance

[...]

Issue #747, opened 2026-08-16
task: NV3001B back buffer -- blit in endFrame instead of erasing the panel live


Strycher:
Parent: #743. Implementation-only task; hardware verification is tracked on the epic.

`NV3001BDisplay::startFrame()` blanked the physical panel and `endFrame()` was empty, so the RC32
erased and redrew in front of the user on every UI update.

`[verified]` Every sibling driver already buffers — `SSD1306Display` (115 envs), `SH1106Display`
(13), `ST7735Display` (15), `ST7789Display` (3), `LGFXDisplay` (1). `NV3001BDisplay` and
`ST7789LCDDisplay` (14 envs) were the two exceptions.

## Approach

`fillPhysicalRect()` is the **only** function that touches the panel, so intercepting that one
function buffers every draw — including `drawChar()`, which calls it once per lit font pixel and
previously cost up to 35 SPI transactions per character.

- 220 x 128 x 2 = 56,320 B, heap-allocated so static RAM is unchanged (124,776 B).
- **PSRAM first** — the RC32 has 8 MB embedded, and 8 MHz SPI (~1.1 MB/s) is the blit bottleneck,
  not PSRAM read speed, so internal DRAM stays free. Falls back to internal, then to the original
  direct path.
- Pixels stored **byte-swapped**, so the blit is a plain `writeBytes` with no per-pixel conversion.
- Allocation failure is a **supported state**: degrades to direct-to-panel and says so on Serial,
  so a failed alloc can never be misdiagnosed as "the fix did not work".
- `clear()` blits its own result — it is standalone, not inside a `startFrame`/`endFrame` pair,
  so it would otherwise draw into RAM and never appear.

## Scope

One driver. No API change, no call-site change, **no screen change**. `NV3001BDisplay` is used by
one env, so blast radius is RC32 only.

## Acceptance (this task)

- `heltec_rc32_companion_radio_usb` and `heltec_rc32_repeater` build clean. ✅

Visible-flicker confirmation on `rc32-bench-1` is the epic-level acceptance, not this task.

Agent: BlueElk (session f148bd05)

Strycher:
Implemented. `fillPhysicalRect()` writes into a 56,320 B PSRAM-first heap buffer and `endFrame()` blits it in one transfer; static RAM unchanged. Allocation failure degrades to the original direct path and reports it, so it can never be misread as the fix failing.

[...]

PR #762, opened 2026-08-17
epic(#743): RC32 colour splash, dark Offband theme, and antialiased JetBrains Mono

### LINKED ISSUE #749: task: colour-image primitive + native-resolution splash path (enables a designed splash)
Parent: #743. **Enabling work — this is what has to exist before any designed asset can be shown.**

A design session is being spun off to produce a better colour splash graphic. Today that asset
literally cannot be drawn. Three verified blockers.

## Blocker 1 — there is no colour-image primitive

`DisplayDriver` has exactly one image call
`[verified: src/helpers/ui/DisplayDriver.h:44]`:

```cpp
virtual void drawXbm(int x, int y, const uint8_t* bits, int w, int h) = 0;
```

**XBM is 1-bit monochrome.** The current splash is `offband_mark` + `offband_wordmark` XBMs
(`examples/companion_radio/ui-new/offband_logo.h`, generated by `scripts/gen-offband-logo.py`,
from #153). There is no way to draw a colour image at all.

### API decision to make

Add something like:

```cpp
virtual void drawRGB565(int x, int y, const uint16_t* px, int w, int h);
```

**Recommendation: give it a default (non-pure-virtual) implementation so the 115 mono-OLED envs
need no changes.** Making it pure virtual forces an implementation into every driver in the tree
for no benefit. Options for the default body, in preference order:

1. **No-op** — colour art simply does not appear on mono panels; those boards keep their XBM
   splash. Simplest, zero risk.
2. **Threshold/dither to 1-bit and forward to `drawXbm`** — one asset everywhere, but dithered
   art on a 128x64 OLED will likely look worse than the purpose-made XBM.

Prefer (1) and keep the XBM path as the mono splash.

## Blocker 2 — the splash draws in logical space, which is not the panel resolution

| | |
|---|---|
| Physical panel | **220 x 128** landscape (memory 128x220, rotated) |
| Logical UI canvas | **128 x 64** — what every screen draws in |
| Scale | **1.72x horizontal, 2.0x vertical — non-uniform** |

`[verified: NV3001B_LOGICAL_WIDTH/HEIGHT vs NV3001B_SCREEN_WIDTH/HEIGHT, DISPLAY_SCALE_X/Y]`

The whole UI is authored for MeshCore's baseline 128x64 mono OLED and stretched. A full-res asset
drawn through logical coordinates gets scaled **non-uniformly** — circles become ellipses.

So the splash needs a path that addresses the panel at native resolution, bypassing
`scaleX()`/`scaleY()`, without disturbing how every other screen renders.

## Blocker 3 — the font is a 5x7 bitmap, upscaled as pixel blocks

`font5x7` (`NV3001BDisplay.cpp:149`) is drawn by `drawChar()` scaling each font pixel into a
rectangle via `textPixelScaleX/Y`. That blockiness is exactly why the UI reads as dated.

**Deliberately NOT in this task.** A real font renderer is either Adafruit GFX free fonts, a
second larger bitmap font for the splash only, or LovyanGFX — and LovyanGFX would solve blockers
1 and 3 together. That decision belongs with the port assessment on #743, not buried here.

## Asset budget, for the design brief

- Canvas **220 x 128**, RGB565.
- Full-screen uncompressed = **56,320 B (55 KB)** of flash. Affordable (companion is 670 KB of
  6,553,600) but a logo region beats full-bleed; RLE if it grows.
- Pipeline: extend `scripts/gen-offband-logo.py` to emit an RGB565 C array from PNG.

## Constraints that must survive

- **MeshCore attribution stays** (#153). MIT fork obligation, not a style choice.
- **Do not duplicate the splash again.** #705 (open, P1) exists precisely because there are
  already five splash implementations, four shipping MeshCore branding. This work should land in
  the single shared component #705 produces, or explicitly sequence behind it.
- Mono panels (115 SSD1306 + 13 SH1106 envs), e-ink, and 27 no-display variants must keep working
  unchanged.

## Acceptance

- A colour RGB565 asset renders at native 220x128 on `rc32-bench-1`, undistorted.
- Mono/e-ink/no-display envs build and behave exactly as before.
- No screen other than the splash changes behaviour.

Agent: BlueElk (session f148bd05)

Strycher:
Parent epic: **#743**. One epic branch, one PR, per CLAUDE-BASE. Closes #749, #757, #758, #761.

Everything here is confined to the RadioCore RC32 by construction. **The ~115 mono-OLED envs, the
e-ink variants and the 27 no-display variants are untouched** — verified, not asserted (see
Evidence).

## What landed

| # | Change |
|---|---|
| **749** | `DisplayDriver::drawRGB565()` — colour-image primitive, **defaulted to a no-op** so the other 11 drivers need no edit; `Rgb565Blit.h` (pure `clip()` + `blitSwapped()`); NV3001B override writing into #747's back buffer at native panel resolution |
| **757** | Dark palette in Offband brand colours; both blues gone — title banner `0x001F` and `corp_blue` `0x001A` |
| **758** | Antialiased JetBrains Mono at 13px replaces the 1x2-stretched `font5x7` |
| **761** | Heap readout is opt-in via `-D OFFBAND_SHOW_HEAP`, off by default |

## The two design decisions worth reviewing

**The native-resolution path is a row copy into `frame_buf`, not a second SPI path.** #747's buffer
is already indexed in physical pixels, so it *is* the native-resolution surface. A direct-SPI blit
would have been actively wrong: it writes to the panel while every other draw buffers, so the next
`endFrame()` would erase the splash. The unbuffered fallback (#747 supports a null buffer rather
than losing the display) shares `clip()` with the buffered path so the two cannot disagree on
geometry.

**Antialiasing is a dividend of #748, not an independent feature.** Blending a glyph edge means
reading the pixel under it. Over SPI that is a read-back per pixel; against `frame_buf` it is two
instructions. This could not have been built the day before #748 merged.

## Evidence

- **Native suite 202/202.** 40 new tests: blit/clip geometry, `DisplayDriver` contract (they stop
  compiling if `drawRGB565` is ever made pure virtual), palette invariants, two generators.
- **Guards proven to fail, not merely to pass.** Drifting `SPLASH_LOGICAL_H` 64→80 breaks the
  generator/driver agreement test; leaving `primary_txt` black fails contrast; leaving `title_bkg`
  blue fails the no-blue check. A guard nobody has watched fail is not a guard.
- **Containment measured:** the splash asset symbol links into the colour env only (`0x41a0` in the
  map) and is **absent** from the mono and no-display builds. Regenerating the 1-bit
  `offband_logo.h` yields a **byte-identical** file.
- **Cost:** static RAM **±0**, heap **±0** against `firmware-base`. Flash +17,172 B for #749
  (16,800 asset + 372 code), +3,764 B net for #758 (3,990 table − retired `font5x7`).
- **Builds:** `heltec_rc32_companion_radio_usb`, `..._diag_nocrashlog`,
  `heltec_rc32_without_display_companion_radio_usb`, `heltec_v4_companion_radio_ble` — all SUCCESS.
- **Hardware: owner-verified on `rc32-bench-1`** across four flashes — splash, palette, 10px and
  13px text. Owner selected 13px.

## Gemini review — two rounds

Round 1: INT_MIN UB in `clip()` **fixed**; a 440 B stack row in the fallback **cut to a fixed 64 B
chunk** (its VLA suggestion rejected — a VLA is still stack, sized at runtime, and is a GCC
extension in C++). A third finding, a "critical" typo `offband_splash_rgb55`, was **refuted**: no
such text exists, the call site reads `offband_splash_rgb565`, and that env builds clean.

Round 2 (final state): clean except the generator mirroring panel geometry the C++ owns with only a
comment holding them in sync. Now **derived** where possible and **parsed-and-asserted** where not.

## Things a reviewer should push on

- **Size-2 text is not hardware-verified.** At 13px a size-2 cell is 36px tall against 21px
  originally. Four call sites use `setTextSize(2)`. Size 1 is a perfect drop-in (advance 6 → 8 is
  size-1's only change, and the splash layout absorbs it); size 2 is not.
- **The unbuffered `drawRGB565` fallback has never executed on hardware.** The RC32 always had the
  buffer. It is unit-tested and shares its clipping with the buffered path, but #624/#625 (RCC6/C6
  and RC52/nRF52, where a 56 KB buffer against ~91 KB free is doubtful) would be its first real run.
  Recorded on both epics.
- **Splash line positions are a per-LAYOUT exception, not per-display.** Branching on dimension
  would not work: `display.width()/height()` report 128x64 on the RC32 too. `DisplayDriver` exposes
  no physical size by deliberate choice in #749 — this is the second request for it.


Strycher:
Human merge approval received from the owner in chat. `ci-green` passed (run 31998277409), PR is MERGEABLE / CLEAN. Enabling auto-merge with rebase per CLAUDE-BASE — the human decides, automation merges.

PR #768, opened 2026-08-17
fix: three 'get' keys returned an empty reply; four CLI callers transmitted uninitialised stack

### LINKED ISSUE #764: bug: three 'get' keys return an empty reply — garbage over the client CLI (radio, agc.reset.interval, pwrmgt.bootmv)
**What:** three `get` commands return nothing at all — and over the client CLI they return whatever junk was on the stack.
**Where:** `src/helpers/CommonCLI.cpp`, `handleGetCmd()`.
**What form:** `get radio`, `get agc.reset.interval`, `get pwrmgt.bootmv`.

Reported from the field: a user on MC 1.15.0 stock who upgraded to Offband ran `get radio` from the client app and got garbage text back. This is **not** a prefs-migration failure — 1.15.0 Path A records are 291 bytes, which has always been in the layout table, and a migration refusal yields valid default text, not garbage.

## Root cause

`d4d417ee fix(#299): exact-token match for CLI get keys` converted the `get` chain from prefix `memcmp` to exact-token `isKey`. For most keys it hoisted the new `isKey` guard **and** moved the body. For three keys it hoisted the guard and **left the body behind** on the old `memcmp` arm, which is now unreachable:

```c
} else if (isKey(config, "radio")) {          // line 1614 - EMPTY BODY, matches "radio"
} else if (memcmp(config, "radio.rxgain", 12) == 0) {
  ...
} else if (memcmp(config, "radio", 5) == 0) {  // line 1623 - the real body, now unreachable
  sprintf(reply, "> %s,%s,%d,%d", freq, bw, sf, cr);
```

Same shape at `agc.reset.interval` (empty 1582, body 1585) and `pwrmgt.bootmv` (empty 1742, body 1747).

All five lines blame to the same commit. The three arms have been dead since **2026-07-18**.

## Why the two transports differ

The symptom depends entirely on whether the caller initialised `reply`:

| Path | Buffer | Symptom |
|---|---|---|
| Serial console — `examples/simple_repeater/main.cpp:1157` | `char reply[160]; reply[0] = 0;` | **silent** — the `if (reply[0])` guard suppresses output |
| Client / over-the-air — `examples/simple_repeater/MyMesh.cpp:736` | `uint8_t temp[166];` never zeroed on the non-retry path | **uninitialised stack sent to the peer** |

That is exactly the reported garbage. The buffer-initialisation half is filed separately — it is a general defect, not specific to these three keys.

## Blast radius

Three keys, confirmed by a mechanical scan of every dispatch chain in `src/helpers/` and `examples/` (not by eyeballing):

- `get radio` — the core frequency/bandwidth/SF/CR readout. Affects **every role and every board**.
- `get agc.reset.interval`
- `get pwrmgt.bootmv` — nRF52 boot voltage.

The same scan cleared the rest: `WifiConfigProvider.cpp`, `RemoteCommand.cpp`, `companion_radio/MyMesh.cpp`, `simple_repeater`, `simple_room_server`, `simple_secure_chat`, `simple_sensor/SensorMesh.cpp`. Three other hits were false positives (a deliberate documented empty body at `CommonCLI.cpp:1041`; `home`/`default` arms disambiguated by arity; `ExtraFS/` and `reboot` appearing in genuinely separate chains).

Residual dead code from the same conversion, harmless but hiding the live defects: duplicate `radio.rxgain` (1611/1615) and `pwrmgt.bootreason` (1734/1743) arms with identical bodies.

`set` is unaffected — the conversion only touched `handleGetCmd`.

## Fix

1. Restore the three bodies to their `isKey` arms; delete the unreachable `memcmp` duplicates.
2. Fold the duplicate `radio.rxgain` arms into one. Upstream MeshCore has no `#if defined(USE_SX1262)…` guard on this key; the guarded copy added by #299 plus the surviving unguarded original means both variants behave identically today, so consolidating to the upstream shape is behaviour-preserving.
3. Add `scripts/check_cli_dispatch.py` as a CI guard so this defect class cannot land again: it fails on a key-matching `else if` with an empty body, and on a key matched twice within the same contiguous chain where the earlier arm subsumes the later one. Chain-aware and arity-aware, so the false positives above stay green.

## Verification

The guard must fail on the current tree and pass after the fix, and `get radio` must be confirmed on hardware over **both** transports — serial (currently silent) and the client (currently garbage).

---
Agent: **FoggyCanyon** (Agent Mail, `app-c-dev-crosswire`)

### LINKED ISSUE #765: bug: over-the-air CLI sends uninitialised stack when a command writes no reply
**What:** the over-the-air CLI handler sends uninitialised stack memory to the requesting peer whenever a command writes no reply.
**Where:** `examples/simple_repeater/MyMesh.cpp` ~736, in the remote-CLI (`TXT_TYPE_CLI_DATA`) path.
**What form:** the peer receives a text frame containing whatever bytes happened to be on the stack.

Found while root-causing #764. That issue is three specific `get` keys with empty bodies; **this** is the amplifier that turned a silent no-op into garbage on the wire, and it is independent of those three keys.

```c
uint8_t temp[166];                 // never initialised
char *command = (char *)&data[5];
char *reply = (char *)&temp[5];    // points into uninitialised storage
if (is_retry) {
  *reply = 0;                      // zeroed ONLY on the retry path
} else {
  handleCommand(sender_timestamp, command, reply);
}
int text_len = strlen(reply);      // strlen over uninitialised memory
if (text_len > 0) { ...transmit... }
```

On the non-retry path `reply` is never zeroed before the call. Any command that returns without writing produces a `strlen()` over uninitialised stack, and the result is transmitted.

Compare the serial caller, `examples/simple_repeater/main.cpp:1157`, which gets this right:

```c
char reply[160];
reply[0] = 0;
the_mesh.handleCommand(0, command, reply);
if (reply[0]) { Serial.print("  -> "); Serial.println(reply); }
```

## Why this is worth its own issue

Fixing #764 removes the three commands that currently trigger it. It does **not** remove the defect: any future handler that returns early, hits a compiled-out `#if` branch, or forgets to write on an error path resurrects it — silently, and only on the radio transport, which is the one hardest to observe.

The bytes disclosed are prior stack frames of the packet-handling path. The remote CLI is admin-authenticated, so the recipient is not arbitrary, but the firmware should not be handing out stack contents.

## Fix

Zero the reply buffer unconditionally before dispatch, so "handler wrote nothing" always yields "no reply sent" on every transport, exactly as it already does on serial:

```c
*reply = 0;
if (!is_retry) {
  handleCommand(sender_timestamp, command, reply);
}
```

Audit the other `handleCommand` call sites for the same pattern while here.

---
Agent: **FoggyCanyon** (Agent Mail, `app-c-dev-crosswire`)

Strycher:
Closes #764
Closes #765

**What:** three `get` commands have returned an empty reply since 2026-07-18, and over the client CLI that empty reply came back as stack garbage.
**Where:** `src/helpers/CommonCLI.cpp` `handleGetCmd()`, plus four `handleCommand` call sites.
**Reported by:** a user upgrading from MC 1.15.0 stock who ran `get radio` in the app.

This is **not** a prefs-migration failure. A 1.15.0 Path A record is 291 bytes, which has always been in the layout table, and a migration refusal produces valid default text rather than garbage.

## Root cause

`d4d417ee fix(#299): exact-token match for CLI get keys` converted the get chain from prefix `memcmp` to exact-token `isKey`. For most keys it hoisted the new guard *and* moved the body. For three keys it hoisted the guard and left the body on the old arm, which the new one now shadows:

```c
} else if (isKey(config, "radio")) {          // matches, writes nothing
} else if (memcmp(config, "radio", 5) == 0) { // the real body, unreachable
```

It compiles. No compiler considers it dead code. And the serial console hides it, because that caller zeroes its buffer and suppresses empty replies — so on serial the symptom is silence, not garbage. Only the radio transport shows it.

## Why the transports disagreed — the second defect

| Path | Buffer | Symptom |
|---|---|---|
| Serial (`simple_repeater/main.cpp:1157`) | `char reply[160]; reply[0] = 0;` | silent |
| Over-the-air (`simple_repeater/MyMesh.cpp:736`) | `uint8_t temp[166]`, zeroed only on the retry branch | **uninitialised stack transmitted** |

An audit of every `handleCommand` call site found **four** defective, not one:

- `simple_repeater/MyMesh.cpp` — over-the-air
- `simple_room_server/MyMesh.cpp` — over-the-air, admin branch only; the other branches happen to zero it
- `simple_sensor/SensorMesh.cpp` — over-the-air, never zeroed on any path
- `simple_sensor/main.cpp` — **serial**, missing `reply[0] = 0`

`simple_sensor` was the only role missing it on both transports, so it had no safe transport to fall back on. Fixing the three keys removes today's triggers but not the mechanism — any handler that returns without writing would resurrect it — which is why the buffers are fixed here too.

## Changes

1. Three bodies restored onto their `isKey` arms; shadowed `memcmp` duplicates removed.
2. `cad` and `extra.sf` converted to `isKey` — two arms the original conversion missed that still prefix-matched junk suffixes.
3. `#if defined(USE_SX1262)…` dropped from the `radio.rxgain` getter. `rx_boosted_gain` is an unconditional field and its `set` counterpart is unguarded, so the getter should be too. Upstream MeshCore agrees.
4. Reply buffer zeroed at the four call sites above.
5. `scripts/check_cli_dispatch.py` + 17 unit tests, wired **blocking** into `config-lint`.

## The guard

Two checks, both scoped to contiguous if/else-if chains: a key-matching arm with an empty body, and a key claimed twice in one chain where the earlier arm tests nothing but that key. Chain- and arity-aware, so `home`/`default` split by argument count, and `ExtraFS/` appearing in three separate handlers, stay green.

It reports **8 findings on the unfixed tree and none after**. The 17 tests pin both the regression shape and every false positive an earlier draft produced — a guard that cries wolf gets switched off.

## Verification

- **Builds 5/5 SUCCESS:** `Heltec_WSL3_sensor`, `heltec_v4_repeater_telemetry`, `heltec_v4_room_server`, `heltec_v4_companion_observer_wifi`, `RAK_4631_repeater` (the nRF52 leg that exercises the `pwrmgt.bootmv` `#ifdef` restructure).
- Guard clean; 17/17 unit tests pass.
- Gemini review completed. One point adopted — a sharper rationale for the `radio.rxgain` guard removal. Two findings verified wrong: it reported `get pwrmgt.bootreason` as deleted, but that handler is intact at `CommonCLI.cpp:1728` (it sat above the diff hunk's context), and its predicted preprocessor false-positive does not occur, since directive lines make a body non-empty.

**Not hardware-verified** — owner elected to skip the flash to get this to the reporting user. The `get radio` output path itself is unchanged code, simply reachable again.

## Scope checked and cleared

`set` chain, companion `0xC5` dispatch, observer CLI, `simple_secure_chat`, `WifiConfigProvider`, `RemoteCommand`. A separate scan for the related failure mode — an arm whose only `reply` write sits inside an unguarded `#ifdef` — found no instances; every conditional arm carries an `#else`.

---
Agent: **FoggyCanyon** (Agent Mail, `app-c-dev-crosswire`)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,21 @@ jobs:
       - name: Font generator self-test
         run: python scripts/test_gen_jbm_font.py
 
+      # #764. `get radio` returned an empty reply for ten months: #299's conversion
+      # hoisted three `isKey` guards without their bodies, leaving the real handlers
+      # on unreachable arms below. It compiles, it is not dead code to any compiler,
+      # and the serial console hides it -- that caller zeroes its buffer and skips
+      # empty replies. Only the radio transport showed it, as stack garbage (#765).
+      #
+      # BLOCKING, unlike the advisory check below: this guard carries its own unit
+      # tests (including every false positive an earlier version produced), and the
+      # tree is clean under it, so it cannot wrongly fail somebody else's PR.
+      - name: CLI dispatch guard self-test
+        run: python scripts/test_cli_dispatch.py
+
+      - name: No empty or shadowed arms in the CLI dispatch chains
+        run: python scripts/check_cli_dispatch.py
+
       # ADVISORY until #648 (the capability inventory) validates what this asserts.
       # ci-green is a hard merge gate for every PR in the repo (#477), so a check
       # built on unvalidated assumptions must not block other people's work. The
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -736,9 +736,11 @@ void MyMesh::onPeerDataRecv(mesh::Packet *packet, uint8_t type, int sender_idx, 
       uint8_t temp[166];
       char *command = (char *)&data[5];
       char *reply = (char *)&temp[5];
-      if (is_retry) {
-        *reply = 0;
-      } else {
+      // #765: zero BEFORE dispatch, not only on the retry path. `temp` is
+      // uninitialised stack, so a handler that writes nothing would otherwise
+      // leave strlen() below reading -- and transmitting -- whatever was there.
+      *reply = 0;
+      if (!is_retry) {
         handleCommand(sender_timestamp, command, reply);
       }
       int text_len = strlen(reply);
--- a/examples/simple_room_server/MyMesh.cpp
+++ b/examples/simple_room_server/MyMesh.cpp
@@ -463,12 +463,14 @@ void MyMesh::onPeerDataRecv(mesh::Packet *packet, uint8_t type, int sender_idx, 
                           PUB_KEY_SIZE);
 
       uint8_t temp[166];
+      // #765: `temp` is uninitialised stack. Every branch below except the admin
+      // dispatch already zeroed it, so the one path that calls handleCommand was
+      // the one path that could transmit stack bytes. Zero it up front instead.
+      temp[5] = 0; // no reply
       bool send_ack;
       if (flags == TXT_TYPE_CLI_DATA) {
         if (client->isAdmin()) {
-          if (is_retry) {
-            temp[5] = 0; // no reply
-          } else {
+          if (!is_retry) {
             handleCommand(sender_timestamp, (char *)&data[5], (char *)&temp[5]);
             temp[4] = (TXT_TYPE_CLI_DATA << 2); // attempt and flags,  (NOTE: legacy was: TXT_TYPE_PLAIN)
           }
--- a/examples/simple_sensor/SensorMesh.cpp
+++ b/examples/simple_sensor/SensorMesh.cpp
@@ -586,6 +586,7 @@ void SensorMesh::onPeerDataRecv(mesh::Packet* packet, uint8_t type, int sender_i
         uint8_t temp[166];
         char *command = (char *) &data[5];
         char *reply = (char *) &temp[5];
+        *reply = 0;   // #765: uninitialised stack otherwise; strlen() below would transmit it
         handleCommand(sender_timestamp, command, reply);
 
         int text_len = strlen(reply);
--- a/examples/simple_sensor/main.cpp
+++ b/examples/simple_sensor/main.cpp
@@ -162,6 +162,7 @@ void loop() {
   if (len > 0 && command[len - 1] == '\r') {  // received complete line
     command[len - 1] = 0;  // replace newline with C string null terminator
     char reply[160];
+    reply[0] = 0;   // #765: the `if (reply[0])` guard below reads this either way
     the_mesh.handleCommand(0, command, reply);  // NOTE: there is no sender_timestamp via serial!
     if (reply[0]) {
       Serial.print("  -> "); Serial.println(reply);
--- a/scripts/check_cli_dispatch.py
+++ b/scripts/check_cli_dispatch.py
@@ -1,0 +1,332 @@
+#!/usr/bin/env python3
+"""Static guard for the CLI dispatch chains (#764).
+
+`d4d417ee` converted `handleGetCmd` from prefix `memcmp` to exact-token `isKey`.
+For most keys it hoisted the new guard *and* moved the body. For three keys it
+hoisted the guard and left the body behind on the old arm:
+
+    } else if (isKey(config, "radio")) {          // matches, writes nothing
+    } else if (memcmp(config, "radio", 5) == 0) { // the body, now shadowed
+      sprintf(reply, "> %s,%s,%d,%d", ...);
+
+`get radio` then returned an empty reply for ten months. Nothing caught it: it
+compiles, it is not dead code by any compiler's reckoning, and the symptom is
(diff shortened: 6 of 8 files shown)


Strycher:
Closed at the owner's direction — not because the code is wrong, because the sequence is.

The guard that validates this fix was bundled into the same PR as the fix. CI green here proves only that the guard does not false-positive on already-corrected code; it never ran against the broken tree, so there is no evidence visible to a reviewer that the guard catches anything. The claim that it reports 8 findings pre-fix exists only in an agent's terminal.

Correct order, to be agreed before anything goes up again: the guard lands first and demonstrably goes red against the unfixed tree, then the fix lands and the log goes quiet, then advisory-vs-required is decided deliberately rather than by me.

The work is preserved on `fix/764-cli-empty-get-replies`. #764 and #765 remain open.

---
Agent: **FoggyCanyon** (Agent Mail, `app-c-dev-crosswire`)

Issue #773, opened 2026-08-17
epic: ship an RC32 beta binary to RadioCore testers


Strycher:
Working checklist for getting a versioned, reproducible RC32 image into a beta tester's
hands. Ordered — each item unblocks the next. Written 2026-08-17 after the overnight #702 /
#763 / #766 session.

## Path to a tester binary

- [ ] **1. Merge #772** — adds the four rc32 envs to `release-envs.txt`. Open, builds
      verified, awaiting review. *Without this the release pipeline emits no RC32 asset.*

- [ ] **2. Land `fix/704-rc32-no-gps`** — 4 commits, unmerged, no PR yet. Needs a PR opened
      and merged. **This is the real blocker on tester safety**, not #772:
      - `4c0d212b` disable GPS — its enable pin is the **VDD_SPI strap**
      - `bfcafb97` strip peripherals the RC32 does not have
      - `25c1ef80` align to Heltec's own board definition
      - `3467ef54` `DISPLAY_ROTATION=0` — landscape, owner-confirmed
      Both display/GPS changes are already hardware-confirmed; they just never landed.

- [ ] **3. Add rc32 envs to `ci.yml`** — no rc32 env is in the CI matrix, so `ci-green` does
      not cover them and nothing catches a break on push. Mirror the four release envs.

- [ ] **4. Hardware-verify the STOCK env on `rc32-bench-1`** — one flash, then confirm boot,
      display, BLE pairing, radio. **No RC32 companion env has ever been booted in stock
      form.** All bench work to date ran on `..._ble_diag`, which locally sets
      `ENV_INCLUDE_GPS=0` and `DISPLAY_ROTATION=0` and therefore masks exactly what #704
      fixes. This step is the one that would otherwise burn the tester.

- [x] ~~**5. Confirm the tester's sub-variant** -- RC32-L62 vs RC32-68.~~
      **REMOVED FROM THE CHAIN (owner, 2026-08-19).** This was never a delivery
      dependency and should not have been written as one. We publish artifacts; testers
      decide what to flash. Confirming one tester's hardware blocks nothing, produces no
      artifact, and does not belong in a chain whose items "each unblock the next".
      The surviving fact needs no task: **our RC32 assets are L62-only.** RC32-68 (HaLow,
      HT-HC01_V2) is a different pinout and would be a new variant -- new work, its own
      feature/epic, if ever wanted. Asking a tester what they own is a support message,
      not a tracked item.

[...]

Issue #766, opened 2026-08-17
bug: RC32 getBattMilliVolts() returns 0 with a LiPo on VBAT


Strycher:
## What

`HeltecRC32Board::getBattMilliVolts()` returns **0 on every sample** on `rc32-bench-1` with a charged LiPo wired to VBAT.

Observed on the wire via the #763 UART0 power telemetry (`OFFBAND_POWER_TELEMETRY`, 30 s cadence):

```
[2026-08-17T06:27:42.389Z] [122375] [pwr] mv=0 up=122s
[2026-08-17T06:28:12.389Z] [152375] [pwr] mv=0 up=152s
```

## Failing layer (isolated, not guessed)

`variants/heltec_rc32/HeltecRC32Board.cpp:58` returns `adc_mult * raw`. `adc_mult` is pinned at
`ADC_MULTIPLIER` (4.9) and `setAdcMultiplier(0)` coerces back to it, so it can never be zero.

Therefore a zero result requires **`raw == 0`** — i.e. `analogReadMilliVolts(PIN_VBAT_READ)` is
returning nothing. The defect is in the ADC read, not the scaling.

## Refuted

**Pin contention on GPIO7.** The RC32 LoRa map (`variants/heltec_rc32/variant.h`) is
SCK 11 / MISO 13 / MOSI 12 / CS 10 / DIO1 14 / RESET 9 / BUSY 1. Nothing else in the variant claims
GPIO7 or GPIO15. `PIN_VBAT_READ=7` is ADC1_CH6 on the S3, consistent with `ADC_CHANNEL_6`.

## Live candidates

1. **Pin setup never runs.** `ESP32Board.h:38` does `pinMode` + `adcAttachPin` for `PIN_VBAT_READ`;
   `HeltecRC32Board` may not reach that path.
2. **`ADC_CTRL_ENABLED=HIGH` polarity inverted.** Several Heltec boards enable the divider by
   driving the control pin **low**.
3. **Divider not populated** on the RadioCore carrier (hardware, not firmware).

A fourth possibility the probe also covers: `analogReadMilliVolts()` returning 0 from an eFuse
calibration failure while `analogRead()` counts are healthy — a different defect that reporting
only mV would hide.

## Instrumentation (flashed)

`OFFBAND_POWER_TELEMETRY_PROBE` reports raw counts and calibrated mV at **both** ADC_CTRL
polarities, separating all four in one flash:

```
[...]

Issue #625, opened 2026-08-11
feature: RadioCore RC52 (nRF52840) firmware support


Strycher:
Bring the **RadioCore RC52** (RA62A-HF + nRF52840, SX1262 + FEM) up on Offband. Parent: #622.

## Owner role target

**BLE companion, serial companion, repeater with screen or headless. No WiFi — BLE only.** (nRF52840 has no WiFi; this matches the silicon.)

## Prior art

No upstream MeshCore variant. n30nex ships MIT, 1.17.0-based RC52-L62 builds: BLE companion, headless low-power repeater, room server headless, and room server with NV3001B TFT. Their notes: UF2 bootloader for normal installs; do not modify the SoftDevice or bootloader.

## Pin map (from the owner's RC52 pinout diagram, `docs/radiocore/RC52_nRF52/`)

- LoRa: `NSS=P0.13  MISO=P0.14  MOSI=P0.22  BUSY=P0.24  SCK=P0.25  NRST=P1.00  DIO1=P0.11`
- FEM: `FEM_EN=P0.26  VFEM_CTRL=P0.16  FEM_LNA_CTRL=P1.07`
- Battery: `ADC_Ctrl=P0.04` gating `ADC_IN=P0.31`
- Other: `USER=P1.10  nRF_RST=P0.18  SWDIO=P0.30  NFC1=P0.10  nRF_RX=P0.07  nRF_TX=P0.08`

## HAZARD — enable-gated battery divider, the exact #602 shape

`ADC_Ctrl` (P0.04) gates `ADC_IN` (P0.31). This is the same enable-gated-divider pattern that **bricked the Wio Tracker L1 Pro on 2026-08-10**: `SAFEBOOT_VBAT_ENABLE_ACTIVE` was copied from `xiao_nrf52` without checking which way that board drives the pin, so SafeBoot disabled the divider, sampled it dead, read below `SLEEP_MV` and deep-slept before USB init — at full charge, looking completely dead. See #602 / #620.

**Rule for this Epic: any SafeBoot battery config for RC52 takes its polarity from measuring THIS board. Never copy it from a sibling variant.** Confirm with a real reading before flashing anything with SafeBoot enabled.

## OPPORTUNITY — second board class with a real LNA bypass

[...]

Issue #624, opened 2026-08-11
feature: RadioCore RCC6 (ESP32-C6) firmware support


Strycher:
Bring the **RadioCore RCC6** (RA62A-HF + ESP32-C6, SX1262) up on Offband. Parent: #622.

## Owner role target

**WiFi; WebUI is the natural path.** ESP32-C6 = WiFi 6 + BLE. This is the unit where a browser-based config/dashboard makes sense.

## Prior art

No upstream MeshCore variant. n30nex ships an MIT, 1.17.0-based `variants/heltec_rcc6` with a full env set — `repeater, room_server, companion_radio_ble, companion_radio_usb, companion_radio_wifi, companion_radio_web_ap, kiss_modem`, each in with-display and without-display form — plus a separate MQTT observer/repeater firmware with an authenticated LAN WebUI.

Offband already supports ESP32-C6 (`esp32c6_base`; `xiao_c6`, `meshsmith_photon_esp32c6`, `lilygo_tlora_c6`), which de-risks the MCU side. Offband has **never** run its Observer stack on a C6 — `OFFBAND_OBSERVER` appears only in `heltec_v3`, `heltec_v4`, `xiao_s3_wio`.

## Verified pin facts

n30nex's TFT pins match the owner's RCC6 TFT Guide **exactly** — two independent sources:

`SCL=4  SDA=15  CS=18  DC=3  RST=0  EN=2 (active LOW)  BL=1 (active HIGH)`, software SPI.

LoRa: `SCLK=21 MISO=20 MOSI=22 NSS=23 DIO1=19 BUSY=10 RESET=8`. Button 9. Battery `ADC_CTRL=5, VBAT_READ=6, multiplier 4.95`.

## HEAP IS THE HEADLINE RISK — measure it first

A community RCC6 repeater dashboard reports **free heap 91 KB, minimum 17 KB** on a working C6 build. Offband's `OFFBAND_TLS_HEAP_FLOOR_BYTES` is **80 KB**, guarding a ~72 KB TLS handshake transient (`src/helpers/wifi_observer/WifiObserverConfig.h:85`).

An Offband TLS observer on this board would sit a few KB above its own floor — the same shape as the HV3 heap-floor deadlock (#521, fixed via #534 lazy client alloc). `[hypothesis: untested — the 91 KB figure is a community build, not ours]`

[...]

PR #787, opened 2026-08-17
feat(#786): RC32 bench instrumentation — fuel gauge, INA probe, command queue, runtime analysis

### LINKED ISSUE #786: task: land the RC32 bench instrumentation (fuel gauge, INA probe, command queue, runtime analysis)
Tooling built and hardware-verified during the 2026-08-17 RC32 session, currently existing
only in a working tree. Landing it so it is not lost and so RCC6 (#624) and RC52 (#625)
inherit a working bench instead of rebuilding one.

## What lands

**`tools/diag/rc32-boot-740/rc32_uart_sniffer_v3/rc32_uart_sniffer_v3.ino`** (modified)

- Board-agnostic pin map -- `SNIFFER_GENERIC_S3` switches between the Feather and a generic
  ESP32-S3 WROOM. On a WROOM the symbolic `RX` resolves to GPIO44, which IS U0RXD and is wired
  to the onboard USB-UART bridge; sniffing there puts two drivers on the console UART.
- **MAX17048 fuel gauge on `Wire1`**, reporting cell voltage and modelled charge every 30 s
  (same cadence as the RC32's `[pwr]` line, so the series align without interpolation).
  `Wire1` is both house convention (Wire = onboard, Wire1 = plugged-in) and forced: the
  MAX1704x has no address-select pin and the Feather already carries one onboard at 0x36.
- **INA current-monitor identification probe** across 0x40..0x4F, checking ID registers at
  *both* `0xFE/0xFF` (INA226/260/3221) and `0x3E/0x3F` (INA228/237/238). Address alone is not
  identity when addresses are strap-configurable.
- Bring-up follows `EnvironmentSensorManager.cpp:637` exactly: on ESP32 the pins are arguments
  to `begin()`; on nRF52 they must be set before it. Wrong order does not warn, it silently
  fails to come up.
- Targeted probes only, never a bus scan (#294: a blind scan wedges the C6 I2C peripheral at
  0x0d and never returns -- relevant with RCC6 next).

**`tools/diag/rc32-boot-740/scripts/capture.py`** (modified)

- **Command queue.** `--send` opens the port itself and therefore cannot run while a capture
  holds it (Windows serial ports are exclusive). The docstring previously claimed otherwise,
  which made the command surface useless exactly when it mattered -- a reset is only worth
  issuing if something is recording what it triggers.
- `--queue RST` now writes to `<out>.cmd`; the running capture picks it up within ~1 s and
  forwards it over the port it already owns. Logged as `<<< INJECT <cmd>` immediately above
  whatever it causes.
- Verified on hardware: injection and `>>> PONG SNIFFER-v3` landed in the same log 1 ms apart.
- Corrected the docstring rather than leaving it asserting the wrong thing.

**`tools/diag/rc32-boot-740/scripts/battery_runtime.py`** (new)

- Projects runtime from a `[pwr]` capture using two methods that fail differently -- slope fit
  and charge accounting -- and reports a RANGE when they disagree rather than a false number.
- Discards a leading warm-up window: a freshly charged cell shed 19 mV in 90 s, which
  linearly extrapolated predicts death within the hour.
- Anchors the charge curve to the observed full-charge voltage. The owner's pack was full at
  4130 mV; an unanchored table called that 85% and opened the accounting already claiming
  ~375 mAh consumed.
- Selects the trailing contiguous run of valid samples, so pre-run zeros (from the reversed
  battery leads in #766) cannot anchor the fit -- with them included it reported the battery
  *charging* at +6482 mV/h.

**`docs/rc32-beta-checklist.md`** (new)

- Mirror of #773. GitHub was mid-incident and #773 was unreachable exactly when it was needed;
  the issue remains source of truth and the file says so.

## Verification

All hardware-verified on `rc32-bench-1` + `feather-sniffer` during the session: gauge readings
cross-check against the RC32's own ADC within ~20-28 mV, command injection round-trips, and
the runtime analysis correctly refuses to project from insufficient data.

Strycher:
Closes #786.

Tooling built and hardware-verified during the 2026-08-17 RC32 session. **Diagnostics only —
no firmware source is touched**, so there is no risk to any shipping env. Landing it so RCC6
(#624) and RC52 (#625) inherit a working bench rather than rebuilding one.

## Sniffer sketch

- **Board-agnostic pin map** (`SNIFFER_GENERIC_S3`). On a generic ESP32-S3 WROOM the symbolic
  `RX` resolves to **GPIO44 — U0RXD**, wired to the onboard USB-UART bridge. Sniffing there
  puts two drivers on the console UART and presents as a wiring fault rather than a config
  error.
- **MAX17048 fuel gauge on `Wire1`** — voltage + modelled charge every 30 s, matching the
  RC32's `[pwr]` cadence so the two series align without interpolation. `Wire1` is house
  convention *and* forced: the MAX1704x has no address-select pin and the Feather already
  carries one onboard at 0x36.
- **INA identification probe** over 0x40–0x4F, reading ID registers at **both** `0xFE/0xFF`
  (INA226/260/3221) **and** `0x3E/0x3F` (INA228/237/238). Address is not identity when
  addresses are strap-selectable — probing one location only would find an INA228 at 0x40 and
  misreport it as an INA219. Identification only; the current decode gets written once the
  part is known rather than shipping three untested decoders.
- Bring-up copies `EnvironmentSensorManager.cpp:637`: ESP32 passes pins **to** `begin()`,
  nRF52 sets them **before**. Wrong order fails silently.
- Targeted probes throughout, **never a bus scan** — #294 records a blind scan wedging the C6
  I2C peripheral at 0x0d and never returning. RCC6 is a C6.

## Capture command queue

`--send` opens the port itself, so it could never run while a capture held it — Windows serial
ports are exclusive. The docstring claimed the opposite, making the command surface useless
exactly when it mattered: issuing a reset is only worth doing if something is recording what it
triggers.

`--queue RST` now writes `<out>.cmd`; the running capture forwards it over the port it already
owns and logs `<<< INJECT <cmd>` directly above the result.

**Verified on hardware:**

```
[2026-08-17T20:39:00.654Z] <<< INJECT PING
[2026-08-17T20:39:00.655Z] >>> PONG SNIFFER-v3  @up=5428s
```

## Runtime analysis (`battery_runtime.py`)

Two methods that fail differently — slope fit and charge accounting — reporting a **range**
when they disagree rather than a confident wrong number. Three guards, each from an actual
error made during the session:

- **Warm-up discard.** A freshly charged cell shed 19 mV in 90 s; extrapolated linearly that
  predicts death within the hour.
- **Curve anchoring.** The owner's pack was full at 4130 mV; the unanchored table called that
  85% and opened the accounting already claiming ~375 mAh consumed.
- **Trailing-run selection.** Pre-run zeros from the reversed battery leads (#766) otherwise
  anchor the fit — with them included it reported the battery *charging* at +6482 mV/h. An
  estimator that can emit a positive discharge slope is reporting a parsing bug as physics.

## Checklist mirror

`docs/rc32-beta-checklist.md` mirrors #773. GitHub was mid-incident and #773 was unreachable
exactly when it was needed. The issue stays source of truth and the file says so in its header.

## Verification

Hardware-verified on `rc32-bench-1` + `feather-sniffer`: gauge readings cross-check against the
RC32's own ADC within **20–28 mV** (independently validating the 390K/100K + 4.9 sense chain
from #766), command injection round-trips, and the analysis correctly refuses to project from
insufficient data.

Live capture logs under `evidence/` are deliberately not committed.

### DIFF
--- a/docs/rc32-beta-checklist.md
+++ b/docs/rc32-beta-checklist.md
@@ -1,0 +1,114 @@
+# RC32 beta binary — working checklist
+
+> **Mirror of [#773](https://github.com/OffbandMesh/meshcore-firmware/issues/773).**
+> The issue is the source of truth (board discipline); this file exists so the checklist is
+> readable when GitHub is unavailable — it was unreachable during the 2026-08-17 incident,
+> which is why this copy exists. If the two disagree, the issue wins. Update both together.
+
+Snapshot: 2026-08-17, after the overnight #702 / #763 / #766 session.
+
+---
+
+## Path to a tester binary
+
+Ordered — each item unblocks the next.
+
+- [x] **1. Merge #772** — four rc32 envs added to `.github/release-envs.txt`.
+      **Merged `36b6936a`, 2026-08-17T15:14Z**, confirmed on `firmware-base`.
+      *Without this the release pipeline emitted no RC32 asset at all.*
+
+- [ ] **2. Land `fix/704-rc32-no-gps`** — 4 commits, unmerged, **no PR yet**.
+      **This is the real blocker on tester safety**, not #772:
+
+      4c0d212b  disable GPS on heltec_rc32 -- its enable pin is the VDD_SPI strap
+      bfcafb97  strip peripherals the RC32 does not have
+      25c1ef80  align heltec_rc32 to Heltec's own board definition
+      3467ef54  DISPLAY_ROTATION=0 -- landscape, owner-confirmed
+
+      Both the GPS and rotation changes are already hardware-confirmed. They simply never
+      landed. Until they do, the stock env ships with GPS **enabled** on a pin that is the
+      VDD_SPI strap.
+
+- [ ] **3. Add rc32 envs to `ci.yml`** — no rc32 env is in the CI matrix, so `ci-green` does
+      not cover them and nothing catches a break on push. Mirror the four release envs.
+
+- [ ] **4. Hardware-verify the STOCK env on `rc32-bench-1`** — one flash, then confirm boot,
+      display, BLE pairing, radio.
+      **No RC32 companion env has ever been booted in stock form.** Every bench result to
+      date came from `..._ble_diag`, which locally sets `ENV_INCLUDE_GPS=0` and
+      `DISPLAY_ROTATION=0` — masking exactly what #704 fixes. This is the step that would
+      otherwise burn the tester.
+
+- [ ] **5. Confirm the tester's sub-variant** — RC32-**L62** (HT-RA62A LoRa) vs RC32-**68**
+      (HT-HC01_V2 Wi-Fi HaLow). Different pinouts; only L62 is supported. A HaLow unit needs
+      a new variant — new work, not a build. Ask before promising anything.
+
+- [ ] **6. Cut `offband-v1.5.0-beta3`** — requires the release preview and an explicit human
+      "ship it" per `VERSIONING.md`. Must come **after** items 2 and 4.
+
+- [ ] **7. Send the tester the release link** — not a CI artifact, not a file from the bench.
+
+---
+
+## Carried debt from the overnight session
+
+- [ ] **#763 — level-gate the UART0 mirror.**
+      It is a synchronous bounded busy-wait at 115200 baud and now runs regardless of capture
+      state. With `BLE_DEBUG_LOGGING=1` it blocked inside `setup()`: boot 2 s → 118 s → never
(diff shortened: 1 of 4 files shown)


Strycher:
Human approval received from Ben in session f148bd05. Re-validated against the remote immediately prior: 1 commit, 4 files, base `firmware-base`, mergeable=MERGEABLE. Checks incomplete at approval time (20 SUCCESS / 27 pending, none failing) — auto-merge armed so it lands only on a green matrix.

Issue #294, opened 2026-07-14
Photon-1W ESP32-C6 companion_radio_usb: client can't connect over USB (handshake timeout)


Strycher:
## Summary

The MeshSmith Photon-1W **ESP32-C6** `companion_radio_usb` build does not service the companion protocol over USB — the client sees COM port `303A:1001` but the connection **times out / never handshakes**. Verified on hardware (`photon-c6`, 2026-07-14).

## Ruled out (all verified)

- **Flash**: full merged image (our bootloader `0x0` + partitions `0x8000` + otadata `0xe000` + app `0x10000`) written and **hash-verified** on-chip. Not a bad flash.
- **Partition layout**: device partition table read back = standard dual-OTA, `app0 @ 0x10000` — our app is in the right partition.
- **Bootloader**: replaced with ours via the full flash (originally app-slot kept MeshSmith's — also failed).
- **USB config**: `ARDUINO_USB_MODE=1` + `ARDUINO_USB_CDC_ON_BOOT=1` — correct for C6 (HWCDC / USB-Serial-JTAG, its only USB).
- **Companion binding**: `main.cpp:246 serial_interface.begin(Serial)` — correct; `ArduinoSerialInterface::isConnected()` hard-returns true.
- **Client**: the same client connects fine to other Offband devices (RAK / Heltec 1.1.x) over serial — not a client/version problem.

## Symptom
- COM `303A:1001` enumerates and is stable (no reboot loop).
- Client lists it, connection attempt never handshakes (timeout).
- Board LED solid amber (likely charge LED; ambiguous).
- Serial console silent on passive reads (companion build; may also be HWCDC drop).

## Open question / next step
Is the app booting past `setup()` / radio init on the C6, or is the companion RX/TX path over the C6 HWCDC not moving frames? Needs a **visibility build** (boot banner + `MESH_DEBUG` on a drained terminal) to see whether the device receives the client's `<`-framed APP_START and attempts a `>` reply. `setTxTimeoutMs(0)` (`main.cpp:159`) considered but weak — the working Heltec V4 companion is also HWCDC and calls it.

**Epic:** #193 (Photon-1W ESP32-C6). Hardware unit: `photon-c6`.

PR #803, opened 2026-08-18
docs(#802): RCC6 bring-up handoff from the RC32 session

### LINKED ISSUE #802: docs: RCC6 bring-up handoff from the RC32 session
Carries RCC6 findings out of the 2026-08-17 RC32 session so #624 does not re-derive them: the verified P1 header pinout, the explicitly UNVERIFIED LoRa pin map, the section 3.2 module-side-numbering trap, C6-specific hazards (blind I2C scan wedges the peripheral; the UART0 beacon is untested on RISC-V), bench state, and the schematic-extraction method that works.

Confidence markers are deliberate -- the RC32 session lost hours twice to claims asserted from partial reads, so nothing is stated more strongly than it was established.

Also posted to #624.

Strycher:
Closes #802. Docs only.

Carries RCC6 findings out of the 2026-08-17 RC32 session so #624 starts from evidence rather than re-deriving it.

**The confidence markers are the point.** `[verified:]` means an artifact produced the evidence and it is cited; `[hypothesis:]` means it is a lead and has not been confirmed. That distinction exists because the RC32 session lost hours twice to claims asserted from partial reads.

**Verified** — the P1 carrier header pinout, read from the rendered schematic symbol. U0TXD (12), CHIP_PU (18), USER_Key (5) and VBAT (2) match the RC32 positions, so the carrier footprint is family-consistent. But **pin 20 is VDD_3V3 here** where the RC32 sniffer wiring uses pin 20 as GND — carrying that habit across ties 3.3 V to the sniffer's ground rail.

**Not verified, flagged loudly** — the LoRa pin map. An early row-pairing appeared to match datasheet §3.2.1, which then turned out to list module-side pin names rather than carrier GPIOs, so the agreement proved nothing. Re-derive by rendering the MCU symbol.

**C6 hazards recorded** — a blind I2C scan wedges the peripheral at 0x0d and never returns (#294), and the UART0 boot beacon has never been tested on RISC-V despite being nominally portable.

Also includes the PyMuPDF extraction method (row-pair to generate leads, render to confirm) and current bench state, including that `feather-sniffer` is mid-discharge-run and must not be repointed.

### DIFF
--- a/docs/radiocore/rcc6-bringup-handoff.md
+++ b/docs/radiocore/rcc6-bringup-handoff.md
@@ -1,0 +1,145 @@
+# RCC6 (ESP32-C6) bring-up — handoff
+
+Findings carried out of the 2026-08-17 RC32 session so RCC6 work does not re-derive them.
+Companion to **#624** (epic: RadioCore RCC6 firmware support).
+
+> **Read the confidence markers.** `[verified:]` means an artifact produced the evidence and it
+> is cited. `[hypothesis:]` means it is a lead and has NOT been confirmed. The RC32 session lost
+> hours twice to claims that were asserted from partial reads, so nothing here is stated more
+> strongly than it was established.
+
+---
+
+## 1. What exists today
+
+- **No `variants/heltec_rcc6`**, no `boards/heltec-rcc6.json`. Nothing in tree.
+- **Upstream MeshCore has no RCC6 either** — we would be first. There is no reference variant to
+  copy from, unlike most boards in this repo.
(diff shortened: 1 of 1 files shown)


Strycher:
Human approval received from Ben in session f148bd05. Re-validated against the remote immediately prior: 1 commit, 1 file (`docs/radiocore/rcc6-bringup-handoff.md`), base `firmware-base`, mergeable=MERGEABLE, checks 45 SUCCESS / 2 pending, none failing. Auto-merge armed so it lands only on a green matrix.

Issue #808, opened 2026-08-18
task: pio-flash --known-port operator port-assertion for passively-unidentifiable bridge devices


Strycher:
## Problem

Some registered boards cannot be identified passively by `pio-flash`:

- **Bridge-class, serial-less, with a registry rival.** `hv3-bench` (Heltec V3, CP2102 `10C4:EA60`, MAC `3c:0f:02:eb:f9:0c`) has no unique `usb_serial` (factory-default `0001`) and no per-device instance hash — its `InstanceId` ends in the shared default serial and its only parent is the root hub. Another serial-less same-class entry exists (`wroom-sniffer`, also `10C4:EA60`), so `resolve_device()` refuses with the #468 "registry overlap" rivals error. `--approve-class-match` does **not** override that path.

Result: a device the operator can physically identify (it is the only board on a known COM port) is unflashable by construction, with no legitimate escape.

## Solution (implemented, pending land)

Add `--known-port COMx` to `preview` / `confirm` / `monitor`. The operator asserts which present port is the device; it bypasses **only** the passive-guess refusals (rivals / multi-candidate). Guardrails held:

- The named port must still be **present** and a **serial-less candidate of the entry's own bridge class** — a port carrying a real serial can never be claimed this way.
- `confirm` still runs `_verify_bridge_mac()` before any write — `--known-port` returns `bridge_provisional=True`, so the SoC MAC gate always fires. **The human names the port; the MAC decides.**
- `--known-port` also sets `operator_asserted`, which satisfies the #503 `--approve-class-match` gate in read-only consumers (`monitor`, `info`) that cannot MAC-verify. Operator naming the exact port is a *more specific* authorization than approving a class guess, so it is not a weakening. (Owner-approved this semantics 2026-08-17.)
- `send` is deliberately **not** covered — it hard-refuses provisional matches by policy (it writes config-mutating CLI).

## Validation

Used end-to-end to flash `hv3-bench` with the #739 observer build: MAC-verify fired and matched before the write. Soak monitor attached via `monitor --known-port` with the gate cleared.

## Scope / relations

- Sub-issue of #371 (identify devices exactly). Builds directly on #468 (flash-time MAC verify) and #503 (serial-first identity).
- **Unblocks #807** — that P1 bootloader-discovery fix touches the same file. No line overlap (this changes `resolve_device` + subparsers; #807 is in `_discover_bootloader_port()`), so #807 rebases cleanly once this lands.

## Definition of done

[...]

Issue #807, opened 2026-08-18
bug: pio-flash bootloader discovery cannot see devices whose USB port survives download-mode entry


Strycher:
`pio-flash confirm` cannot flash an ESP32 whose USB port survives download-mode entry. It refuses after 20s with:

```
REFUSE: no bootloader port appeared within 20s after the trigger (matched neither
the device serial 588C812F91F8 nor a new serial-less 303A port with PID != 1001).
Last candidates: none
```

## Root cause

`[verified: scripts/pio-flash.py:979,987]`

```python
979  before_coms = {p["com"] for p in before}
987  new = [p for p in now if p["com"] not in before_coms]
```

`_discover_bootloader_port()` filters enumeration down to ports **absent before the trigger**, then applies its serial and vendor/PID matching to that set only.

An ESP32-C6 (and any native USB-Serial-JTAG part) keeps the **same COM port** entering download mode — the peripheral is in the SoC, the PID stays `303A:1001`, and Windows does not re-enumerate a different port. So `new` is permanently empty, no candidate can ever be evaluated, and the wait times out regardless of how healthy the device is.

The function's own docstring states the correct principle:

> *"the chip serial is constant across USB-mode changes ... so a new port carrying the SAME serial as the running port IS the device -- regardless of what VID:PID the new mode presents."*

The serial-equality check is right. It is applied to the wrong set. A device that never leaves its port is unreachable by construction.

## Scope — not RCC6-specific

Affects any ESP32 with native USB where the port survives download entry. `HARDWARE.local.md` already records "expect a **manual RST press** after any esptool op" for `rc32-bench-1` (native-USB S3) — that friction is plausibly this same defect rather than a board quirk, though that has not been confirmed. `[hypothesis: untested]`

Bridge-attached boards (CP2102/CH340) are unaffected: the bridge keeps its own port and the SoC behind it never re-enumerates, which is the case the current logic was written for.

## Evidence that the device is healthy

At refusal time all three interfaces were present and `OK`, on the same instance as registration:

```
USB\VID_303A&PID_1001&MI_00\8&3B623357&0&0000   (COM45)   OK
USB\VID_303A&PID_1001&MI_02\8&3B623357&0&0002   (JTAG)    OK
USB\VID_303A&PID_1001\58:8C:81:2F:91:F8         (composite) OK
```

`bootstrap` and `backup` on the same board both connected successfully, because those let esptool perform its own reset-and-connect rather than waiting for a port transition.

[...]

Issue #503, opened 2026-07-31
bug: find_in_registry gates on VID:PID before serial — a registered board goes 'unregistered' when it changes USB identity


Strycher:
## The defect

`find_in_registry` (`scripts/pio-flash.py:310-324`) filters candidate entries by VID:PID **before** it ever looks at the USB serial:

```python
for name, entry in (registry.get(kind_key) or {}).items():
    entry_vid_pids = entry.get("vid_pid") or []
    if vid_pid not in entry_vid_pids:
        continue                      # <-- device is discarded here

    # TIER 1 -- USB serial. Device-unique and port-independent: this IS identity.
    serials = entry_usb_serials(entry)
```

The comment on the line immediately after the gate asserts the serial "IS identity." It cannot be — a device whose current VID:PID isn't already listed on its entry is discarded before the serial is read. The nRF52840 chip DEVICEID is **globally unique**; it needs no VID:PID pre-filter to be trustworthy. Gating it inverts the stated design: VID:PID is doing the identifying, and the serial is only a tiebreaker among devices that already passed a VID:PID check.

**Owner's framing (2026-07-31):** the entire point of keying on the chip serial is to be VID:PID-independent. A tool that ignores VID:PID only *after* matching VID:PID does not ignore VID:PID.

## Reproduction — live, this session

`sensecap-t1000e-1`, a stock Seeed SenseCAP **T1000-E** (nRF52840, ships with Meshtastic), chip DEVICEID `ACEDFAE4BC94D153`:

| Stage | VID:PID | COM | `pio-flash list` verdict |
|---|---|---|---|
| At registration | `239A:8029` | 17 | `device:sensecap-t1000e-1` |
| After a 1200 bps touch | `2886:0057` | 33 | **`unregistered`** |

Same physical board. Same chip serial, byte for byte. The wrapper lost track of a device it had registered minutes earlier, purely because the board presented a different USB identity — which is **normal, expected behaviour**: `2886:0057` is the stock Meshtastic application identity (`BusReportedDeviceDesc = T1000-E`), and `239A:xxxx` is the Adafruit bootloader family. Every nRF52 board alternates between at least two USB identities by design.

Downstream effect: `confirm` refused with *"no bootloader port appeared within 20s (expected a new 239A port with PID != 8029)"* while the board sat plainly visible on `2886:0057` with the matching serial. The flash is blocked, and the device cannot even be resolved for a Tier-B serial read.

## Why the obvious fix is the wrong one

[...]

PR #811, opened 2026-08-18
fix: pio-flash bootloader discovery for devices whose USB port survives download entry

### LINKED ISSUE #810: task: land the pio-flash same-port bootloader discovery fix
Landing task for the `_discover_bootloader_port()` same-port fix. Parent **#807** carries the defect analysis and stays open until the fix merges.

Branch: `fix/807-bootloader-same-port`
Commit: `77dcfdf`

## What lands

Fall back to the still-present running port when no port transition occurs, gated on **serial equality only** — no class matching, no relaxed ambiguity refusals. Plus a post-deadline re-check so a slow genuine transition wins over the same-port fallback.

## Verification

- 8 new cases in `scripts/test_pio_flash_bootloader_discovery.py`
- 3 of them fail against pre-fix code and pass after (confirmed by stashing the fix, not assumed): same-port discovery, duplicate-serial refusal, slow-transition race
- Full pio-flash suite: 17 passed, 1 failed — the failure is `test_legacy_hash_still_matches_when_no_serial_recorded`, verified failing on `fd8df8cb` so it predates this change and #808
- Gemini 2.5 Pro adversarial review: fallback cannot select a wrong board, #503 guarantee intact; its one finding (the race) was fixed rather than dismissed

Exists to satisfy the pre-push open-issue gate. Parent 807 closes on merge, owner-authorized.

Strycher:
Closes #810. Parent **#807** stays open until this merges.

## The defect

`_discover_bootloader_port()` built its candidate set by subtracting the pre-trigger port list, then applied every match rule to that set alone:

```python
new = [p for p in now if p["com"] not in before_coms]
```

That requires the device to **change COM ports** entering download mode. True for bridge-attached boards — the CP2102/CH340 keeps its own port while the SoC behind it reboots — and for nRF52, where the app CDC port is replaced by a separate DFU port.

It is not true when the USB device **is** the SoC. On ESP32-C3/C6/S3 in USB-Serial/JTAG mode the download-mode peripheral is in the same silicon: the port does not drop, the PID stays `303A:1001`, and the serial is unchanged. Enumeration before and after the trigger is byte-identical, so `new` is empty by construction, no candidate can ever be evaluated, and the function spins for the full timeout before refusing a device that is sitting right there.

Observed on `rcc6-bench-1` (ESP32-C6): `bootstrap` and `backup` both connect fine — they let esptool do its own reset-and-connect — while `confirm` refused with *"no bootloader port appeared within 20s"*. Reported as the same wall on `rc32-bench-1` (ESP32-S3).

## The fix

After the transition loop expires, fall back to the still-present running port, gated on **serial equality only**.

Identity is not relaxed. This is the same guarantee the fast path uses, and the fallback explicitly **cannot** match on vendor+PID — a class match still may never claim the flash target (#503). Ambiguity still refuses rather than picks.

A same-port match cannot confirm download mode was actually entered, since enumeration is identical either way. It only establishes which device is on which port. esptool syncs next and fails loudly if the ROM loader is not answering, so a failed trigger surfaces there rather than being silently flashed over — and the code says so out loud rather than implying a verified transition.

## Review findings — fixed, not dismissed

Gemini 2.5 Pro reviewed adversarially against "can this select the wrong physical board". It confirmed the fallback cannot, and that #503 is intact:

> *"This new path is strictly more secure than the pre-existing fallback for new ports (which allows a class-based match on serial-less ports). By adding an identity-only path, you have made the tool safer and more correct, not less."*

It raised **one race**: a slow device completing its transition just as the loop expires would have had the stale app-mode port claimed by the fallback, failing esptool sync for a device that was about to be ready. Safe, but a spurious failure.

Closed by taking one final enumeration after the deadline and checking it for a transitioned port **before** considering the same-port fallback. The loop body is extracted to `_match_transitioned()` so both paths share one implementation instead of duplicating the match rules.

Audit log: `docs/llm-consultations/2026-08-18-807-review-gemini-gemini-2.5-pro.log`

## Tests

8 cases in `scripts/test_pio_flash_bootloader_discovery.py`.

**Three fail against the pre-fix code and pass after** — verified by stashing the fix and re-running, not assumed:

- `test_same_port_device_is_found_by_serial`
- `test_same_port_fallback_refuses_on_duplicate_serial`
- `test_slow_transition_wins_over_same_port_fallback`

The rest pin what must not regress: no serial means no fallback; a differing serial is never claimed by class; genuine transitions still win; the serial-less changed-PID path still works; an absent device still refuses.

Unit tests rather than hardware because reproducing this needs a native-USB part **and** a bridge part **and** an nRF52 cycled through download mode per run. Behaviour is purely a function of two port lists. Hardware verification is complementary, not replaced — and is the next step on `rcc6-bench-1`.

## Pre-existing failure, not from this change

`scripts/test_pio_flash_device_match.py::test_legacy_hash_still_matches_when_no_serial_recorded` fails on `firmware-base` today. Verified failing against `fd8df8cb`, so it predates both this change and #808. Untouched here.

Worth noting separately: #808 merged reporting `ci-green 48/48` with that test already red, which suggests the matrix does not run the `scripts/test_pio_flash_*.py` suite at all. Filing separately rather than widening this PR.

### DIFF
--- a/scripts/pio-flash.py
+++ b/scripts/pio-flash.py
@@ -981,8 +981,12 @@ def _discover_bootloader_port(before: list, vendor: str, running_pid: str,
     running_serial = norm_serial(running_serial)
     deadline = time.time() + timeout
     last_seen: list = []
-    while time.time() < deadline:
-        now = enumerate_ports()
+
+    def _match_transitioned(now: list):
+        """Match a device that DID change ports. Returns the port, or None.
+
+        Refuses outright on ambiguity -- never returns a guess."""
+        nonlocal last_seen
         new = [p for p in now if p["com"] not in before_coms]
         if running_serial:
             by_serial = [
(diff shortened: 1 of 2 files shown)


Strycher:
Human approval received in chat: "merge #811". Enabling auto-merge (rebase); GitHub will merge once `ci-green` passes.

Agent: GreenTower (session c161ba8f)

Issue #779, opened 2026-08-17
task: zero the reply buffer at the four handleCommand call sites (fix for #765)


Strycher:
Parent: #765. Epic: #774. **Blocked by #775** — the guard lands, advisory, first.

**Files: 4** — one line each, the same idiom applied uniformly:

| File | Transport | Current defect |
|---|---|---|
| `examples/simple_repeater/MyMesh.cpp` | over-the-air | `temp[166]` zeroed only on the `is_retry` branch |
| `examples/simple_room_server/MyMesh.cpp` | over-the-air | zeroed on every branch except the admin dispatch |
| `examples/simple_sensor/SensorMesh.cpp` | over-the-air | never zeroed on any path |
| `examples/simple_sensor/main.cpp` | **serial** | missing `reply[0] = 0` |

Invariant to establish: **the reply buffer is empty before dispatch**, so "handler wrote nothing" always yields "no reply sent" on every transport — which is what the correct callers already do.

## Sizing note (deliberate deviation, flag if you disagree)

4 files exceeds the 1–3 target. Kept as one task because it is four applications of a single one-line idiom establishing one invariant; splitting it would put the same invariant in two PRs and invite one half landing without the other. Meets the rule's intent — single session, single PR, zero design decisions. No file overlap with #778, so the two can run in parallel.

## Acceptance

- [ ] All four call sites zero the buffer before dispatch, not per-branch.
- [ ] A handler that writes nothing produces no transmitted reply on either transport.
- [ ] Full CI matrix green, including a sensor and a room_server env.

---
Agent: **FoggyCanyon** (Agent Mail `app-c-dev-crosswire`, session `2d2b37f1`)

Issue #778, opened 2026-08-17
task: restore the three empty get bodies in CommonCLI (fix for #764)


Strycher:
Parent: #764. Epic: #774. **Blocked by #775** — the guard lands, advisory, first.

**Files: 1** — `src/helpers/CommonCLI.cpp`

- Restore the bodies onto the `isKey` arms for `radio`, `agc.reset.interval`, `pwrmgt.bootmv`.
- Delete the shadowed `memcmp` duplicates left behind by #299's conversion.
- Convert the two arms #299 missed — `cad`, `extra.sf` — so they stop prefix-matching junk suffixes.
- Fold the duplicated `radio.rxgain` arms into one. Rationale: `rx_boosted_gain` is an unconditional field and its `set` counterpart is unguarded, so the getter should be unconditional too. Upstream MeshCore agrees.

No design decisions remain; all resolved on #764.

## Acceptance

- [ ] `get radio` returns `freq,bw,sf,cr`; the other two return their values.
- [ ] `get radioX` / `get cadfoo` reach the `??: %s` fallback, not a prefix match.
- [ ] The #775 guard, already on `firmware-base`, goes quiet on this file.
- [ ] Full CI matrix green.

Touches `CommonCLI.cpp`, which #312 and #359 also touch — per the File-Convergence Rule those sequence against this rather than running beside it.

---
Agent: **FoggyCanyon** (Agent Mail `app-c-dev-crosswire`, session `2d2b37f1`)

Issue #776, opened 2026-08-17
chore: decide advisory vs required for the CLI dispatch guard in config-lint


Strycher:
Parent epic: #774. Blocked by #775 (guard lands advisory) and the #764/#765 fixes (log goes quiet).

**OWNER DECISION — not an agent's to make.** This issue exists because an agent already made it once without asking, in a code comment, and it had to be rolled back.

`config-lint` feeds `ci-green`, a required merge gate for every PR in this repo (#477). Promoting a check inside it to blocking changes fleet-wide merge gating: every other session inherits it on their next rebase. That is a governance change, not a test change.

**One-line diff** once decided — remove `continue-on-error: true` from the guard step.

## Preconditions before this can be decided

- [ ] #775 merged, guard advisory, CI log showing it reporting findings on the unfixed tree.
- [ ] #764 / #765 merged, same log now quiet.
- [ ] Agent Mail notice to all active sessions BEFORE the flip, not after — the #716 precedent.

## Prior art for the decision

`check_env_capability_claims.py` is deliberately advisory pending #648, on the stated grounds that a check built on unvalidated assumptions must not block other people's work. Whether this guard has cleared that bar is the question to answer here.

---
Agent: **FoggyCanyon** (Agent Mail `app-c-dev-crosswire`, session `2d2b37f1`)

Issue #774, opened 2026-08-17
epic: CLI dispatch and reply integrity — the chain does what the source says, and every reply is well-formed


Strycher:
**What:** the CLI dispatch chain does not reliably do what its source says, and replies are not reliably well-formed.
**Where:** `src/helpers/CommonCLI.cpp` primarily, plus the `handleCommand` call sites in each role.
**Why now:** a user upgrading from MC 1.15.0 stock ran `get radio` in the client app and got mojibake back. That was neither a display bug nor a prefs-migration failure — it was uninitialised stack transmitted over the air because the handler wrote nothing.

## The pattern this epic exists to close

Six issues filed independently over five weeks turn out to be one theme. Two independent failure modes compose into it:

1. **The chain silently does not match the source.** A handler is present, compiled in, and reachable-looking, but never runs — because a preceding arm claimed its key, or its body was left behind on a shadowed arm, or a contiguous window of arms fell out of the chain entirely. No compiler warns; nothing is dead code by any static measure.
2. **A reply is not guaranteed to be well-formed.** No caller-side contract requires the buffer to be initialised, size-bounded, or written at all. So "handler wrote nothing" becomes "transmit whatever was on the stack," and reply formatting uses unbounded `sprintf`/`strcpy` with no size plumbing.

Neither mode is visible on the serial console, because that caller zeroes its buffer and suppresses empty replies. Both are visible over the radio, which is the transport hardest to observe and the one real users are on.

## Children

| Issue | Opened by | Type | State |
|---|---|---|---|
| #764 | FoggyCanyon | bug P1 | three `get` keys (`radio`, `agc.reset.interval`, `pwrmgt.bootmv`) return an empty reply |
| #765 | FoggyCanyon | bug P1 | four `handleCommand` call sites transmit uninitialised stack when a handler writes nothing |
| #657 | FoggyCanyon | bug P2 | `led` / `display` handlers unreachable on the 1.17.0 base despite being compiled into the binary |
| #359 | CloudyCliff | task P2 | CLI reply formatters use unbounded `sprintf` with no buffer-size plumbing |
| #312 | *unstamped* | bug P2 | unbounded `strcpy` of command text into `tmp[132]` |
| #775 | FoggyCanyon | task P1 | static dispatch guard — lands advisory, BEFORE the fixes it validates |
| #776 | FoggyCanyon | chore P2 | advisory-vs-required gating decision (owner's, not an agent's) |
[...]

PR #816, opened 2026-08-18
feat(#775): advisory CLI dispatch guard — lands before the fixes it validates

### LINKED ISSUE #775: task: static CLI dispatch guard — must fail on the pre-fix tree before any fix lands
Parent epic: #774

**Deliverable:** `scripts/check_cli_dispatch.py` + `scripts/test_cli_dispatch.py` (2 files).

Two checks, scoped to contiguous `if`/`else if` chains in the known dispatch files:

1. **empty-body** — an arm matching a string key that writes nothing to `reply`.
2. **shadowed-key** — a key claimed twice in one chain where the earlier arm tests nothing but that key, so the later arm can never see its own key.

Chain-aware and arity-aware: `home`/`default` split by argument count, and `ExtraFS/` across three separate handlers, must stay green. Escape hatch `// cli-dispatch-guard: allow-empty <reason>` for a deliberately empty arm.

## Sequencing — this is the point of the task

**This lands BEFORE the fixes it validates, and it lands advisory.**

A guard introduced alongside its fix proves nothing: green only shows it does not false-positive on corrected code, and CI never runs it against the broken tree. The required artifact is **a CI log showing the guard reporting findings against the unfixed tree** — which cannot be produced once the fix has landed.

Advisory (not blocking) because a red PR cannot merge, and the whole purpose is to land it while it is still reporting findings.

## Acceptance

- [ ] Unit tests cover the true positives AND the false positives an earlier draft produced (a guard that cries wolf gets switched off).
- [ ] Wired into `config-lint` with `continue-on-error: true`.
- [ ] **CI log on this PR shows the findings against the unfixed tree** — linked in a comment here as the evidence.
- [ ] Promotion to blocking is explicitly NOT in this task. See the separate gating decision.

---

Strycher:
Closes #775. Under epic #774.

**Read this before reacting to the red ✗:** the advisory scan step in `config-lint` is *expected to fail on this branch*. That failure is the deliverable. `ci-green` still passes.

## Why this lands first, and alone

The fixes are deliberately **not** in this PR.

A guard that lands in the same change as the bug it catches proves nothing. Green would only show it doesn't false-positive on already-corrected code — CI would never once have run it against the broken tree. The evidence a reviewer can actually verify is **a CI log of this guard firing on the unfixed tree**, and that log stops being producible the moment #778 lands.

A blocking check cannot land red. So the guard goes in advisory, reports, and the promotion to required is decided separately.

```
this PR reports 8 findings  →  #778 / #779 land  →  the log goes quiet  →  #776 decides required-or-not
```

#776 is an owner decision, not an agent's: `config-lint` feeds `ci-green`, a required merge gate for every PR in this repo (#477), so promoting a step to blocking changes fleet-wide merge gating.

## What it checks

Two things, scoped to contiguous `if`/`else if` chains in the known dispatch files:

| | |
|---|---|
| **empty-body** | an arm matching a string key that writes nothing to `reply` |
| **shadowed-key** | a key claimed twice in one chain where the earlier arm tests nothing but that key, so the later arm can never see its own key |

Chain- and arity-aware. `home`/`default` split by argument count, and `ExtraFS/` appearing in three separate handlers, are pinned by tests as must-stay-green — an earlier draft flagged all three, and a guard that cries wolf gets switched off. Escape hatch for a deliberately empty arm:

```c
} else if (isKey(config, "x")) {   // cli-dispatch-guard: allow-empty <why>
```

`DEFAULT_TARGETS` is **not** globbed — a new dispatch file is a deliberate addition, not a silent omission.

## Known limits, documented in the module

- Keys must be string literals; a macro or ternary is invisible. No dispatch arm in the tree spells its key any other way.
- An arm whose only write sits inside `#ifdef` is not reported, because the directive lines make the body non-empty. A separate scan for exactly that case found **no instances** — every conditional arm carries an `#else`.

## What it found here — the bug it exists for

`get radio`, `get agc.reset.interval` and `get pwrmgt.bootmv` have returned an empty reply since **2026-07-18**. #299's `memcmp`→`isKey` conversion hoisted those three guards without moving their bodies; the bodies stayed on arms the new guards now shadow.

It compiles. No compiler calls it dead. And the serial console hides it — that caller zeroes its buffer and suppresses empty replies, so on serial the symptom is silence. Only the radio transport showed it, as uninitialised stack, because that caller never zeroed its buffer (#765). A user upgrading from stock MeshCore 1.15.0 saw mojibake in the client app and reasonably assumed his prefs migration had failed. It had not.

## Verification

- Scan reports **8 findings, exit 1** against `da56a157` — current tip, not a stale base.
- JSON artifact is written *before* that exit, so the report survives the failing step and uploads.
- Self-test **17/17**.
- `ci-green` passes; both steps are `continue-on-error`.
- Gemini review run against SAFELANE + CLAUDE-BASE + the full files, with no question list supplied — returned no findings.

## Not closed by this PR

#764 and #765 stay open. They are the bugs; #778 and #779 are their fixes and are blocked on this landing first.

---

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -73,6 +73,33 @@ jobs:
       - name: Font generator self-test
         run: python scripts/test_gen_jbm_font.py
 
+      # #775 (epic #774). Guards the defect class behind #764/#765: an `else if`
+      # that matches a CLI key and writes nothing, or a key claimed twice in one
+      # chain so the later arm is unreachable. `get radio` returned an empty reply
+      # for ten months that way -- it compiles, no compiler calls it dead, and the
+      # serial console hides it because that caller zeroes its buffer and skips
+      # empty replies. Only the radio transport showed it, as stack bytes.
+      #
+      # ADVISORY, deliberately. The fixes (#778/#779) have NOT landed yet, so this
+      # step is expected to REPORT FINDINGS here -- that log is the evidence the
+      # guard catches the real defect, and it cannot be produced once the tree is
+      # fixed. A blocking check cannot land red. Promotion to required is a
+      # separate owner decision, tracked as #776.
+      - name: CLI dispatch guard self-test (advisory)
+        continue-on-error: true
+        run: python scripts/test_cli_dispatch.py
+
+      - name: No empty or shadowed arms in the CLI dispatch chains (advisory)
+        continue-on-error: true
+        run: python scripts/check_cli_dispatch.py --json-out "$RUNNER_TEMP/cli-dispatch.json"
+
+      - name: Upload CLI dispatch report
+        if: always()
+        uses: actions/upload-artifact@v4
+        with:
+          name: cli-dispatch
+          path: ${{ runner.temp }}/cli-dispatch.json
+
       # ADVISORY until #648 (the capability inventory) validates what this asserts.
       # ci-green is a hard merge gate for every PR in the repo (#477), so a check
       # built on unvalidated assumptions must not block other people's work. The
--- a/scripts/check_cli_dispatch.py
+++ b/scripts/check_cli_dispatch.py
@@ -1,0 +1,332 @@
+#!/usr/bin/env python3
+"""Static guard for the CLI dispatch chains (#764).
+
+`d4d417ee` converted `handleGetCmd` from prefix `memcmp` to exact-token `isKey`.
+For most keys it hoisted the new guard *and* moved the body. For three keys it
+hoisted the guard and left the body behind on the old arm:
+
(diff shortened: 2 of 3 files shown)


Strycher:
Human approval received from the repo owner in session: "merge it."

Recording per SAFELANE §3 — the human makes the decision, automation performs the merge. Enabling auto-merge with `--auto --rebase`; no manual or force merge.

State at approval: 48/48 checks SUCCESS, `ci-green` green, `mergeable`. The advisory scan step reported its 8 findings as intended and did not gate — evidence captured on #775 before merge, since it cannot be reproduced once #778/#779 land.

---

Issue #746, opened 2026-08-16
bug: esp-mqtt auto-reconnect was never disabled — brokers reconnect behind the pool, bypassing backoff/budget/rotation (latent, since #53)


Strycher:
## Summary

esp-mqtt's client-level **auto-reconnect defaults ON and was never disabled**, so it reconnects error-disconnected brokers **behind the pool's back** — firing `MQTT_EVENT_CONNECTED` → `onConnected` independent of the pool's rotation, budget, and backoff logic. This is a latent violation of the #53 design ("the pool owns the connect lifecycle; esp_mqtt will not auto-retry"): the comment asserted the behaviour, the config never enforced it.

Pre-existing since the observer's esp-mqtt integration. Not introduced by #739 — but #739 is what made it **visible**, because #739 added a state (`Failed`) that must stay sticky and the auto-reconnect resurrected it.

## Evidence

Soak on hv3-bench (`task/739-flaky-broker-penalty`, before this fix), 3 TLS brokers, two of them flaky (MeshMapper, CoreComms):

- s3 (CoreComms, 140 connection errors) reached terminal `BrokerState::Failed` — correct.
- **But s3 then rotated out 20 times *after* it first went Failed.** A `Failed` broker is excluded from the pool's re-drive candidate set and `tryConnect` no-ops it, so *neither pool promotion path* touched it — yet it kept reaching `Up`.
- The only remaining path: `onConnected` fired from an esp-mqtt event with no state guard (`rt_.state = BrokerState::Up;` unconditional). The broker's client was still allocated (releaseClient only runs on rotate-out), so esp-mqtt auto-reconnected it and un-Failed it.

Log preserved: `hv3-soak-739-v1-leaky-2026-08-16.log`.

## Why it matters beyond #739

Any error-disconnected broker keeps its client and is auto-reconnected by esp-mqtt on a ~10 s timer, regardless of:

- **Backoff** — the pool's escalating backoff schedule is bypassed; esp-mqtt reconnects on its own 10 s cadence.
- **Budget** — `OFFBAND_MAX_LIVE_TLS` admission is bypassed; a broker can re-establish a live mbedTLS context without the pool's budget check.
- **Rotation** — a broker the pool intends to keep parked (`HeldBudget`) can reconnect itself.

[...]

PR #817, opened 2026-08-18
bug: flaky TLS broker steals equal rotation budget — sticky penalty + terminal Failed + client reaper (#739)

### LINKED ISSUE #739: bug: a flaky TLS broker steals equal rotation budget — backoff resets on intermittent success and never terminally disables
## Observed

4.6 h soak on hv3-bench (`offband-v1.5.0-beta1-7-g1dc2d21`, the beta2 candidate), 3 TLS brokers enabled behind `OFFBAND_MAX_LIVE_TLS=1`:

| Slot | Broker | Auth | UP | Backoff | Errors |
|---|---|---|---|---|---|
| 1 | mqtt2.okimesh.org | anon | 275 | **0** | **0** |
| 2 | MeshMapper | jwt | 282 | 229 | 156 |
| 3 | CoreComms.net | jwt | 279 | **536** | 305 |

Broker 0 (plaintext, exempt from rotation) and broker 1 (anon wss) were flawless. The two external JWT brokers were **intermittently** flaky — each reached UP ~as often as the healthy broker AND held full 60 s dwells when connected, but each also failed hundreds of times (s2: mbedTLS handshake errors `class=3`; s3: connection errors `class=4 connack=0`).

**Net effect: 46% of the soak had NO TLS broker up**, and one ~38-minute window drove the publish ring past its depth → packet drops on the healthy rotating broker. (Broker 0 dropped nothing — the always-on primary was complete. See the payload assessment doc / #710.)

## Root cause of the amplification

The trigger is external (flaky remote brokers), but the firmware treats a flaky broker the same as a healthy one, so a single bad broker degrades the whole TLS feed. Two mechanisms:

**1. Backoff resets fully on any success (`MqttBroker::onConnected` sets `retry_count = 0`).**

```cpp
uint32_t brokerBackoffMs(uint32_t retry_count) {
    static const uint32_t kSchedule[] = {5000, 15000, 30000, 60000, 120000};
    ...
}
void MqttBroker::onConnected(uint32_t now_ms) {
    ...
    rt_.retry_count = 0;   // <-- a single success wipes all accumulated penalty
}
```

The escalating schedule works for a *dead* broker but not a *flaky* one: an intermittent broker keeps succeeding just often enough to reset `retry_count` to 0, so it oscillates near the low end (5-15 s) of the schedule forever and keeps claiming rotation turns equal to a reliable broker. It is never meaningfully penalized for a 50%-failure rate.

**2. No terminal state.** The schedule caps at 120 s and retries forever. A permanently-unreachable configured broker churns at 120 s intervals indefinitely, consuming a rotation turn each cycle (allocate ~55 KB mbedTLS context -> fail -> free), and nothing tells the operator the broker is unusable.

## Proposed fix (owner-directed)

A failing broker gets a **gradually increasing backoff until it succeeds** — that resists resetting on intermittent success — and, past a threshold where continued retry is pointless, is **disabled and errors loudly** so the operator knows a configured broker is unusable rather than silently stealing rotation budget.

Concretely, to design:

- **Penalty that survives intermittent success.** Track recent failure rate (e.g. an EWMA or a windowed failure count), not just a `retry_count` that a single success zeroes. Backoff derived from that so a flaky broker escalates even while occasionally connecting. Optionally: only fully reset after M *consecutive* clean dwells.
- **Terminal disable + loud error.** After N escalations (or T total unreachable time), stop retrying, mark the broker a distinct terminal state (not silent `Down`), and surface it: `mqtt status` field, a serial/caplog line, and the client wire contract (coordinate like #715). "Disabled" here means removed from the rotation candidate set until the operator re-enables or reconfigures it -- it must not look like an operator `disable`.
- **Rotation-budget fairness.** The higher-order goal: a flaky/failing broker must not consume the 1-wide TLS budget on equal terms with healthy brokers. Escalating sticky backoff achieves most of this by making the flaky broker eligible less often; the terminal state finishes it.

## Not this issue

- #720 (a broker reconnecting *faster* than the dwell resets `went_up_ms` and blocks rotation) is a related rotation-robustness gap but a different mechanism.
- Ring depth (#710/#727) is not the fix -- steady-state drops were zero; the drops were the *symptom* of the TLS outage, not a sizing problem.
- Why MeshMapper/CoreComms specifically are flaky (path, server load, JWT, or handshake-time heap pressure at the observed ~19 KB heap min) is worth a separate look but is external to this firmware change.

## Acceptance

- A broker failing on a fixed fraction of attempts escalates its backoff over time instead of oscillating at the schedule floor.
- A permanently-unreachable configured broker reaches a terminal disabled+errored state within a bounded time and stops consuming rotation turns.
- The terminal state is visible to the operator (status + log), distinct from an operator-initiated disable.
- Native test over the promotion/backoff decision (the #708 `BrokerRotationSelect.h` pattern) covering: healthy, intermittent-flaky, and permanently-dead brokers.

Strycher:
Closes #739. Refs #746 (auto-reconnect finding — addressed by the reaper, not the global disable; leaving its closure to owner judgment).

## The bug (#739)

The only backoff state was `retry_count`, reset to 0 on **any** successful connect. Fine for a *dead* broker, wrong for a *flaky* one: an intermittent broker connects just often enough to zero its penalty, so it oscillates at the 5s backoff floor forever and keeps claiming an equal rotation turn. Measured on `hv3-bench`: two ~50%-failing remote JWT brokers held **46% of a soak with no TLS broker up**.

## The change (3 commits)

1. **`BrokerHealth.h`** — dependency-free, host-testable policy object. `fail_penalty` is **sticky**: a single success does not clear it; only `OFFBAND_BROKER_SUCCESS_BAR` (3) consecutive clean full-dwells rehabilitate. Past `OFFBAND_BROKER_TERMINAL_PENALTY` (8) → terminal `Failed`.
2. **Lifecycle wiring** — `noteFailure()` at the three failure sites; backoff derives from `fail_penalty`; `onCleanDwell()` feeds rehab.
3. **Failed handling + reaper** — `onConnected()` ignores a late CONNECTED for a `Failed` broker (a racing esp-mqtt reconnect can't resurrect it); `tryConnect()` refuses to start `Failed`; the pool worker **reaps** (`releaseClient`) any `Failed` broker so its ~60KB TLS context is freed and it stops holding a client / a rotation turn.

## Why auto-reconnect stays ON (the reverted over-correction)

An earlier attempt on this branch globally set `disable_auto_reconnect`. A **~7h hardware soak** showed that terminally failed a *healthy* broker and collapsed rotation (`live=0/1` up to 96%). It was reverted; the pool's own reconnect isn't robust enough to be the sole driver yet. Only the `Failed` terminal state is reaped. **That catastrophic intermediate commit is collapsed out of this branch's history** — it does not land.

## Validation

- **`test_broker_health` 7/7** (native), incl. `FlakyAlternatingNeverRehabilitatesAndEscalates` reproducing the soak scenario.
- **Gemini 2.5-pro reviewed** — safe to merge; all 6 adversarial checks pass (healthy broker can't get stuck `Failed`; reaper/auto-reconnect race handled with no leak/corruption; `client_lock_` discipline maintained; `Failed` correctly leaves rotation; state machine correct). No bugs found.
- **~7h `hv3-bench` soak on this exact build**: `live=1/1` **99.3%** (2190/2205 samples), 329 clean rotate-outs, heap stable ~42K→~94K on rotate, **0 spurious terminal failures**. One lone transient TLS error exercised the penalty path correctly (+1 → Backoff, recovered — not terminal).

## Known scope

The **terminal-`Failed`-sticks** path (criterion 1) is covered by native tests + the reaper logic, but wasn't organically reached in the soak (no broker failed 8× — the config's brokers are reliable). Adjust-later is fine; the mechanism is unit-proven and Gemini-verified.

### DIFF
--- a/src/helpers/wifi_observer/BrokerHealth.h
+++ b/src/helpers/wifi_observer/BrokerHealth.h
@@ -1,0 +1,79 @@
+#pragma once
+
+#include <cstdint>
+
+// Broker connection-health tracker (#739).
+//
+// Dependency-free (no Arduino/ESP headers) so it builds under [env:native] and
+// is unit-testable in isolation -- same pattern as MqttRingLog.h and
+// BrokerRotationSelect.h. The pool owns a BrokerHealth per broker and feeds it
+// two events; this owns the POLICY of how a broker's reliability translates into
+// backoff, rehabilitation, and a terminal give-up.
+//
+// WHY THIS EXISTS. Before #739 the only backoff state was retry_count, reset to 0
+// on any success (MqttBroker::onConnected). That works for a DEAD broker but not
+// a FLAKY one: an intermittent broker succeeds just often enough to zero its
+// penalty, so it oscillates at the 5 s schedule floor forever and keeps claiming
+// rotation turns equal to a reliable broker. Measured on hv3-bench: two remote
+// JWT brokers at ~50% failure kept 46% of the soak with NO TLS broker up.
+//
+// The penalty is STICKY IN BOTH DIRECTIONS:
+//   - a single success does NOT clear it (only a proven clean streak does), so a
+//     flaky broker cannot cheaply reset;
+//   - a broker must earn its way back with kSuccessBar CONSECUTIVE clean dwells,
+//     at which point it is fully rehabilitated to a healthy running state.
+// A failure escalates the penalty and zeroes any clean-streak progress. Past
+// kTerminalPenalty the broker is declared Failed -- dropped from rotation and
+// surfaced loudly -- instead of churning a rotation turn forever.
+
+namespace offband {
+
+// Consecutive clean full-dwells required to fully rehabilitate a broker whose
+// penalty is non-zero. One dwell is ~60 s, so 3 = ~3 min of unbroken service.
+#ifndef OFFBAND_BROKER_SUCCESS_BAR
+  #define OFFBAND_BROKER_SUCCESS_BAR 3u
+#endif
+
+// Penalty level at which a broker is declared terminally Failed and removed from
+// the rotation candidate set. The backoff schedule has 5 steps (5..120 s); this
+// gives several full escalations before giving up.
+#ifndef OFFBAND_BROKER_TERMINAL_PENALTY
+  #define OFFBAND_BROKER_TERMINAL_PENALTY 8u
+#endif
+
+struct BrokerHealth {
+    uint16_t fail_penalty = 0;   // escalation level; indexes the backoff schedule
+    uint16_t clean_streak = 0;   // consecutive clean full-dwells since last failure
+
+    // A connection failed (disconnect-with-error, handshake failure, refused).
+    // Escalates the penalty and discards any clean-streak progress -- a single
+    // failure means the broker has NOT proven stable.
+    void onFailure() {
+        if (fail_penalty < 0xFFFF) fail_penalty++;
+        clean_streak = 0;
+    }
+
+    // The broker held the budget for a FULL dwell and was rotated out cleanly
+    // (no error). This is the only event that earns rehabilitation -- NOT a bare
+    // connect, which a flaky broker does constantly. Only kSuccessBar of these in
+    // a row clears the penalty.
+    void onCleanDwell() {
+        if (fail_penalty == 0) return;        // already healthy, nothing to earn
+        if (++clean_streak >= OFFBAND_BROKER_SUCCESS_BAR) {
+            fail_penalty = 0;                 // rehabilitated
+            clean_streak = 0;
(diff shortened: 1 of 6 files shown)


Strycher:
Human approval received from Ben in session 82e07da7 ("Seems like where we are at is worth push/PR/merge, with opportunity to adjust later"). Re-validated against remote: 3 commits (BrokerHealth + wire + reaper), 6 files, base firmware-base. Native test_broker_health 7/7, observer build SUCCESS, Gemini 2.5-pro safe-to-merge, ~7h hv3 soak live=1/1 99.3%.

Issue #777, opened 2026-08-17
task: epic #774 integration test — full get/set surface over BOTH serial and the over-the-air client path


Strycher:
Parent epic: #774. **Final task in the epic — must close before #774 can close.**

Per CLAUDE-BASE Epic Completion Protocol: individual task tests prove each piece works in isolation; this proves they work together. Human sign-off only — agents cannot close the epic.

## What this verifies, and what it deliberately does not

This issue proves the reported bug is **dead on hardware**, by flashing one board pre-fix and post-fix and comparing. That is all it does.

It does **not** sweep the full `get`/`set` surface. An earlier version of this issue did, which was scope I wrote myself and which does not belong here: the sweep needs tooling that does not exist yet (batch send, redaction at source, latency capture) and is tracked separately. Growing a verification task into a tooling project is what that separation exists to prevent.

The over-the-air leg cannot be a sweep in any case — the #765 defect only fires when a handler writes *nothing*, and after #778/#779 none do. What it *can* prove is that the reported symptom is gone, and that takes one command on two firmwares.

## Scope — over-the-air before/after (owner-driven, three messages)

This is the only step that needs a human, because the over-the-air CLI is gated on `client->isAdmin()` (`simple_repeater/MyMesh.cpp:705`) and admin comes from matching `_prefs.password` (`:105`). The agent drives every flash and every serial check; the owner sends the messages from the client app, exactly as the reporting user did.

Target: a bench board with the shipped default `ADMIN_PASSWORD "password"` (`simple_repeater/MyMesh.cpp:36`) — a public compile-time default, not a credential.

**On PRE-fix firmware (before #778/#779):**

- [ ] `get radio` → **expected: mojibake / garbage text.** This is the reproduction. Without it, everything below only shows that correct firmware looks correct.

**On POST-fix firmware:**

- [ ] `get radio` → a well-formed `freq,bw,sf,cr` reply.
- [ ] `get cadfoo` → `??: cadfoo`, confirming the prefix-match hardening rather than a `cad` value.

## Boards

| Board | Role | Why |
|---|---|---|
| T096 | repeater | bare, shipped default password — the before/after target |
| ST-P | repeater | second ESP32, serial sweep |
| RAK4631 | repeater | **nRF52** — the only leg where `get pwrmgt.bootmv` / `get pwrmgt.bootreason` return real values instead of `ERROR: Power management not supported` |

[...]

PR #821, opened 2026-08-18
fix(#778): restore the three empty get bodies — the guard goes quiet

### LINKED ISSUE #778: task: restore the three empty get bodies in CommonCLI (fix for #764)

Strycher:
Closes #778. Fixes the code half of #764. Under epic #774.

**This is the PR the #775 guard was landed early to validate.** Same script, same commit, before and after:

| Tree | `check_cli_dispatch.py` |
|---|---|
| `firmware-base` @ `ecc3fdb0` | **8 findings**, exit 1 — [CI log](https://github.com/OffbandMesh/meshcore-firmware/actions/runs/32111156502/job/95630686528) |
| this branch | **clean**, exit 0 |

That before/after pair is the reason the guard shipped first. Had it landed with this change, green would only have shown it doesn't false-positive on already-corrected code.

## The fix

`get radio`, `get agc.reset.interval` and `get pwrmgt.bootmv` have returned an empty reply since **2026-07-18**. #299's `memcmp`→`isKey` conversion hoisted the guard for those three keys without moving the body; the body stayed on an arm the new guard now shadows. Each body is restored onto its `isKey` arm and the shadowed duplicates removed.

**The deleted `pwrmgt.bootreason` duplicate is not a loss.** The live handler is at line 1728 — and unlike the copy removed here, it carries the `#else` that answers on boards without `NRF52_POWER_MANAGEMENT`. The removed copy was the broken one.

## Three extras, owner-approved

1. **`cad`** and **`extra.sf`** were still prefix-matching, so `get cadfoo` answered with the `cad` value instead of falling through to `??: cadfoo`. Same defect class, same function — two arms #299 missed.
2. **The two identical `radio.rxgain` arms are folded into one.** One was wrapped in `#if defined(USE_SX1262) || defined(USE_SX1268) || defined(USE_LR1110)`, the other was not, so every board already compiled an unguarded handler and the conditional only ever gated a duplicate. `rx_boosted_gain` is an unconditional member (`CommonCLI.h:67`) and the setter at line 1270 is unguarded.

## Verification

**Builds 6/6.** `Heltec_WSL3_sensor`, `heltec_v4_repeater_telemetry`, `heltec_v4_room_server`, `heltec_v4_companion_observer_wifi`, `RAK_4631_repeater` (nRF52, exercising the `pwrmgt.bootmv` `#ifdef` path), and **`LilyGo_T3S3_sx1276_repeater`** — a board defining none of the three radio macros, built specifically to prove the `#if` removal cannot break a non-SX1262 target. `radio.rxgain` is present in that binary.

**Adversarial review.** Gemini was given SAFELANE, CLAUDE-BASE and the full files with no question list, and returned DO NOT MERGE on two points. Both were run down rather than argued away:

- *"You deleted the `pwrmgt.bootreason` handler."* Wrong — line 1728, above the diff hunk it could see.
- *"Removing the `#if` may break non-SX1262 builds."* A risk claim that structure alone couldn't settle, so the SX1276 board was built. It compiles.

Its third point — that the extras are scope — was put to the owner, who approved them.

## Not closed here

**#764 stays open.** This is its code fix, but the bug closes on the owner's call after #777, the epic integration test that exercises the full surface over **both** transports. Every defect in this epic was invisible on serial; a serial-only check would have passed while all of them were live.

**#765 is untouched** — the four callers that transmit uninitialised stack. That is #779, and it shares no files with this.

---


Strycher:
## The guard went quiet — verified in CI, not locally

| Tree | `check_cli_dispatch.py` in CI | Log |
|---|---|---|
| `firmware-base` @ `ecc3fdb0` | `8 finding(s) across 9 file(s)` | [run 32111156502](https://github.com/OffbandMesh/meshcore-firmware/actions/runs/32111156502/job/95630686528) |
| this branch @ `72ec3a48` | `check_cli_dispatch: 9 file(s) clean` | [run 32190119520](https://github.com/OffbandMesh/meshcore-firmware/actions/runs/32190119520/job/95882396726) |

Same script, same 9 files, two CI runs. Zero `empty-body` and zero `shadowed-key` lines in this one.

This is the artifact the sequencing existed to produce, and it could not have been produced by a PR that carried both the guard and the fix.

**CI: 48/48 SUCCESS**, no failures of any kind.

---

Strycher:

Re-validated against the remote immediately prior: **1 commit**, **1 file** (`src/helpers/CommonCLI.cpp`), `mergeable=MERGEABLE`, checks **48/48 SUCCESS**, base `firmware-base`. Unchanged from what was approved.

`mergeStateStatus=BEHIND` — base moved while CI ran. Updating the branch (server-side, no force-push), then auto-merge with rebase.

---

Issue #771, opened 2026-08-17
task: add heltec_rc32 envs to the release matrix


Strycher:
## Problem

`grep rc32 .github/release-envs.txt` returns nothing. The release pipeline produces **no
RC32 binary at all**, so there is nothing to hand a beta tester through the normal channel
(`offband-v*` tag -> release assets). RC32 is also absent from `ci.yml`, tracked separately.

This blocks giving RadioCore beta testers a versioned, reproducible image instead of an
ad-hoc file.

## Change

Add the four standard RC32 envs to `.github/release-envs.txt`, matching the pattern used by
every other board:

- `heltec_rc32_companion_radio_ble`
- `heltec_rc32_companion_radio_usb`
- `heltec_rc32_repeater`
- `heltec_rc32_room_server`

`heltec_rc32_companion_radio_wifi` stays **gated out**, same class as the other
`companion_radio_wifi` envs -- it bakes a placeholder `WIFI_SSID` at compile time and is
unblocked by #325.

The `_diag` / `_diag_nocrashlog` envs are diagnostic instruments and are deliberately not
released. `without_display`, `sensor`, `terminal_chat` and `kiss_modem` follow the existing
convention of not being part of the curated release set.

All four verified building locally before the change.

## Sequencing -- do NOT cut an RC32 release before #704 lands

`fix/704-rc32-no-gps` is unmerged and carries four commits the stock env needs:

    4c0d212b fix(#704): disable GPS on heltec_rc32 -- its enable pin is the VDD_SPI strap
    bfcafb97 fix(#704): strip peripherals the RC32 does not have
    25c1ef80 fix(#704): align heltec_rc32 to Heltec's own board definition
    3467ef54 fix(#704): DISPLAY_ROTATION=0 -- landscape, owner-confirmed

A GPS enable pin on the VDD_SPI strap is a boot hazard, and the stock env currently has GPS
**enabled**. Adding RC32 to the release matrix is correct on its own, but the first release
that includes it must be cut *after* #704 merges, or testers receive the unfixed config.

Note also that no RC32 companion env has been hardware-verified in its **stock** form -- all
recent bench work ran on `..._ble_diag`, which locally overrides `ENV_INCLUDE_GPS=0` and
`DISPLAY_ROTATION=0` and therefore masks exactly what #704 fixes.

Strycher:
[...]

Issue #769, opened 2026-08-17
task: land the MeshLog UART0 mirror + battery telemetry (PR)


Strycher:
PR-landing task for the MeshLog UART0 mirror work under parent issue 763.

## Scope

- Mirror MeshLog/caplog to raw UART0 in parallel with the capture ring.
- **Decouple the UART0 mirror from `g_meshLogEnabled`.** Recording (ring + Serial
  mirror) stays gated on capture; the wire emits independently. The level filter
  still applies to both. The early-out folds to the original single branch in
  stock builds via a `constexpr`, so builds without the flag are unchanged.
- `OFFBAND_POWER_TELEMETRY`: battery line every 30 s for the Heltec beta Q01
  runtime question.
- Remove `BLE_DEBUG_LOGGING=1` from the rc32 diag env (see Known issue).

## Hardware verification (rc32-bench-1)

- `setup:DONE` in 2 s.
- Telemetry live on the UART0 wire with capture never enabled — the decoupling
  working as intended, on a board with no host attached.
- Live discharge run: 4130 -> 4111 mV on a 2500 mAh cell.

## Known issue carried forward (NOT fixed here)

The UART0 mirror is a **synchronous bounded busy-wait** on the TX FIFO at 115200
baud. With `BLE_DEBUG_LOGGING=1` it blocked inside `setup()`: boot went 2 s ->
118 s -> not completing. Removing the debug flood restores normal boot, but the
underlying hazard remains for anyone enabling the mirror alongside a debug-level
firehose.

**Follow-up required before the mirror is enabled more widely: level-gate it**
(e.g. mirror only <= MLOG_ERROR by default, or make the UART0 level independently
configurable). Filed against parent 763.

Blast radius today is contained: `OFFBAND_MESHLOG_UART0` is only set on the rc32
diag env.

Strycher:
Work complete and hardware-verified on rc32-bench-1; closing to land the PR. Level-gate follow-up tracked on parent issue 763.

Issue #763, opened 2026-08-17
task: mirror MeshLog/caplog to UART0 so an external sniffer sees the live log on USB-CDC boards


Strycher:
Owner-requested. Supersedes the narrower "power telemetry over UART" idea — this delivers it as a side effect.

## Problem

On the 15 native-USB-CDC boards, the live log is unreachable exactly when you need it:

- On a `_usb` companion, `Serial` **is** the framed companion protocol, so the MeshLog mirror is
  deliberately OFF (`g_meshLogMirror` false, #411).
- On a `_ble` companion, `Serial` is a debug channel — but it is USB-CDC, so it **dies with the
  chip** and **attaching to it power-cycles the board** (#702). Every log ever captured that way
  came from a boot that succeeded.
- caplog captures to a ring, but that ring is retrieved over the companion API — i.e. over the
  same USB that is unavailable on battery — and it does not capture early boot anyway (earliest
  entry `[45102]` on `rc32-bench-1`, 2026-08-15).

Meanwhile **UART0 (GPIO43 on RC32, header pin 12) sits idle** on those boards and is already
wired to an external Feather sniffer, which survives the chip dying and does not reset it.

## Approach

`mesh_log_line()` (`src/MeshLog.cpp:116`) is the single tee point every console producer routes
through. It already: formats with a `[millis]` prefix, appends to the ring under lock, then
mirrors to `Serial` **non-blockingly** (`g_meshLogMirror && Serial.availableForWrite() >= total`,
per #447).

Add a **second, parallel mirror to UART0** at that same point, using the raw register-level writer
from `BootBeacon.h` (#740) — no driver, no FreeRTOS, no mutex, and bounded so it cannot block
(the #741 lesson).

New flag: **`OFFBAND_MESHLOG_UART0`**. Off by default.

## Guards, both load-bearing

1. **`ARDUINO_USB_CDC_ON_BOOT` only.** On a UART-bridge board (Heltec V3 = `esp32-s3-devkitc-1`)
   `Serial` **is** UART0, so mirroring there would double-print every line. Scope it to the boards
   where UART0 is genuinely idle.
2. **Requires `ENV_INCLUDE_GPS=0` on RC32.** `EnvironmentSensorManager::initBasicGPS()` calls
   `Serial1.setPins(44, 43)`, reassigning GPIO43 — U0TXD, the sniffer wire — partway through
   `setup()`. Documented on #702; must be called out wherever this flag is offered.

## What it unlocks

[...]

Issue #806, opened 2026-08-18
task: scaffold variants/heltec_rcc6 for the RCC6 (ESP32-C6)


Strycher:
Scaffold `variants/heltec_rcc6` so the RCC6 has a buildable, flashable target. Nothing exists in tree today.

## Pin source

Pins are taken from the **schematic's own printed net-to-GPIO table** (`docs/radiocore/vendor/RCC6/RCC6-L62_V1.0-schematic.pdf`, beside the RA62A symbol U5), not from row-pairing or third-party transcription. `[verified: rendered schematic, 2026-08-17]`

```
SPI_CS  GPIO23   SPI_MOSI GPIO22   SPI_CLK GPIO21   SPI_MISO GPIO20
DIO1    GPIO19   SELECT   GPIO7    RESET_N GPIO8    BUSY     GPIO10
VEXT_LDO_Ctrl    GPIO11
```

From the ESP32-C6 symbol (U8): `ADC_Ctrl`=**GPIO5**, `ADC_IN`=**GPIO6**, `USER_Key`=**GPIO9**, `D_N`/`D_P`=GPIO12/13 (native USB).

Cross-checks: matches `n30nex/NeonPocketMC-RCC6` `variants/heltec_rcc6` (MIT) on all seven LoRa pins and all three of the above, and matches `docs/radiocore/README.md:108`.

## Prior art

`n30nex/NeonPocketMC-RCC6` (MIT) ships a complete variant plus `boards/heltec_rcc6.json`. Reuse it as reference with attribution rather than deriving from scratch — see #804, which corrects the handoff's claim that no reference variant exists.

## Board definition

This repo's three existing C6 variants (`xiao_c6`, `lilygo_tlora_c6`, `m5stack_unit_c6l`) all use stock `esp32-c6-devkitm-1` with `min_spiffs.csv` for a 4 MB part. **RCC6 is 16 MB**, so it needs either a board JSON or explicit `board_upload.flash_size` / `maximum_size` overrides in the manner of `xiao_c6`.

## Known open items — do not silently inherit

- **`VEXT_LDO_Ctrl` (GPIO11)** gates `Vext_3V3`, default OFF via a 100K pulldown. n30nex never drives it. See #805 — resolve before relying on the radio.
- **`SELECT` (GPIO7)** on the RA62A_LF/RA62A_HF is undriven by n30nex and its function is unconfirmed.
[...]

PR #825, opened 2026-08-18
feat(#824): publish RC32 tester diagnostic builds on every release

### LINKED ISSUE #824: task: publish RC32 tester diagnostic builds on every release while RadioCore is unreleased hardware
## Decision

**Owner decision, 2026-08-18:** the three RC32 `_testdiag` envs are added to
`.github/release-envs.txt` and published as release assets on **every** release, for as long as
RadioCore remains unreleased hardware.

    heltec_rc32_companion_radio_ble_testdiag
    heltec_rc32_companion_radio_usb_testdiag
    heltec_rc32_repeater_testdiag

Release matrix goes 91 -> 94 envs. All 94 verified to resolve to real `[env:]` sections.

## Why publish diagnostics at all

A beta tester whose board will not boot **cannot reach a CLI**. Hand-built binaries were the
alternative and they go stale the moment `firmware-base` moves -- the set handed out today was
already pinned to `offband-v1.5.0-beta3-6-g925b8126` and would have needed rebuilding by hand on
every subsequent change. Publishing per-release keeps them current for free.

## What makes them safe to hand out

Each **EXTENDS the shipped env** and adds ONLY:

- `OFFBAND_BOOT_BEACON` -- raw UART0 writes from the earliest static ctor, before any driver,
  RTOS or Arduino init. Presence of `APP:CTOR` after a reset separates "fault in our code" from
  "fault at or before the bootloader's jump".
- `OFFBAND_MESHLOG_UART0` -- mirrors the live log to UART0, which survives the chip failing and
  does not power-cycle the board on attach. Since #769 this emits regardless of whether caplog
  capture is enabled, so a tester needs no CLI setup.

**Nothing else differs.** That is the whole point. The older `_diag` envs carried local
`ENV_INCLUDE_GPS=0` and `DISPLAY_ROTATION=0` overrides, and that is precisely why the
GPS-on-a-strapping-pin defect (#704) survived every bench session -- every result came from a
build that had already worked around it. A diagnostic build that diverges diagnoses the wrong
firmware.

Deliberately excluded: `OFFBAND_POWER_TELEMETRY`. Noise when the question is "why will it not
boot", and testers have no battery instrumentation rig.

## Scope and REVISIT triggers

Recorded inline in `release-envs.txt`, not only here:

1. **When RadioCore ships as released hardware** -- reconsider whether diagnostics belong on
   stable releases.
2. **When #763 lands its level gate** -- the UART0 mirror is a synchronous bounded busy-wait on
   the TX FIFO. Combined with debug-level logging it hangs `setup()` (2 s -> 118 s -> never
   completing). These envs do not enable debug logging so they are safe as built, but that
   hazard is exactly why this is scoped to unreleased hardware rather than adopted as a general
   pattern.

## Also delivered

`tools/diag/rc32-tester/` -- README (wiring, flash addresses, how to read boot beacons and ROM
reset causes), the sniffer sketch, and `capture.py`. Packaged and attached to the beta3 release
as `rc32-tester-diag-kit.zip` for the tester currently reporting a non-booting RC32.

## Not in scope

**RCC6 is deliberately untouched.** The RCC6 session owns its own release-matrix entries as part
of landing the variant (#624) -- adding them before `variants/heltec_rcc6` exists would turn the
matrix red.

Related: #771 (rc32 added to the release matrix), #773 (beta epic), #763 (level gate), #704.

Strycher:
Closes #824.

Owner decision 2026-08-18: while RadioCore remains **unreleased hardware**, the three RC32
`_testdiag` envs ship as release assets alongside the shipped builds. Release matrix **91 → 94**;
all 94 verified to resolve to real `[env:]` sections, and the new envs verified building from
this commit.

## Why publish diagnostics at all

A beta tester whose board will not boot **cannot reach a CLI**. The alternative was hand-built
binaries, and those go stale the moment `firmware-base` moves — the set handed out today was
already pinned to a specific SHA and would have needed rebuilding by hand on every subsequent
change.

## What makes them safe to hand out

Each **extends the shipped env** and adds only `OFFBAND_BOOT_BEACON` and
`OFFBAND_MESHLOG_UART0`. **Nothing else differs**, and that is the entire point: the older
`_diag` envs carried local `ENV_INCLUDE_GPS=0` and `DISPLAY_ROTATION=0` overrides, which is
exactly why the GPS-on-a-strapping-pin defect (#704) survived every bench session — every result
came from a build that had already worked around it. A diagnostic that diverges diagnoses the
wrong firmware.

`OFFBAND_POWER_TELEMETRY` is deliberately excluded: noise when the question is "why will it not
boot", and testers have no battery rig.

Since #769 the UART0 mirror emits **regardless of whether caplog capture is enabled**, so a
tester needs no CLI setup. Before that change this kit would have shipped a silent wire.

## Two REVISIT triggers — recorded inline in `release-envs.txt`, not only here

1. **When RadioCore ships as released hardware** — reconsider whether diagnostics belong on
   stable releases.
2. **When #763 lands its level gate** — the UART0 mirror is a synchronous bounded busy-wait;
   combined with debug-level logging it hangs `setup()` (2 s → 118 s → never completing). These
   envs do not enable debug logging so they are safe as built, but that hazard is why this is
   scoped rather than adopted as a general pattern.

## Also lands

`tools/diag/rc32-tester/` — README covering wiring, flash addresses (merged at `0x0`, app-only
at `0x10000`), and how to read boot beacons and ROM reset causes, plus the sniffer sketch and
capture script. The packaged zip is **already attached to the beta3 release**.

**Binaries are gitignored** (`dist/`, `tools/diag/rc32-tester/firmware/`) — git keeps every
version of a binary forever and cannot delta-compress them, so each rebuild would add another
full copy to every clone. Release assets are the right home and cost clones nothing.

## Not in scope

**RCC6 is deliberately untouched.** That session owns its own matrix entries as part of landing
the variant (#624); adding them before `variants/heltec_rcc6` exists would turn the matrix red.

Related: #771, #773, #763, #704, #766.


Strycher:
Superseded by #834.

This PR went `CONFLICTING` when the RCC6 role matrix (`d0718420`, #804/#806) landed in `.github/release-envs.txt` underneath it. Checks were green (48/48) — the block was purely the merge conflict.

#834 carries the identical commit content rebased onto current `firmware-base`, on a fresh branch rather than a force-push of this published one.

**The conflict was semantic, not just textual, and worth recording:** both changes are owner-approved exceptions to the same "diag envs are not released" rule — RC32 `_testdiag` (#824, while RadioCore is unreleased hardware) and RCC6 `_diag` (#806, while RCC6 is beta) — approved days apart, neither aware of the other. The incoming RCC6 comment asserted "That holds for the RC32 and for every other board here", which was accurate when written and false once #824 landed.

Resolved in #834 by keeping both env blocks and rewriting the shared rule comment to name both exceptions, note each is scoped to beta hardware, and state that a third board needs a fresh owner decision rather than inheriting. The RCC6 section is otherwise untouched. 102 envs, all verified resolving, both `_testdiag` envs rebuilt from the rebased commit.

Strycher:
Owner disposition (2026-08-19): BlueElk work per owner (with closed #824); Agent field stamped on both. Agent: WhiteFalcon (session cb12cdd2)

Issue #822, opened 2026-08-18
task: Phase 1 -- one shared splash component, Offband branding on every role


Strycher:
Parent: #705. Phase 1 of two, per the agreed plan.

Collapses **six** independent splash implementations into one shared component and puts Offband branding on every board with a display. **Mono only — no new colour assets, no driver work, no hardware gating.** Phase 2 (colour per driver) is separate and hardware-gated.

Five of six implementations currently ship MeshCore branding unconditionally: `ui-orig`, `ui-tiny`, `simple_repeater`, `simple_room_server`, `simple_sensor`. Only `ui-new` has Offband art.

## Scope

- `src/helpers/ui/OffbandSplash.{h,cpp}` — the one component; all six UITask files call it and delete their own splash code.
- Move the generated 1-bit art out of `examples/companion_radio/ui-new/` to somewhere every role can include, and update `gen-offband-logo.py`'s output path.
- `DisplayDriver`: physical panel size accessors (third request — #749 and #758 both deferred it) and `supportsColourArt()`, defaulted false. Colour drivers declare capability; **variants no longer opt in via a per-board flag**, which is the mechanism I have missed boards with twice.
- Guard: fail when a driver reports colour capability but has no asset configured.
- MeshCore attribution stays (#153, MIT obligation) as text, not as MeshCore artwork.

## Cost

1-bit Offband mark + wordmark is **285 B**, cheap enough for every board including the ~115 mono OLEDs.

## Acceptance

- Exactly ONE splash implementation remains. If the tree ends with a seventh, this failed.
- No MeshCore *artwork* on any role; attribution preserved as text.
- Mono / e-ink / no-display behave as before apart from the branding.
- RC32 colour splash unchanged.
- Every affected env builds.

Agent: BrightCat (session 5970917d)

Issue #804, opened 2026-08-18
task: correct four claims in the RCC6 bring-up handoff (prior art, ADC pins, LoRa map, board template)


Strycher:
Three claims in `docs/radiocore/rcc6-bringup-handoff.md` (merged `fd8df8cb`, #802) are contradicted by evidence produced 2026-08-17. The doc is otherwise sound and its confidence-marker discipline should be preserved — this corrects the content, not the method.

### 1. "No reference variant — we would be first"

**Doc says** (§1): *"Upstream MeshCore has no RCC6 either — we would be first. There is no reference variant to copy from."*

**Actual:** `n30nex/NeonPocketMC-RCC6` (MIT, not a fork, pushed 2026-08-13) ships a complete `variants/heltec_rcc6/` — `heltec_rcc6.cpp`, `heltec_rcc6.h`, `pins_arduino.h`, `platformio.ini`, `target.cpp`, `target.h` — **plus `boards/heltec_rcc6.json`**. It extends `esp32c6_base`, the same base name this repo defines at `platformio.ini:129`.

True only of *upstream* MeshCore. `docs/radiocore/README.md:200,323-324` and #624 both already recorded the n30nex prior art; the handoff did not incorporate them.

### 2. "ADC_IN / ADC_Ctrl / Battery_ADC assignments were never confirmed"

**Doc says** (§3): *"`ADC_IN`/`ADC_Ctrl`/`Battery_ADC` nets exist on the RCC6 schematic, but their GPIO assignments were never confirmed."*

**Actual**, read from the rendered ESP32-C6 symbol (U8) bottom edge:

| Net | Package pin | SoC function | GPIO |
|---|---|---|---|
| `ADC_Ctrl` | 11 | MTDI | **5** |
| `ADC_IN` | 12 | MTCK | **6** |
| `USER_Key` | 15 | GPIO9 | **9** |

All three match n30nex (`PIN_ADC_CTRL=5`, `PIN_VBAT_READ=6`, `PIN_USER_BTN=9`).

`GPIO12`/`GPIO13` carry `D_N`/`D_P` — native USB, which supplies the mechanism for the unconfirmed COM45 enumeration in §7.

### 3. The LoRa pin map is not a hypothesis

**Doc says** (§3): the row-paired map is `[hypothesis:]`, must not be trusted, and must be re-derived.

[...]

PR #832, opened 2026-08-19
fix(#779): zero the reply buffer before dispatch — stop transmitting stack bytes

### LINKED ISSUE #779: task: zero the reply buffer at the four handleCommand call sites (fix for #765)

Strycher:
Closes #779. Fixes the code half of #765. Under epic #774.

**#778 fixed the three keys that triggered the bug. This fixes the mechanism**, so a handler that writes nothing can never again put stack bytes on the air.

## The defect

```c
uint8_t temp[166];                 // uninitialised stack
char *reply = (char *)&temp[5];
if (is_retry) { *reply = 0; }      // zeroed ONLY on the retry path
else { handleCommand(sender_timestamp, command, reply); }
int text_len = strlen(reply);      // over uninitialised memory
if (text_len > 0) { ...transmit... }
```

A handler that returns without writing causes `strlen()` to read uninitialised stack, and the result is transmitted to the requesting peer. That is the mojibake a user saw in the client app after upgrading from stock MeshCore 1.15.0.

**It was invisible on serial** — those callers zero their buffer and suppress empty replies. Only the radio transport showed it: the transport hardest to observe, and the one real users are on.

## Four call sites, one invariant

The reply buffer is empty **before** dispatch, not per-branch.

| File | Transport | Was |
|---|---|---|
| `simple_repeater/MyMesh.cpp` | over-the-air | zeroed only on `is_retry` |
| `simple_room_server/MyMesh.cpp` | over-the-air | zeroed on every branch **except** the admin dispatch — the one path that actually reached `handleCommand` |
| `simple_sensor/SensorMesh.cpp` | over-the-air | never zeroed on any path |
| `simple_sensor/main.cpp` | **serial** | missing `reply[0] = 0` |

`simple_sensor` was the only role missing it on **both** transports, so it had no safe transport to fall back on.

## Verification

**All 15 `handleCommand` call sites audited**, not only the four changed. Twelve own a reply buffer and now zero it before dispatch. Three own no buffer and are correctly untouched:

- `simple_repeater/main.cpp:762` — zeroes with `reply[0] = '\0'`
- `simple_secure_chat/main.cpp:546` — single-argument signature, no reply parameter
- `simple_sensor/SensorMesh.cpp:450` — forwards a buffer its caller owns

**The `temp[4]` adjacency.** In `simple_room_server` the flags byte `temp[4]` sits directly below the reply buffer at `temp[5]`. It is written *after* `handleCommand` returns, so hoisting the zeroing cannot disturb it. Called out because it is the one place where moving an assignment earlier could plausibly have broken something.

**Builds 5/5:** `Heltec_ct62_sensor`, `Heltec_WSL3_sensor`, `heltec_v4_repeater_telemetry`, `heltec_v4_room_server`, `RAK_4631_repeater`.

**Adversarial review.** Gemini was given SAFELANE, CLAUDE-BASE, the diff and the call-site context with no question list. Verdict: recommend merge, no findings. It independently confirmed the `temp[4]`/`temp[5]` ordering and judged the caller-side fix the right layer — a handful of transport call sites versus every present and future command handler.

It noted it could not audit the whole tree for other call sites. That audit is the 15-site check above.

## Not closed here

**#765 stays open.** This is its code fix; the bug closes on the owner's call after #777, the epic integration test that exercises the surface over **both** transports.

## Note for reviewers touching `examples/`

BrightCat has claimed #822 (splash consolidation), which touches every role's `UITask`. The only overlap is `simple_sensor/main.cpp`, and this change is one line inside the serial command handler — nowhere near UI setup. Flagged so a textual conflict is expected rather than surprising.

---

### DIFF
--- a/examples/simple_repeater/MyMesh.cpp
+++ b/examples/simple_repeater/MyMesh.cpp
@@ -736,9 +736,12 @@ void MyMesh::onPeerDataRecv(mesh::Packet *packet, uint8_t type, int sender_idx, 
       uint8_t temp[166];
       char *command = (char *)&data[5];
       char *reply = (char *)&temp[5];
-      if (is_retry) {
-        *reply = 0;
-      } else {
+      // #765: zero BEFORE dispatch, not only on the retry path. `temp` is
+      // uninitialised stack, so a handler that writes nothing would otherwise
+      // leave the strlen() below reading -- and transmitting -- whatever was
+      // already there.
+      *reply = 0;
+      if (!is_retry) {
         handleCommand(sender_timestamp, command, reply);
       }
       int text_len = strlen(reply);
--- a/examples/simple_room_server/MyMesh.cpp
+++ b/examples/simple_room_server/MyMesh.cpp
@@ -463,12 +463,14 @@ void MyMesh::onPeerDataRecv(mesh::Packet *packet, uint8_t type, int sender_idx, 
                           PUB_KEY_SIZE);
 
       uint8_t temp[166];
+      // #765: `temp` is uninitialised stack. Every branch below except the admin
+      // dispatch already zeroed it, so the one path that calls handleCommand was
+      // the one path that could transmit stack bytes. Zero it up front instead.
+      temp[5] = 0; // no reply
       bool send_ack;
       if (flags == TXT_TYPE_CLI_DATA) {
         if (client->isAdmin()) {
-          if (is_retry) {
-            temp[5] = 0; // no reply
-          } else {
+          if (!is_retry) {
             handleCommand(sender_timestamp, (char *)&data[5], (char *)&temp[5]);
             temp[4] = (TXT_TYPE_CLI_DATA << 2); // attempt and flags,  (NOTE: legacy was: TXT_TYPE_PLAIN)
           }
--- a/examples/simple_sensor/SensorMesh.cpp
+++ b/examples/simple_sensor/SensorMesh.cpp
@@ -586,6 +586,7 @@ void SensorMesh::onPeerDataRecv(mesh::Packet* packet, uint8_t type, int sender_i
         uint8_t temp[166];
         char *command = (char *) &data[5];
         char *reply = (char *) &temp[5];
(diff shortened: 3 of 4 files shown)


Strycher:

Re-validated against the remote immediately prior: **1 commit**, **4 files** (`simple_repeater/MyMesh.cpp`, `simple_room_server/MyMesh.cpp`, `simple_sensor/SensorMesh.cpp`, `simple_sensor/main.cpp`), `mergeable=MERGEABLE`, checks **48/48 SUCCESS**, base `firmware-base`. Unchanged from what was approved.

`mergeStateStatus=BEHIND` — base gained four RCC6 commits (#804/#830) while CI ran. None touch these four files, so the update is mechanical, not a conflict. Updating the branch server-side, then auto-merge with rebase.

---

Issue #840, opened 2026-08-19
task: push + PR -- ST7735 palette (#791) and boot-rotation fix (#815)


Strycher:
Push/PR task for the two completed build tasks on branch `fix/815-st7735-rotation`, branched off `firmware-base` @ `16b0056c`.

Exists so the pre-push open-issue gate is satisfied by **this** task rather than by closing #791 or #815, which are not finished until their PR is merged.

## Carries

- `28eec66d` feat(#791) -- Offband dark palette for ST7735, on a shared brand header
- `684424da` fix(#815) -- restore ST7735 boot rotation, unshadow the driver's rotation call

## Evidence

- **Guard proven both directions.** `scripts/check_display_virtuals.py` exits **1** against unfixed `firmware-base`, reporting `COLLISION ST7735Display.cpp:343  static 'setRotation' shares its name with a DisplayDriver virtual`; exits **0** on the fix. Reproducible in CI, not a terminal claim.
- **27/27 ST7735 envs build.** 26 green in one sweep; `Heltec_t1_room_server` failed once with an error that was lost to a `tail` pipe, then clean-rebuilt green twice, and the whole `t1` family clean-builds green. Verified across two runs, not one table -- stated plainly rather than reported as a single clean sweep.
- **#815 hardware-verified by the owner** on `heltec-t096-1`.
- Palette extraction proven a no-op by stripped-object comparison (sha `057f93b0`).

## Note

CI covers only **2 of the 27** ST7735 envs (`Heltec_t096_companion_radio_ble`, `Heltec_t096_repeater`), and not the one that failed. `ci-green` will not arbitrate this matrix -- the local sweep above is the gate.

Parent: #785. #791, #815, #812, #813, #814 all stay open pending the owner's call after merge.

Agent: BrightCat (session 5970917d)

Strycher:
Closing the push/PR task so the pre-push gate is satisfied by this task rather than by #791/#815, which remain open pending the owner's call after merge.

Issue #815, opened 2026-08-18
task: build + test -- restore ST7735 boot rotation


Strycher:
Parent bug #812. Depends on the plan task.

- Implement exactly the agreed binding.
- A test that would have caught this: assert the driver's rotation path is actually invoked, not that a value is merely defined.
- Build all 15 ST7735 envs across t096, t1, tracker, tracker_v2 — this bug spans four boards with two different `DISPLAY_ROTATION` values (1, and 3 from `variant.h`).
- **Hardware verification on `heltec-t096-1`**, the only affected board available. t1, tracker and tracker_v2 cannot be verified — state that rather than implying coverage.
- Confirm #148's `_sys` rotate CLI still behaves per the plan's decision.

**Exit:** owner sign-off on the panel.

Agent: BrightCat (session 5970917d)

Strycher:
Closed on owner authorization (2026-08-19). Rotation fix merged in `c3bb1b2e`, hardware-verified on `heltec-t096-1` on both the companion and repeater roles. `check_display_virtuals.py` guards the shadowing in CI.

Issue #814, opened 2026-08-18
task: plan -- how to bind the rotation call correctly


Strycher:
Parent bug #812. Depends on the reconcile task.

Design decisions, all resolved before code:
- **How to bind.** Qualify the call, rename the file-static function, or make `ST7735Display` a real `override` of `DisplayDriver::setRotation`. These are not equivalent: the virtual's contract is *degrees* (0/180, per #148) while the file-static takes a *rotation index* (0-3). Conflating them is how this bug happened.
- **Whether `supportsRotation()` should now return true** for ST7735, which would expose the `_sys` rotate CLI on these boards — a behaviour change, not just a fix.
- **Whether other call sites are exposed to the same shadowing.** ST7735Display has 2 call sites for a 1-definition static.
- **A guard so this cannot recur silently** — the failure mode is a clean compile with no warning. Consider `-Woverloaded-virtual` / `-Wshadow` scoped appropriately, or a test that asserts MADCTL is written.

**Exit:** owner AGREE before any code.

Agent: BrightCat (session 5970917d)

Strycher:
## #814 PLAN — how to bind the rotation call correctly

Every design decision resolved here so #815 has none left. **Ends at owner AGREE; no code until then.**

### Decision 1 — binding: make it a private MEMBER with a distinct name

Three options were considered:

| Option | Fixes binding | Removes the trap | Semantics |
|---|---|---|---|
| Qualify the call — `::setRotation(...)` | yes | **no** — collision remains for the next reader | index preserved |
| Rename the file-static — `applyPanelRotation()` | yes | yes | index preserved |
| **Make it a private member `ST7735Display::applyPanelRotation(uint8_t index)`** | yes | yes | index preserved |

**Proposed: the third.** It is the only one that converts an *invisible* hazard into a *compiler-visible* one. As a free function it can be silently shadowed again by any future base virtual, with no diagnostic. As a member with a distinct name, a future base virtual of the same name becomes a member-hides-virtual situation, which `-Woverloaded-virtual` can actually diagnose (see Decision 4).

Also: `_colstart` / `_rowstart` / `_xstart` / `_ystart` / `rotation` / `_width` / `_height` are file-scope statics shared by one driver instance. This task will **not** restructure them — that is unrelated cleanup and belongs in its own ticket if wanted.

### Decision 2 — do NOT conflate index with degrees

This is what caused the bug and it must not be re-created:

[...]

Issue #813, opened 2026-08-18
task: reconcile -- why did this render correctly during 1.17.0 testing?


Strycher:
Parent bug #812. **Do this before designing the fix.**

The shadowing mechanism in #812 is verified from the compiled object. What is NOT established is that it is *sufficient* — the owner saw this panel render correctly days ago, and by the mechanism it should already have been broken.

Eliminate, with evidence, not reasoning:
- Which env was previously flashed. The stored advert name ends `-R`, matching the repeater convention (cf. `Offband-RC32-R`); the failing observation is `companion_radio_ble`.
- Whether `Rcmd1` writes MADCTL, leaving the panel in a usable orientation regardless of the discarded `setRotation`.
- The resolved `TFT_eSPI` version. `^2.4.31` is a caret dependency and currently resolves to 2.5.43; a different resolve changes `TFT_eSprite` internals.
- Whether the sprite allocation succeeds under each role's memory budget (repeater 48,648 B static vs companion_radio_ble 164,624 B of 235,520; the sprite wants 25,600 B and `createSprite` failure is logged but not guarded before `endFrame()` blits the pointer).

**Exit:** a model that explains BOTH the failing observation and the working one. If the shadowing alone cannot, say so — the fix design depends on this.

Agent: BrightCat (session 5970917d)

Strycher:
## #813 RECONCILE — findings

**Result: the shadowing mechanism is now proven SUFFICIENT, and all three alternative candidates are eliminated. But it is also proven ENV-INDEPENDENT, which sharpens rather than resolves the "it worked days ago" question.**

### The mechanism is sufficient `[verified]`

The missing piece was what MADCTL actually ends up as:

| Step | MADCTL | Meaning |
|---|---|---|
| `Rcmd1` entry 14 writes `0xC8` | `MY \| MX \| BGR` | **no `MV` — PORTRAIT** |
| `setRotation(1)` *would* write `0xA8` | `MY \| MV \| BGR` | **`MV` set — LANDSCAPE** |
| Shadowed call does nothing | stays `0xC8` | **panel left portrait** |

So the panel is initialised portrait and the correction never runs. Combined with:

- `_xstart`/`_ystart` are `= 0` at `:276-277` and assigned **only inside the dead function**, so `_colstart = 24` never reaches the address window
- `endFrame()` writes `setAddrWindow(0, y, 160, 1)` — 160-wide rows — into an 80-wide portrait window

[...]

PR #841, opened 2026-08-19
fix: ST7735 boot rotation (#812) + Offband dark palette (#785 Part A)

Strycher:
Closes the push/PR task #840. **#791, #812, #813, #814, #815 and #785 stay open** — deliberately no closing keywords; those are the owner's call after merge.

## Two commits, separable

| | |
|---|---|
| `42d4e336` | Offband dark palette for ST7735, on a shared brand header (build task 791) |
| `373a8ed6` | restore ST7735 boot rotation — unshadow the driver's rotation call (build task 815, bug 812) |

## The bug

`ST7735Display.cpp` defined `static void setRotation(uint8_t)` and called it **from a member function**. When `556518a4` added `virtual void setRotation(uint8_t)` to `DisplayDriver`, C++ name lookup — class scope, including base classes, **before** file scope — silently rebound that call to the new empty base method.

Established from the compiled object, not from reading source: `nm` showed the only `setRotation` symbol was `W DisplayDriver::setRotation` with body `bx lr`, and the file-static had been eliminated as unreferenced.

Consequences, all matching the reported photo: `Rcmd1` writes MADCTL `0xC8` (portrait) and the correction to `0xA8` (landscape) never ran; `_xstart/_ystart` stayed 0 so `_colstart=24` never reached the address window; and `endFrame()` kept writing 160-wide rows into an 80-wide window.

Boot rotation has been discarded since 2026-06-15 — every release from `offband-v1.0.0` onward.

## Why a rename, not a qualified call

`ST7735Display::applyPanelRotation(uint8_t index)`, a private member. Qualifying the call fixes the binding but leaves the trap armed for the next reader. As a member with a distinct name, a future collision becomes *member-hides-virtual*, which a compiler can actually diagnose. The header records why it must never be named `setRotation` again, and that the base contract is **degrees** (0/180) while this takes an **index** (0–3) — conflating those is what caused the bug.

## The guard, and why it is in this PR

`scripts/check_display_virtuals.py`. No compiler flag catches this: `-Woverloaded-virtual` fires when a *member* hides a base virtual, whereas here a free function was hidden *by* the inherited member — ordinary, legal lookup.

**Proven in both directions:**

```
unfixed firmware-base:  COLLISION ST7735Display.cpp:343
                        static 'setRotation' shares its name with a DisplayDriver virtual
                        1 collision(s).  exit 1
this branch:            OK: no file-static function collides with a DisplayDriver virtual
                        exit 0
```

An earlier PR was closed for bundling a guard with the fix it catches, so the guard never ran red. The red/green pair above is reproducible — check out `firmware-base`, copy the script in, run it.

## Verification

- **27/27 ST7735 envs build.** Honest detail: 26 green in one sweep; `Heltec_t1_room_server` failed once with an error **lost to a `tail` pipe**, then clean-rebuilt green twice along with its whole family. Verified across two runs, not one table.
- **Hardware-verified by the owner** on `heltec-t096-1` — the display renders correctly after the fix.
- 211/211 native tests. Palette extraction proven a no-op by stripped-object comparison (sha `057f93b0`); the RC32 panel is untouched by this change.
- Object-level proof: `T ST7735Display::applyPanelRotation` is defined in the driver object and the `0xa8` landscape MADCTL immediate is present.

## ⚠ CI will not fully gate this

`ci.yml` covers **2 of the 27** ST7735 envs (`Heltec_t096_companion_radio_ble`, `Heltec_t096_repeater`) — and **not** `Heltec_t1_room_server`, the one that failed. `ci-green` passing is necessary but not sufficient here; the local 27-env sweep above is the real evidence.

## Scope correction

Blast radius stated as 15 envs on #812, then 24 by me. **It is 27** — enumerated from the four variants rather than eyeballed. Correction posted to #812.


Strycher:
Human approval received from Ben in session `5970917d`.

- `state=OPEN`, `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` (blocked on `ci-green` only, not on conflict)
- **3 commits, 8 files** — the two work commits plus the merge from `firmware-base` that `gh pr update-branch` added to clear `BEHIND`. Files unchanged from the two-commit state at 8.
- Checks: 47 total, 1 SUCCESS, remainder QUEUED/IN_PROGRESS. **None failing.**

Auto-merge enabled with `--rebase`; the merge will occur when `ci-green` passes. Not merged manually — the human makes the decision, automation performs the merge.

⚠ Reviewers: `ci.yml` exercises **2 of the 27** ST7735 envs affected by this change and does not include `Heltec_t1_room_server`. `ci-green` here is necessary but not sufficient; the local 27-env sweep in the PR body is the substantive evidence.

**#791, #812, #813, #814, #815 and #785 remain open** and are not closed by this merge — no closing keywords were used, deliberately. Those are the owner's call.

Strycher:
**Correction.** This thread said CI guards the file-static shadowing trap with `scripts/check_display_virtuals.py`. That was false: the script existed, but no workflow ran it. It also scanned only `src/helpers/ui/`, so it could not see variant-local drivers like `variants/heltec_rc52/RC52Display.cpp`.

[#1123](https://github.com/OffbandMesh/meshcore-firmware/issues/1123) fixes both on `epic/864-color-splash` (commit `0bd49eb2`, not yet merged). The guard now scans `variants/*/` too, it has a self-test, and both run in `config-lint` as advisory steps. CI will not run the guard until that epic merges.

Issue #842, opened 2026-08-19
task: push + PR -- shared splash component, Phase 1 (#822)


Strycher:
Push/PR task for #822 (Phase 1 of #705), branch `task/822-shared-splash` off `firmware-base`.

Exists so the pre-push open-issue gate is satisfied by **this** task rather than by closing #822 or #705, which are not finished until merged and hardware-verified.

## What lands

Six independent splash implementations collapse into one: `src/helpers/ui/OffbandSplash.{h,cpp}`. Five of the six drew **MeshCore** artwork, and the colour path was gated on a per-variant `-D OFFBAND_COLOUR_SPLASH`, so RC32 got Offband branding and every other role and panel silently kept MeshCore's.

Call sites migrated: `ui-new`, `ui-orig`, `ui-tiny`, `simple_repeater`, `simple_room_server`, `simple_sensor`. All six `meshcore_logo` arrays deleted.

## Design points

- **Capability IS the asset.** `DisplayDriver::colourSplashArt()` returns the art or `nullptr`; a driver cannot claim colour capability and have nothing to draw, and no per-variant flag can be forgotten. This replaces the flag that missed boards twice.
- **Layout derives from the panel.** The 72×40 techo card cannot take the 119px lockup or three text lines, so it gets a compact mark+version layout. Branding degrades on small panels; it never disappears.
- The #222 version derivation moved out of `ui-new` (reachable only by the companion role) into `offband::shortVersionInto`, split as a pure function — its fixtures previously only exercised a Python re-implementation and now run the shipping C++.
- The `v1.2.3-abcdef` commit-hash trim, hand-rolled at four call sites, happens once.

## Evidence

- **275 native assertions** green, incl. 24 new splash/version tests.
- `check_display_virtuals.py` (landed via #841) green, and now checking **27** DisplayDriver virtuals — confirming the three new ones are seen.
- `nm` on linked images: RC32 carries `offband_splash_rgb565` (0x41a0); `Heltec_v3_repeater` carries `offband_mark` + `offband_word` and no colour asset.

## NOT verified

**No hardware.** `ui-orig` and `ui-tiny` cannot be verified on this bench at all — every `ui-orig` board here is `NullDisplayDriver` and there is no T-Echo Card. The compact layout is untested by construction.

Parent: #705. #822 and #705 stay open pending the owner's call after merge and hardware verification.

Agent: BrightCat (session 5970917d)

Strycher:
[...]

Issue #743, opened 2026-08-16
epic: LCD rendering -- partial updates instead of full-screen repaints, and a splash refresh


Strycher:
The RC32 LCD repaints the **whole screen on every update**, which flickers visibly, and the
splash screen looks dated next to what MeshCore ships. Both are visual-identity problems and both
are owner-raised (2026-08-16).

## 1. Full-screen repaints instead of partial updates

Every UI update currently redraws the entire panel rather than the region that changed. On the
RC32's NV3001B this is visible as flicker on each repaint — already recorded as a known cosmetic
issue against `rc32-bench-1` when the repeater display was first confirmed working (2026-08-14).

Beyond looking bad, full-frame redraws cost SPI bandwidth and CPU on every tick, on a board where
the display shares silicon with the radio and the mesh loop.

Wanted: dirty-region / partial-update rendering — redraw what changed, not the frame. Scope
should establish whether this is an `NV3001BDisplay` concern, a `UITask` concern, or a shared
display-abstraction concern, because the answer decides how many boards benefit.

## 2. Splash screen refresh

Owner: *"it could look much sexier; compared to what MeshCore is doing, we look outdated and
old."*

This is the **aesthetic** half. The **structural** half is already tracked and must not be
duplicated:

- **#705** (open, P1) — *five duplicate splash implementations, four shipping MeshCore branding;
  needs one shared branding component.* That is the consolidation + correct-branding work.
- **#153** (closed) — made Offband prominent on boot, OLED-first, keeping MeshCore attribution.
  Prior art for the constraints.

Sensible sequencing: land #705's single shared component first, then do the visual redesign
**once**, in that one place, rather than five times. This epic should own the redesign and depend
on #705 rather than re-solve it.

Constraints worth stating up front so the redesign does not have to be redone:
- MeshCore attribution is retained (#153) — this is an MIT fork and the attribution stays.
- Panels differ wildly across the fleet: NV3001B colour LCD (RC32), SSD1306 OLED (V3/V4),
  e-ink (E290), and no-display variants. A design that only works on colour is not a fleet design.
- Any user-visible wording is owner-approved before it ships.

## Related

- **#623** — epic: RadioCore RC32 firmware support. The repaint work is first visible there, but
  the display abstraction is not RC32-only.
- **#705** — splash consolidation (blocking the redesign).

⚠ Parenting: filed top-level. If there is a display/UX Feature this should hang under, say so and
I will re-parent rather than guess.

[...]

Issue #705, opened 2026-08-15
bug: five duplicate splash implementations — four ship MeshCore branding; needs one shared branding component


Strycher:
Offband ships **five separate splash implementations**. Four of them render upstream MeshCore's logo, MeshCore's version string, and `https://meshcore.io`.

| UI | branded |
|---|---|
| `companion_radio/ui-new/UITask.cpp` | yes |
| `companion_radio/ui-orig/UITask.cpp` | **no** |
| `companion_radio/ui-tiny/UITask.cpp` | **no** |
| `simple_repeater/UITask.cpp` | **no** |
| `simple_room_server/UITask.cpp` | **no** |
| `simple_sensor/UITask.cpp` | **no** |

Anyone who downloads Offband and flashes a repeater, room server, sensor, or an older companion UI gets a device that boots as MeshCore and points at meshcore.io. That is wrong on its face: they came to Offband for Offband.

## The actual defect is the duplication, not the missing branding

There is **one** product, **one** logo, **one** version. There must not be five splash implementations to keep in sync — branding four more files individually recreates the same problem four times over and guarantees the next upstream merge or new role drifts again.

**Required: a single shared splash/branding component that every role UI calls.** One logo asset, one version-string source, one layout. Roles supply only what genuinely differs (e.g. the node-type line). Adding a new role must not mean writing a new splash.

## Scope — explicitly NOT to be reduced

All six UIs above, and any other user-visible surface carrying upstream identity. Do **not** fix only the role that surfaced this, and do **not** leave any role "upstream-equivalent because we didn't otherwise touch it". If a surface is user-visible on an Offband device, it is in scope.

## Notes

- Splash geometry is currently written for a 128x64 OLED; on the RC32's 128x220 panel the lockup occupies the top ~27%. The shared component must not hardcode a single panel geometry.
- `ui-new` already renders "on MeshCore <version>" deliberately, per VERSIONING.md Pattern B (both identities). That convention should carry into the shared component.
- Surfaced during RC32 bring-up (#623): the repeater boots showing MeshCore branding on new hardware.
- **Exact user-facing wording and layout require the owner's approval before shipping** -- that is a copy sign-off, not a scope reduction.

---
Agent: DustyBarn (session 7bd68768)

Strycher:
## PLAN — one shared splash, Offband branding everywhere

[...]

Issue #222, opened 2026-06-27
Build identity must be visible on OLED + app (settable OFFBAND_BUILD_TAG for flag-only variants) (#216)


Strycher:
Hard testing requirement: you must be able to read exactly which firmware is on the device from the **OLED splash or the app** — not just a serial monitor. Today the app shows a stripped `<offband>-<meshcore>` semver (e.g. `1.1.0-1.16.0`) with no SHA/build-counter, and two builds at the **same commit** that differ only by a compile flag (our D1 A/B) are **indistinguishable by any git-derived string**.

Fix: support a settable `OFFBAND_BUILD_TAG` (`-D OFFBAND_BUILD_TAG='"..."'`) rendered on (a) the OLED splash version line and (b) the app device-info version field (within the 20-char limit). Lets any build — including flag-only variants — self-identify on-device.

Parent: #216. Prereq for the GPS-autobaud bench diagnostic (every test is invalid until build identity is unambiguous).

Strycher:
Landed via PR #234 (merge 9caf29db). Verified on-device + full CI matrix green.

PR #843, opened 2026-08-19
feat: one shared splash component — Offband branding on every role (#705 Phase 1)

Strycher:
Closes the push/PR task #842. **#822, #705 and #743 stay open** — no closing keywords, deliberately. Those are the owner's call, and #822 is **not hardware-verified**.

## The defect

Six independent splash implementations existed. **Five drew the MeshCore logo**, and the colour path was gated on a per-variant `-D OFFBAND_COLOUR_SPLASH`, so the RC32 got Offband branding and every other role and panel silently kept MeshCore's.

That is why a repeater on `firmware-base` still boots to a MeshCore splash today — reported on an RCC6, reproducible on any repeater build.

## What changes

All six call sites now call `offband::drawSplash()` in the new `src/helpers/ui/OffbandSplash.{h,cpp}`:

`ui-new`, `ui-orig`, `ui-tiny`, `simple_repeater`, `simple_room_server`, `simple_sensor`. All six `meshcore_logo` arrays deleted. **No MeshCore artwork remains in the tree**; the MIT attribution is preserved as text ("on MeshCore <ver>"), which was always the obligation — the artwork never was.

### Capability IS the asset

```cpp
virtual const ColourArt* colourSplashArt() const { return nullptr; }
```

A driver declares colour capability *by returning the art*. It cannot claim the capability and have nothing to draw, and no per-variant flag can be forgotten. This replaces `-D OFFBAND_COLOUR_SPLASH` as the decision mechanism — the flag is how boards were missed twice.

### Layout derives from the panel, not a macro

The 72×40 techo card cannot take the 119px lockup or three text lines. Rather than let it overflow silently — the same failure mode that put MeshCore artwork on five roles — the layout is chosen from the panel's own height, and small panels get a compact mark+version arrangement. **Branding degrades; it never disappears.**

### Duplication actually removed

- The **#222 version derivation** lived in `ui-new`, reachable only by the companion role. Moved to `offband::shortVersionInto` and split as a *pure* function. Its fixtures previously only exercised a **Python re-implementation** (`scripts/test_offband_version_splash.py`); they now run the shipping C++.
- The `v1.2.3-abcdef` commit-hash trim was hand-rolled at four call sites. Now done once.
- The repeater's "Turning OFF" screen keeps branding via `drawBrandLockup()` rather than being left blank where the MeshCore logo was.

## Evidence

- **275 native assertions** green, incl. 24 new splash/version tests.
- **9 envs SUCCESS** across all four UI flavours, both colour drivers, and a headless build: `heltec_rc32_companion_radio_ble`, `heltec_rc32_sensor`, `heltec_rc32_without_display_sensor`, `heltec_rcc6_repeater_display`, `Heltec_t096_repeater`, `Heltec_v3_repeater`, `Heltec_v3_room_server`, `LilyGo_T-Echo_Card_companion_radio_ble` (ui-tiny), `MeshTracker_X1_companion_radio_ble` (ui-orig).
- `check_display_virtuals.py` (landed via #841) green, now checking **27** DisplayDriver virtuals rather than 24 — independent confirmation the three new virtuals are visible to the guard.
- `nm` on linked images: RC32 carries `offband_splash_rgb565` (0x41a0); `Heltec_v3_repeater` carries `offband_mark` + `offband_word` and **no** colour asset.

## ⚠ Not verified — read before approving

- **No hardware. None.** This has never rendered on a physical panel.
- **`ui-orig` and `ui-tiny` cannot be verified on this bench at all.** Every `ui-orig` board here is `NullDisplayDriver`; there is no T-Echo Card. They build, and that is all that is known.
- **The compact layout is untested by construction** — it triggers below 128×64, which in the whole tree is only the 72×40 techo card.
- `ui-tiny` **loses its two-phase splash** (logo, then text). It is now one frame. Deliberate, and unverifiable here.

## Trap worth knowing

`SplashInfo` uses an explicit constructor, not a default member initializer. Firmware envs build as `gnu++11`, where an NSDMI makes the struct a non-aggregate and every brace-init call site fails. **The native test env uses a newer standard and did not catch it** — all tests passed while all seven firmware envs failed to compile. Anyone adding a struct to a header shared between `src/` and the native env inherits that trap.


Strycher:
## Hardware verification — PASSED on `heltec-t096-1` (owner, 2026-08-19)

Flashed `Heltec_t096_repeater` from this branch via nRF52 DFU (config preserved, identity confirmed intact afterwards). Owner-confirmed on the physical panel:

| | |
|---|---|
| **Splash** | ✅ Offband lockup + version + `on MeshCore <ver>` + `< Repeater >`. No MeshCore artwork. |
| **Rotation** | ✅ correct — #841's fix holds on the **repeater** role, not just companion |
| **Dark palette** | ✅ correct. No banner, which is right: the repeater role has no title bar. |
| **Power-off lockup** | rendered, but for **<250 ms** — see below |

That closes the "no hardware verification" caveat in the PR body for the **colour-panel + `simple_repeater`** path specifically. Still unverified and still true: `ui-orig` and `ui-tiny` cannot be verified on this bench at all, and the compact layout remains untested by construction.

## Second commit added: `a8bdb929` — #844/#845

The owner flagged the sub-250 ms power-off screen as an observation, not a complaint. It is a real defect and it is **not** from this PR.

`BUTTON_EVENT_CLICK` extends `_auto_off`; `BUTTON_EVENT_LONG_PRESS` did not. With the display already blanked (`AUTO_OFF_MILLIS` 20 s — the normal state), `_auto_off` is in the past, so the long-press turns the panel on, one frame renders, and the auto-off check blanks it **on the same loop pass**. The board then sits dark for the rest of `POWEROFF_DELAY` (3 s).

Pre-existing: `git diff origin/firmware-base..HEAD -- examples/simple_repeater/UITask.cpp` shows the splash commit changed nothing in `_auto_off` / `BUTTON_EVENT` / `turnOn` / `turnOff` / `_powering_off_at`. The splash change altered only what the screen *draws* — which is why the flash became conspicuous enough to notice at all.

Scope verified, not assumed: **`simple_repeater` only.** `simple_room_server` and `simple_sensor` have no long-press power-off path; `ui-new` (:805) and `ui-tiny` (:560) use a menu-driven `HomePage::SHUTDOWN` calling `_board->powerOff()` directly, with no delayed screen and no race.

Kept as a **distinct commit** at the owner's direction so it reviews separately while sharing this PR.

**#844 stays open** — the countdown itself is not yet verified on hardware. That needs: idle the repeater >20 s so the display blanks, then long-press, and confirm the screen stays lit for the full ~3 s.

Strycher:
Human approval received from Ben in session `5970917d`.

- `state=OPEN`, `mergeable=MERGEABLE`, `mergeStateStatus=BLOCKED` (on `ci-green` only, not on conflict)
- **2 commits, 24 files** — `1a2fcb6f` (splash) and `a8bdb929` (#844 auto-off fix), matching what was approved
- Checks: 47 total, 29 SUCCESS, **0 failing**, remainder in progress

Auto-merge enabled with `--rebase`, so both commits land as distinct commits on `firmware-base` rather than being squashed — the separation the owner asked for is preserved in history.

**Approved with a known gap, stated so it is on the record:** `a8bdb929` is **not hardware-verified**. The splash, rotation and palette were confirmed on `heltec-t096-1`; the power-off countdown was not. #844 remains open for exactly that.

**#822, #705, #743 and #844 all stay open.** No closing keywords anywhere in either commit or the PR body — those are the owner's call.

Issue #862, opened 2026-08-19
task: CI runs none of the pio-flash test suites — wire them in


Strycher:
Parent epic: #736. **Blocked by #861** — wiring CI before the red test is retired turns every PR in the repo red.

**Files: 1-2** — `.github/workflows/ci.yml`, plus `__main__` runners in two test files if that route is chosen.

## The gap

```
python tests config-lint runs:   check_cli_dispatch, check_env_capability_claims,
                                 test_cli_dispatch, test_env_capability_claims,
                                 test_gen_jbm_font, test_gen_offband_logo

pio-flash suites in the repo:    test_pio_flash_bootloader_discovery
                                 test_pio_flash_bridge_identity
                                 test_pio_flash_device_match
                                 test_pio_flash_mac

occurrences of "test_pio_flash" in ci.yml:   0
```

**None of the flash tool's tests gate a merge.** That is the tool enforcing one-approval-per-flash (#500), MAC verification before any write (#468), serial-first identity (#503) and anti-impersonation (#323). Its suite has never blocked anything.

The direct evidence that this matters: #861's test went red on 2026-07-31 and sat on the mainline for three weeks. Nothing could have caught it.

## What the fix actually requires

**1. Neither runner covers all four, and both fail silently.** See the correction comment below for the measured table. `pytest` collects nothing from `mac` and `bridge_identity`; direct execution no-ops `bootloader_discovery`. Both exit 0 having run nothing. The fix is to normalise the four onto one convention, not to pick a runner.

**2. Do not assume pytest is available.** `config-lint` pip-installs `platformio` and `pillow` only. Normalising onto direct execution avoids the question entirely.

**3. Linux is unverified.** CI is `ubuntu-latest` and this is a Windows-centric tool. Encouraging: `pio-flash.py` has no Windows-only module-level imports — `serial` is imported lazily inside `cmd_send`, the rest is stdlib — and the fixtures are plain strings (`"8&519AF3A"`), not live API calls. So it should work. **That is a prediction, not a result.** The first CI run settles it, and if a suite turns out to be genuinely host-dependent, say so and gate it rather than deleting the assertion.

## Advisory or blocking

Land it **advisory first** and read a few real runs before promoting. Precedent in this repo: the capability-claims check is advisory pending #648, and #775's guard landed advisory precisely so its behaviour could be observed before it gated anyone's work. Promotion is an owner decision (see #776 for the same question on a different check).

## Acceptance

[...]

Issue #861, opened 2026-08-19
chore: retire the stale pio-flash legacy-hash test — red on firmware-base since #503


Strycher:
Parent epic: #736.

**Files: 1** — `scripts/test_pio_flash_device_match.py`.

`test_legacy_hash_still_matches_when_no_serial_recorded` has been **failing on `firmware-base` since 2026-07-31**. Verified by bisect — `1 failed, 9 passed` at `c636d2dd`, `3bb7bd0b`, `da56a157` and current HEAD. It fails *at* the commit that introduced the behaviour change, not at some later point.

## The code is right; the test is stale

The test asserts that a port **carrying a USB serial** matches a registry entry that has **no serial recorded**, on the strength of a Windows port-path hash.

`c636d2dd fix(#503): serial-first global matching` deliberately removed that. `find_in_registry` now returns early on any port that exposes a serial, and the legacy port-path path is reachable only for serial-less ports. The code says so itself:

> `PASS 2 -- only for ports that expose NO serial`
> `A legacy hash match ... is weaker than a serial match and follows the socket, not the board.`

That was an **owner ruling**: VID:PID and port-path are device *class* and *socket*, never identity. Restoring the old behaviour would undo it.

**So the fix is to retire or rewrite the test, not to change `pio-flash.py`.**

## Consequence worth documenting while here

Any registry entry created before serials were recorded will now refuse to resolve if that device presents a serial. That is the intended tightening, not breakage — the remedy is re-running `pio-flash bootstrap` for the entry so its `usb_serial` is recorded. Worth a line in the refusal message or the docs, because "not attached to this host" does not obviously mean "your registry entry predates serial recording".

## Acceptance

- [ ] The test either asserts the current, ruled behaviour (serial-bearing port + serial-less entry → refuse) or is deleted with a note pointing at #503.
- [ ] `test_pio_flash_device_match.py` passes in full.
- [ ] No change to `pio-flash.py` matching logic.

## Why it went unnoticed for three weeks

CI runs **none** of the `pio-flash` test suites — tracked separately. This issue fixes the red test; that one stops the next one hiding.

Origin: `c636d2dd`, Agent WhiteFalcon (session `cb12cdd2`).

---
Agent: **FoggyCanyon** (Agent Mail `app-c-dev-crosswire`, session `2d2b37f1`)

Strycher:
[...]

Issue #852, opened 2026-08-19
task: sweep the full CLI get/set surface over serial with shape assertions


Strycher:
Parent epic: #774.

**Files: 1** — a sweep script under `scripts/`. Depends on #849 (redaction), #850 (batch mode), #851 (latency).

## What it does

Enumerate every CLI key **mechanically from source** — currently **47 `get`, 43 `set`, 51 distinct** — issue each over serial in one batch run, and assert each reply is well-formed.

Mechanical enumeration is the point. Hand-listing is exactly how `radio`, `agc.reset.interval` and `pwrmgt.bootmv` went ten months returning empty replies while everyone assumed the surface was covered.

## Assertions, not values

For each key: reply is non-empty, is a real answer rather than the `??: %s` fallback, has a plausible length, and arrived within the window.

**Secret-bearing keys are swept like everything else.** `prv.key`, `guest.password` and `bridge.secret` are tested for *shape*, never recorded by *value* — #849 redacts at the source with length preserved, so the assertions still hold. A deny-list was considered and rejected: excluding them means three keys nobody ever tests, which is the failure this whole epic exists to close.

## What this covers that the #775 guard does not

The guard is static. It catches an arm that is structurally incapable of replying — empty body, shadowed key. It cannot catch a key that compiles fine and still misbehaves at runtime: a handler that returns early on device state, one whose reply depends on hardware that is absent, or one that is simply slow.

This sweep is the runtime half of the same question.

## Not the over-the-air path

Serial only. The over-the-air CLI is admin-gated and needs a client, so it cannot be swept unattended — see #412 for that gap. Serial is also the right place to start: it establishes the baseline that makes an over-the-air discrepancy meaningful.

## Acceptance

- [ ] Key list generated from source at run time, never hardcoded.
- [ ] All 51 keys exercised over serial in one batch run.
- [ ] Every failure classified: empty reply, `??:` fallback, timeout, malformed.
- [ ] No secret value in the output — verified by inspecting the artifact, not assumed.
- [ ] Result committed as a baseline transcript so future runs are a diff, not a fresh judgement call.

---
Agent: **FoggyCanyon** (Agent Mail `app-c-dev-crosswire`, session `2d2b37f1`)

Strycher:
[...]

Issue #736, opened 2026-08-16
epic: repository infrastructure — structure, tooling, hooks and developer environment integrity


Strycher:
## Why this epic exists

Offband has 40+ epics covering firmware, hardware and features. **None covers the repository itself** — its structure, its tooling, its hooks, and whether the developer environment actually does what the documentation says it does.

That gap has already cost real work. #735 found that `HARDWARE.local.md` was symlinked into the **retired pre-Offband repo** (`Strycher/LoRa`) from 2026-06-14 to 2026-08-16, so **977 lines / 49 device entries of Offband hardware documentation were written outside this repository** — never diffed, reviewed or committed here. It survived two months because `CLAUDE.md` documented the symlink as correct, so every agent read the instruction and complied. The defect and its documentation reinforced each other.

That class of problem is invisible to feature-scoped epics. Nobody owns it, so nobody looks.

## Scope

Work that keeps the repository sound rather than adding firmware capability:

- **Repository structure and boundaries** — what belongs in this repo, what must never live outside it, no dependencies on retired or sibling repos
- **Tooling integrity** — `pio-flash`, `llm-consult`, `companion_harness`, the `dw`/Citadel path: do they resolve paths correctly, are they version-controlled, do their docstrings match their behaviour
- **Hook correctness** — flash/OTA/agent-mail guards, git lifecycle hooks, and whether they actually fire in the contexts where work happens
- **Documentation that agents execute** — `CLAUDE.md`, `HARDWARE.md`, `HARDWARE.local.md`, `REPOCONFIG`. These are not prose; agents follow them literally, so an error here propagates to every session until someone notices
- **Developer environment reproducibility** — worktrees, secrets handling, per-host vs committed state

## Known members and candidates

- **#735** — `HARDWARE.local.md` symlinked into the retired `Strycher/LoRa` repo (fix committed, PR pending)

Observed but not yet filed, surfaced while working #735:

- **`scripts/llm-consult.py` is gitignored** (`.gitignore:93`) and untracked. The Gemini review tool that the feature flow *requires* is not version-controlled in this repo. May be intentional (synced from standards) — needs a decision either way.
- **#639** — `core.hooksPath` is relative, so git-lifecycle hooks are **inert inside worktrees**, which is where all work actually happens. Already tracked; belongs under this epic.
[...]

PR #863, opened 2026-08-19
fix(#849): redact secret CLI replies in pio-flash send, at the source

### LINKED ISSUE #849: task: redact secrets in pio-flash send output, at the source
Parent epic: #736.

**Files: 1-2** — `scripts/pio-flash.py` (send output path), reusing existing `scripts/log_redact.py`.

## Why this is a precondition, not polish

`pio-flash send` prints the device's reply straight to stdout. Three CLI keys return secrets:

| Key | Returns |
|---|---|
| `get prv.key` | the node's private key — **serial-only** by design (`sender_timestamp == 0`), which is exactly the path serial tooling uses |
| `get guest.password` | a password |
| `get bridge.secret` | a shared secret |

Anything reading that stdout — an agent, a log, a session transcript — receives the value. Scrubbing afterwards is too late; by then it has already been written down.

The alternative considered and **rejected**: a deny-list that refuses to send those keys. It keeps them out of logs by never testing them, which leaves three keys permanently unverified. That is how the #764 keys went unnoticed for ten months.

## The mechanism already exists

`scripts/log_redact.py` carries a last-resort net for label-less CLI replies — `  > value` becomes `> <redacted:cli-reply>`. It exists because a real SSID leaked into chat this way (#382). It is simply not wired into `pio-flash send`.

## Acceptance

- [ ] `pio-flash send` routes reply output through `log_redact` before printing.
- [ ] A reply to a secret-bearing key prints redacted, with **length preserved** so shape assertions still work.
- [ ] Non-secret replies are unchanged — `get radio` must still print `> 910.525,62.5,7,5` verbatim, or every existing use of this tool regresses.
- [ ] Test covering both: one secret key redacted, one ordinary key untouched.

---

Strycher:
Closes #849. Under epic #736. First of the four tasks the CLI sweep (#852) needs.

## The leak

`pio-flash send` streamed the device's reply straight to stdout. Three CLI keys return secrets:

| Key | Returns |
|---|---|
| `get prv.key` | the node's private key — **serial-only by design** (`sender_timestamp == 0`), which is exactly the path this tool uses |
| `get guest.password` | a password |
| `get bridge.secret` | a shared secret |

Anything reading that stdout — an agent, a log, a session transcript — got the value. Scrubbing afterwards is too late. This is #382 in a second tool: a real SSID once reached chat as `  > <value>`, with no label for any pattern to match on.

## Two constraints that pull against each other

1. A secret reply must not reach stdout in the clear.
2. An ordinary reply must pass through **byte-for-byte** — `get radio` has to keep printing `> 910.525,62.5,7,5`, or every existing use of this tool regresses and #852 can assert nothing.

`log_redact`'s last-resort net redacts *any* `> value` line. It satisfies (1) and destroys (2). So the net became opt-out, and `send` — which already knows the command it issued — decides per command.

## Changes

**`log_redact.py`** — `redact_line` gains `cli_reply_net: bool = True`. Defaulting True means `_cap_serial.py` and `companion_harness.py` are untouched; they keep the blanket net, which is correct for capture paths that *don't* know the command. The marker now carries the redacted length — `<redacted:cli-reply:7>` — so #852 can assert a secret key **answered** without recording what it said.

**`pio-flash.py`** — `_is_secret_command()` matches the key **exactly after the verb, never by prefix**. Prefix matching is the defect class this repo just finished removing from the firmware CLI (#764); reintroducing it here would classify `get prv.keyfoo` as secret and, worse, could miss a real secret key on a near-match.

`_ReplyRedactor` buffers to line boundaries before redacting. **This is the part the issue didn't anticipate:** serial arrives in arbitrary reads, so a secret can straddle two of them — redacting each chunk as it lands misses `  > hun` + `ter2\r\n` entirely. The flush path covers a reply that never terminates its line.

## The rejected alternative

A deny-list refusing to *send* those three keys. The owner rejected it: it keeps secrets out of logs by never testing them, leaving three keys permanently unverified — which is exactly how #764's three keys spent ten months returning empty replies.

## Verification

**12 tests written first**, all failing on missing attributes, all passing now. The ones that matter:

- a secret split across chunk boundaries
- a trailing partial line with no newline
- length preserved through redaction
- **byte-identical passthrough** for ordinary replies
- `get prv.keyfoo` correctly *not* classified as secret

`test_log_redact.py` and `test_observer_log_redaction.py` still pass — neither other caller regressed.

**Gemini adversarial review** — SAFELANE, CLAUDE-BASE and the full diff, no question list supplied. No findings; it independently confirmed the chunk-boundary handling, the default-True safety for shared callers, and that the marker change is additive.

## What is NOT verified

**Hardware.** The end-to-end proof is `get prv.key` returning a redacted marker and `get radio` unchanged on a live board. Offered, not yet run.

**CI does not gate any of this.** There are zero references to `test_pio_flash` in `ci.yml` — four existing suites, none run, which is how the red test in #861 sat on `firmware-base` for three weeks. This PR adds a fifth ungated suite. Tracked as #862; deliberately not fixed here.

---

### DIFF
--- a/scripts/log_redact.py
+++ b/scripts/log_redact.py
@@ -76,6 +76,13 @@ def redact_positions(line: str) -> str:
 
 _SECRET_WORDS = (r"password|passwd|pwd|psk|secret|token|apikey|api_key|bearer|pass")
 
+# Named so a caller that already KNOWS which command it issued can opt out of
+# just this one rule (#849). `pio-flash send` has to keep printing
+# `> 910.525,62.5,7,5` verbatim for ordinary keys while still redacting
+# `get prv.key`; a blanket net cannot do both. The length is carried so a
+# sweep can assert that a secret key ANSWERED without recording what it said.
+_CLI_REPLY_RE = re.compile(r"(^\s*>\s*)([^\r\n]+)")
+
 _REDACTIONS = [
     # WiFi SSID, in the shapes the firmware actually prints.
     (re.compile(r"(SSID\s*=\s*)([^;\r\n]+)", re.IGNORECASE), r"\1<redacted:ssid>"),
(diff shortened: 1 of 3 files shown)


Strycher:

Re-validated against the remote immediately prior: **1 commit**, **3 files** (`scripts/log_redact.py`, `scripts/pio-flash.py`, `scripts/test_pio_flash_send_redaction.py`), `mergeable=MERGEABLE`, checks **48/48 SUCCESS**. Unchanged from what was approved.

`mergeStateStatus=BEHIND` — base moved while this sat. Updating the branch server-side, then auto-merge with rebase.

Landing **without hardware verification** — that was offered and not taken, and it is recorded on the PR body rather than glossed. The end-to-end proof remains `get prv.key` returning a redacted marker and `get radio` unchanged on a live board.

---

Issue #899, opened 2026-08-20
bug: unconfigured broker slots spam NVS ERROR logs at 29/sec — 97% of serial output, splices CLI replies


Strycher:
**Cause:** `readBrokerConfig()` reads ~10 string keys from NVS unconditionally, on every reconcile, for every broker slot — including slots that are **disabled and unconfigured**, where our own storage design guarantees those keys are absent. Arduino's `Preferences::getString()` logs at **ERROR** for each absent key. The MQTT worker re-drives every 500 ms.

**Ask:** stop reading keys we already know are absent.

## Measured on hardware

Captured on **ST-P** (`heltec_v4_repeater_telemetry`, current `firmware-base`), 346 s window, brokers unconfigured:

| | |
|---|---|
| NVS `ERROR` lines | **10,135** |
| Rate | **29.3 / sec**, continuous |
| Median gap | **9 ms** |
| Share of **all** serial output | **97.1%** (884,912 / 911,485 bytes) |

Six distinct lines, ~1,700 each:

```
[E][Preferences.cpp:483] getString(): nvs_get_str len fail: username NOT_FOUND
[E][Preferences.cpp:483] getString(): nvs_get_str len fail: password NOT_FOUND
[E][Preferences.cpp:483] getString(): nvs_get_str len fail: jwt_token NOT_FOUND
[E][Preferences.cpp:483] getString(): nvs_get_str len fail: jwt_email NOT_FOUND
[E][Preferences.cpp:483] getString(): nvs_get_str len fail: jwt_owner NOT_FOUND
[E][Preferences.cpp:483] getString(): nvs_get_str len fail: iata_override NOT_FOUND
```

**Not board-specific.** The owner reports the same noise while reading RC52 UART off the Rx pin, so it is live in the RC52 tree (#625) as well.

## Mechanism

`src/helpers/wifi_observer/ConfigSchema.cpp:141` — `enabled` is read **first and then ignored**:

```c
out.enabled = p.getBool(kKeyBrokerEnabled, false);   // read...
String url  = p.getString(kKeyBrokerUrl, "");        // ...never used to short-circuit
...
String u  = p.getString(kKeyBrokerUsername, "");     // 6+ more, all absent by design
```

Absent is the **designed** state. `writeBrokerConfig` removes empty fields rather than storing blanks, and the function's own comment says a missing key returning its default "is correct".

[...]

Issue #896, opened 2026-08-20
task: round-trip timing on single pio-flash send — reaches the volatile commands batch cannot


Strycher:
Parent epic: #892. Feature: #891. Serves #893.

**Files: 1** — `scripts/pio-flash.py`.

**What:** report round-trip timing on a single `pio-flash send`, the same fields batch already records.

## Why single `send` and not just batch

Batch mode is **`get`-only by construction** and stays that way — the owner's direction is that volatile commands do not ride in on one approval. `_is_read_only_command` refuses every command #893 needs:

```
wifi          -> REFUSED        get radio     -> ACCEPTED
wifi on 5     -> REFUSED        get prv.key   -> ACCEPTED
wifi off      -> REFUSED
wifi reset    -> REFUSED
```

`[verified: ran _is_read_only_command against each, this session]`

So the instrument built in #851 cannot take the measurement #893 asks for. Single `send` has no such gate — it is one command under one approval, which is exactly the right shape for a state-touching command a human is watching.

## Scope

- [ ] `--timing` (or always-on) reporting `elapsed_s`, `first_byte_s`, `reply_bytes` for the one command.
- [ ] Reuse `_read_until_idle(origin=...)` and `TransportReply` from #851 — **no second timing implementation.** Two copies drift, and the copy that drifts is the one that lies.
- [ ] Same origin discipline: the clock starts **before** the write, so `first_byte_s` and `elapsed_s` share a zero point and can be correlated.
- [ ] `--json-out` so a run is a record rather than a screenshot.
- [ ] Redaction unchanged — output still routes through `log_redact` per #849.

## What this can and cannot measure

**Can:** the serial round trip, end to end, for any command including volatile ones.

**Cannot:** the over-the-air path. On that path the client is the host, so there is no host-side clock to measure from. That is #895's job, and the two are complementary rather than alternatives — this one bounds the serial baseline, #895 attributes what happens inside the device.

## Acceptance

- [ ] `pio-flash send <dev> "wifi"` reports a round-trip time.
- [ ] Timing derives from the #851 helpers, not a parallel implementation.
- [ ] A fast command reports milliseconds rather than the read window — the fixed-window bug from #851's review must not reappear here.
- [ ] Secret-bearing replies stay redacted, with timing and length still reported.

---
Agent: **FoggyCanyon** (Agent Mail `app-c-dev-crosswire`, session `2d2b37f1`)

Strycher:
[...]

Issue #893, opened 2026-08-20
task: attribute the ~20-25s wifi CLI round-trip — handler, send path, or client


Strycher:
Parent epic: #774. Feature: #891.

**Reported:** the client-side session observed one or more `wifi` CLI commands taking **~20–25 s** to complete and return over the client path, and was *adamant that the time was spent in the repeater* rather than in the transport or the client.

That is a relayed claim from another agent, so it is a report to verify, not a finding. Verifying the half that can be settled from source before anyone puts a probe on it.

## The handlers cannot account for 20–25 s

`[verified: src/helpers/CommonCLI.cpp:983-1006 and examples/simple_repeater/main.cpp:645, read this session]`

| Command | What it does | Blocking? |
|---|---|---|
| `wifi reset` | `wifi_telemetry_reset_state()` + `strcpy` | no |
| `wifi off` | `wifi_telemetry_set_persistent(0)` + `strcpy` | no |
| `wifi on N` | sets a duration flag + `sprintf`. Comment: *"will reconnect on next publish"* | no — **it does not connect** |
| `wifi` | `wifi_telemetry_get_status()` — reads counters and `millis()`, formats with `snprintf` | no |

None of them touch the radio, the AP, or the network. There is no connect, no DNS, no broker handshake on this path. **The claim "the repeater took 20–25 s to run the command" is not supported by the source.**

## The reading that reconciles both claims

The repeater can still be where the time goes, without any handler being slow: **the reply is composed instantly and then waits to be sent.**

- The reply travels back over LoRa, subject to airtime, `TXT_ACK_DELAY`, retry and path selection.
- The CLI reply is emitted from the main loop. If `loop()` is mid-burst — a WiFi association or an MQTT publish — the reply waits for it. That is genuinely *the repeater's* latency, and equally genuinely **not the command's**.

Both statements can be true at once: time is spent on the repeater, and the handler is instantaneous. Those need separating before anyone changes code.

## Falsifiable test

Issue the same `wifi` commands over the **serial console**, which bypasses the LoRa send path entirely, and compare with the over-the-air round trip.

> **Prediction:** serial `wifi` / `wifi on N` return in **milliseconds**.
>
> - If serial is fast and over-the-air is 20–25 s → the time is in the send/schedule path, not the command. Attribute it there and stop looking at handlers.
> - If serial is **also** 20–25 s → the model above is wrong, something in `loop()` blocks, and that is a real firmware defect worth chasing.

[...]

Issue #892, opened 2026-08-20
epic: CLI verification toolchain — sweep the whole surface, safely and unattended


Strycher:
Parent Feature: #891.

**What:** the bench capability to interrogate the whole CLI surface — safely, unattended, with timing — so "does every key still work?" becomes a command instead of a research project.

## Why this needs to exist

#764 was three `get` keys returning an empty reply for ten months. The static guard from #775 now catches the *structural* form of that defect on every PR. It cannot catch a key that compiles fine and misbehaves at runtime: a handler that returns early on device state, one whose reply depends on absent hardware, or one that is simply slow.

Answering that needs a sweep, and a sweep was impossible for reasons that had nothing to do with the firmware:

- `pio-flash send` takes **one command per run**, and the flash discipline is **one human approval per run**. 47 `get` keys means 47 approvals — so it never gets run.
- Three keys return secrets. `get prv.key` returns the node's private key and is serial-only by design, which is exactly the path a serial sweep uses. A naive sweep writes it to a log.
- Nothing measures response time, so the owner's standing symptom — *CLI responses slow to return, sometimes never return, spotty handling* — has never been localised to firmware or client.

## Children, in dependency order

| Task | | State |
|---|---|---|
| **#849** | redact secrets in `send` output, at source | merged (`cb37fd39`) |
| **#850** | batch mode — N commands, one port, one approval | committed, unpushed |
| **#851** | per-command latency capture | complete, uncommitted |
| **#852** | the sweep itself — 51 keys, shape assertions | not started |

#849 is first deliberately: batch mode multiplies the blast radius of unredacted output, so one unlucky key in a 47-command run writes a secret to a log.

## The decision that shaped this epic

A deny-list refusing to *send* the three secret-bearing keys was proposed and **rejected by the owner**: it keeps secrets out of logs by never testing them, leaving three keys permanently unverified — which is exactly how #764's three keys survived. So the sweep sends everything and redacts the reply at source, with length preserved so shape assertions still hold.

## What "done" looks like

- One approved run interrogates all 51 keys and reports, per key: answered / empty / timed out, the reply, its length, and how long it took.
- No secret value appears in any output — verified by inspecting the artifact, not assumed.
- A serial baseline exists that can be contrasted against the over-the-air path, so the owner's slow-response symptom can be attributed rather than guessed at.

## Known limit, recorded rather than hidden

[...]

PR #905, opened 2026-08-20
epic(#892): CLI verification toolchain — sweep the whole surface, on hardware

### LINKED ISSUE #901: task: epic #892 integration — land the CLI verification toolchain and re-verify on hardware post-merge
**Cause:** epic #892's four task issues (#849, #850, #851, #852, #896) are complete and closed at testing. The epic itself needs one task that stays open to carry the push/PR through to merge, per the epic workflow — task issues close by or during testing; the PR rides on a separate open task, closed with the merge.

**Ask:** land `epic/892-cli-verification-toolchain` and verify the toolchain end-to-end on hardware after merge.

## What ships in this epic

| Task | What |
|---|---|
| #849 | secret redaction at source |
| #850 | `send-batch` — N read-only commands, one port, one approval |
| #851 | per-command latency capture |
| #896 | round-trip timing on single `send` |
| #852 | the CLI surface sweep |

Plus three fixes that only real hardware could surface:

- **`round_trip_s`** — `elapsed_s` measured until the read loop *returned*, which waits for the idle timeout, so every reading silently carried the tool's own 1.0s timeout. Invisible against instant test fakes; obvious on a board answering in 0.2s.
- **ESP32 port-open settle** — opening a port asserts DTR/RTS and reboots an ESP32. The 0.2s settle was tuned on an nRF52, which does not reboot on open. On ESP32 the first commands land mid-boot and return empty — **indistinguishable from the #764 defect the sweep hunts**, i.e. the tool would have manufactured its own findings.
- **Unanchored redaction** — the redactor anchored on a reply starting its line; #899's UART splicing moves it off column 0, and a node private key was written to a capture file in full.

## Evidence already gathered

Hardware run on **ST-P** (`heltec_v4_repeater_telemetry` @ `firmware-base`), 79 commands:

- **47/47 keys answer, 0 empty**
- **32/32 probes fall through, 0 prefix-match leaks**
- #764/#765 re-confirmed on ESP32-S3 (previously verified only on nRF52)
- Found #899 (P1)

**110 assertions, six suites, green after rebase.** Two adversarial review passes; 6 findings, 5 fixed, 1 justified.

## Definition of done

- [ ] PR opened against `firmware-base`
- [ ] `ci-green` passes (full matrix — a scripts-only change must still not break it)
- [ ] Human review and **merge** — a PR is not done until `gh pr view --json state` reports `MERGED`
- [ ] Post-merge: re-run the sweep from `firmware-base` and confirm identical results to the pre-merge run
- [ ] This task closed **with the merge**

## Known limitations shipping with this, recorded not hidden

- **`#ifdef` blindness** — the sweep enumerates keys from source without preprocessor awareness, so compile-gated keys report `unknown-key` on envs that exclude them. Observed live: 5 `bridge.*` keys on an env with no bridge flag. Real gap, not fixed here.
- **Settle is a timeout, not a probe** — a boot stage silent for >3.0s still defeats it. A timeout cannot distinguish "up" from "quiet" in principle; the complete fix is an active probe.
- **Timings are void on a chattering device** — surfaced deliberately as a `timings-void` finding rather than suppressed. #893 cannot be measured until #899 is fixed.

Strycher:
Closes #901 on merge. Epic #892, under Feature #891.

**Scripts only.** No firmware source, no pio env, no variant touched — `git diff --name-only origin/firmware-base...HEAD` is entirely `scripts/`.

## Why this exists

Three CLI `get` keys returned an empty reply for **ten months** (#764). Nothing caught it: the code compiled, no compiler called it dead, and the serial console concealed it because that caller zeroes its buffer and suppresses empty replies. Only the over-the-air path exposed it — as uninitialised stack on the wire (#765).

#775 (merged) is the **structural** guard. This is the **runtime** half: sweep every key against a real device and assert the shape of every reply.

## What landed

| Task | What |
|---|---|
| #849 | secret redaction at source |
| #850 | `send-batch` — N read-only commands, one open port, **one approval** |
| #851 | per-command latency capture |
| #896 | round-trip timing on single `send` |
| #852 | the CLI surface sweep |

Batch is `get`-only **by construction** — volatile commands are refused, not deny-listed. The sweep does **not** invoke `pio-flash` itself: flash discipline is one human approval per RUN, and a wrapper that shells out hides the run behind another layer. The operator issues it.

The key list is read **from firmware source every run**. Hand-listing is precisely how three keys hid for ten months.

## Hardware verification — ST-P, `heltec_v4_repeater_telemetry` @ `firmware-base`

79 commands live:

- **47/47 keys answer, 0 empty**
- **32/32 junk-suffix probes fall through, 0 prefix-match leaks**
- **#764/#765 re-confirmed fixed on ESP32-S3** — previously verified only on nRF52

```
get radio               -> > 910.525,62.5,7,5
get agc.reset.interval  -> > 0
```

The run also found **#899** (P1): unconfigured broker slots spam NVS ERROR at 29.3/sec — 97.1% of all serial output — and splice CLI replies mid-word.

## Three defects only hardware could surface

Recorded because each is a class, not a one-off:

1. **`round_trip_s`** — `elapsed_s` measured until the read loop *returned*, and that return waits for the idle timeout. Every reading silently carried the tool's own 1.0s timeout. Invisible against instant fakes; obvious on a board answering in 0.2s.
2. **ESP32 port-open settle** — opening a port asserts DTR/RTS and **reboots an ESP32**. The 0.2s settle was tuned on an nRF52, which doesn't. On ESP32 the first commands land mid-boot and return empty — **indistinguishable from the #764 defect being hunted.** The tool would have manufactured its own findings. Suppressing DTR was rejected: Arduino's ESP32 USB-CDC gates TX on DTR, so a device opened with it deasserted may never reply.
3. **Redaction anchor** — the redactor anchored on a reply starting its line. #899's UART splicing moves it off column 0, so a node private key was written to a capture file in full. *A firmware defect disabled a security control in a separate program by invalidating a structural assumption.*

## Adversarial review — `gemini-2.5-pro`, two passes, 6 findings

**Fixed (5):**

| Sev | Finding | Fix |
|---|---|---|
| CRITICAL | short secret leaked when a splice destroyed the `>` marker (`Loop hunter2` — 7 chars, no marker, no keyword) | fail closed: on a secret command, if no rule fired, replace the whole line. We don't need to *find* a value to refuse to print it |
| CRITICAL | the fail-closed branch was suppressed by `<redacted:` **arriving from the device** | keyed on `red == stripped` — whether we redacted is our own fact, read from our own output |
| HIGH | `??:` carve-out matched anywhere, so `... ??: prv.key > <key>` bypassed the guard | requires `startswith` |
| HIGH | settle exited early on a silent boot stage | grace 1.0 → 3.0s, ambiguity documented |
| HIGH/MED | timing overclaimed under chatter; a single legitimate log-bearing reply could void a run | `timings-void` finding, slow-detection disabled, threshold 2+ |
| LOW | unbounded `--commands-file` | `is_file()` + size check before read + count cap |

**Justified, not fixed (1 + notes):**

- **Settle is a timeout, not a probe.** A stage silent >3.0s still defeats it. A timeout cannot distinguish "up" from "quiet" in principle. Documented in-code as residual.
- **Marker length on the fail-closed path** is the line length, not the value length. The sweep asserts marker *presence*, not an exact length, and this only arises on a capture already known corrupted.
- **TOCTOU on `--commands-file`.** An attacker who can swap files between `stat()` and `read()` already owns the machine; this is a local dev tool. The FIFO/size-0 half *was* real and is fixed.
- **`#ifdef` blindness.** Enumeration is not preprocessor-aware, so compile-gated keys report `unknown-key` on envs excluding them — observed live (5 `bridge.*` keys). Real gap, recorded against #852.

## Evidence

**110 assertions across six suites, green after rebase onto current `firmware-base`.** Pure — no serial port, no device, no `pio-flash` invocation.

```
test_pio_flash_send_redaction.py   18    test_pio_flash_batch_latency.py   13
test_pio_flash_send_batch.py       24    test_pio_flash_send_timing.py     16
test_cli_sweep.py                  22    test_cli_dispatch.py              17
```

Re-analysing the real ST-P capture after the review fixes yields **one honest `timings-void` finding** where it previously produced **67 bogus `slow` findings**.

## Not claimed

- **#893 is not answered.** Every timing from the ST-P run is void — the 0.17–17.2s spread is idle-detection tripping over 29 log lines/sec, not latency. The 0.201s serial baseline from the T096 (quiet board) is the only clean number. **#893 needs #899 fixed first.**
- **CI runs none of these suites** (#862). These 110 assertions gate nothing on a PR yet.

### DIFF
--- a/scripts/cli_sweep.py
+++ b/scripts/cli_sweep.py
@@ -1,0 +1,245 @@
+#!/usr/bin/env python3
+"""Sweep the firmware's CLI surface and assert every key still answers (#852).
+
+WHY THIS EXISTS
+
+`get radio`, `get agc.reset.interval` and `get pwrmgt.bootmv` returned an empty
+reply for ten months (#764). Nothing caught it: the code compiles, no compiler
+calls it dead, and the serial console hides it because that caller zeroes its
+buffer and suppresses empty replies. Only the radio transport showed it -- as
+uninitialised stack (#765).
+
+The static guard from #775 now catches the STRUCTURAL form of that defect on
+every PR. It cannot catch a key that compiles fine and misbehaves at runtime:
+a handler that returns early on device state, one whose reply depends on
+hardware that is absent, or one that is simply slow. This is the runtime half.
+
+HOW IT RUNS, AND WHY IT IS SPLIT
+
+    enumerate-keys  ->  commands file  ->  [operator runs pio-flash]  ->  analyse
+
+The sweep does NOT invoke `pio-flash` itself. The flash discipline is one human
+approval per RUN of that tool, and a wrapper that shells out to it hides the run
+behind another layer. The operator issues the run and sees it happen.
+
+    python scripts/cli_sweep.py emit --out cmds.txt
+    python scripts/pio-flash.py send-batch <device> --commands-file cmds.txt \\
+        --json-out sweep.json
+    python scripts/cli_sweep.py analyse --json sweep.json
+
+THE KEY LIST COMES FROM SOURCE, EVERY RUN
+
+Hand-listing is precisely how three keys hid for ten months while everyone
+assumed the surface was covered. If a key exists in the dispatch chain, it gets
+swept -- including the serial-only ones, and including the secret-bearing ones,
+whose replies are redacted at source by #849 with length preserved so the shape
+assertion still holds.
+
+A deny-list that skipped the secret keys was proposed and rejected by the owner:
+it keeps secrets out of logs by never testing them, which leaves three keys
+permanently unverified. That is the failure this whole exercise exists to close.
+"""
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import sys
+from pathlib import Path
+
+# Matches a key literal in a dispatch arm, whichever matcher is used. `memcmp`
+# arms are included deliberately: #299's conversion left some behind, and those
+# are exactly the ones that go stale unnoticed.
+_KEY = re.compile(
+    r'\b(?:isKey|memcmp)\s*\(\s*config\s*,\s*"((?:[^"\\]|\\.)+)"'
+)
+
+class NoKeysFound(Exception):
+    """Enumeration produced nothing. That is a failure, not an empty surface."""
+
+
+DEFAULT_SOURCE = "src/helpers/CommonCLI.cpp"
+
+# A reply slower than this is surfaced. Not a failure -- a question worth
+# asking, and the thread #893 is pulling on.
+DEFAULT_SLOW_S = 1.0
+
+_PRINTABLE_OK = set("\r\n\t")
+
+
+def _fn_body(src: str, name: str) -> str:
+    """Brace-matched body of one CommonCLI method, or "" if absent."""
+    i = src.find(f"void CommonCLI::{name}")
+    if i < 0:
+        return ""
+    b = src.find("{", i)
+    if b < 0:
+        return ""
+    depth = 0
+    for j in range(b, len(src)):
+        if src[j] == "{":
+            depth += 1
+        elif src[j] == "}":
+            depth -= 1
+            if depth == 0:
+                return src[b:j + 1]
+    # Unbalanced braces. Returning the tail would sweep in every `isKey` call
+    # from every function that follows -- inventing keys that are not CLI keys
+    # at all. Fail closed: no body, no keys, and the caller's require_keys
+    # check turns that into a loud error rather than an empty sweep.
+    return ""
+
+
+def enumerate_keys(source: str, require_keys: bool = False):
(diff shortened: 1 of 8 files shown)


Strycher:

Strycher:
Human approval received from Ben in session 2d2b37f1 ("Merge #901 on green").

- 7 commits, 8 files (all `scripts/` — no firmware source, no pio env)
- `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`
- **48/48 checks SUCCESS**, including the full `ci-green` build matrix

Auto-merge enabled with `--rebase`. Integration task #901 closes with the merge.

Issue #891, opened 2026-08-20
feature: CLI surface integrity — every key answers correctly, on every transport, and stays that way


Strycher:
**What:** the CLI answers correctly on every key, over every transport, and keeps doing so.

**Why this is a Feature and not just a bug:** the defect that started it was one key returning nothing. The reason it survived ten months is that nothing in this repo could answer *"does every key still work?"* — not on the firmware side, and not from the bench. Closing that gap is the capability being built here.

## The problem this closes

A user upgrading from stock MeshCore 1.15.0 ran `get radio` in the client app and got mojibake back. The reply was uninitialised stack, transmitted because the handler wrote nothing and the caller never zeroed its buffer.

Three things let it survive:

1. **It compiled.** #299's prefix-to-exact-token conversion hoisted three `isKey` guards without their bodies. No compiler calls that dead code.
2. **Serial hid it.** The console caller zeroes its buffer and suppresses empty replies, so the symptom there was *silence*. Only the radio transport showed it — the transport nobody watches, and the one users are actually on.
3. **Nothing swept the surface.** 47 `get` keys, 43 `set` keys, and no way to ask all of them a question.

## Epics under this Feature

| Epic | Covers |
|---|---|
| **#774** — CLI dispatch and reply integrity | The defects and their fixes: #764, #765 (both fixed and hardware-verified), #657, #359, #312, plus the static guard #775, the gating decision #776, and the hardware integration test #777 |
| **CLI verification toolchain** (filed as this Feature's second child) | The bench capability that proves it and keeps proving it: #849 redaction, #850 batch mode, #851 latency capture, #852 the sweep itself |

Both belong under one Feature rather than two. The toolchain exists *because* of #764 — #852's acceptance criteria reference it directly — and separating them leaves behind a batch-mode serial tool with a secret deny-list and a latency field that nobody can explain.

## Deliberately not in scope

**`pio-flash`'s own test hygiene.** #861 (a stale test, red on `firmware-base` since 2026-07-31) and #862 (CI runs none of the four `pio-flash` suites) stay under **#736**. Those are about the bench tool's own coverage, not about the CLI.

## What "done" looks like

- Every `get`/`set` key returns a well-formed reply or an explicit error — never silence, never non-printable bytes — verified over **both** serial and the over-the-air client path.
- A static guard prevents the structural form of the defect from landing again, and its CI log demonstrably fires on the broken tree *before* the fix.
[...]

Issue #871, opened 2026-08-19
bug: ungated direct Serial writers corrupt framed protocol on shared-UART roles (CLI Rescue is the live risk)


Strycher:
Reported by **DustyBarn** during #704 RC32 bring-up. Filing on their behalf at the owner's direction. Their concern is sound; two specifics in their report did not check out, and both are recorded below so nobody wastes time on them.

## The concern

On shared-UART roles the framed companion protocol and the debug console run on the **same** `Serial`. The codebase already knows this and gates the MeshLog mirror accordingly:

```cpp
meshLogSetMirror(!usb_serial_interface.isConsoleSharedWithProtocol());   // examples/companion_radio/main.cpp:391
```

**But that gate only covers MeshLog.** Any writer that calls `Serial.print*` **directly** is not gated by `g_meshLogMirror` at all, so on a shared-UART role its plain text is spliced into the middle of binary protocol frames. The receiver's length-prefixed decoder then desyncs — it consumes the injected bytes as payload, shifts its frame boundaries, and resyncs mid-stream.

That is a real defect class: a length-prefixed protocol cannot share a transport with an ungated text writer.

## Ungated writers that DO exist

```
examples/companion_radio/MyMesh.cpp:3422   Serial.println("========= CLI Rescue =========");
examples/companion_radio/MyMesh.cpp:3447+  Serial.printf(...)   // ~20 call sites through enterCLIRescue()
examples/companion_radio/main.cpp:177      Serial.print("\n=== Offband build: ...")   // boot banner
```

Practical exposure: **CLI Rescue is the live risk.** If rescue mode is entered while a client is connected on a shared-UART role, every one of those writes corrupts the framed stream. The boot banner is comparatively safe — it fires before the protocol is carrying traffic — but it is still an ungated writer on the same line.

## Two specifics from the report that did NOT check out

Recording these so they are not chased:

1. **`crashLogDump` / a 5 s deferred CrashLog re-dump** — no such symbol exists in `examples/companion_radio/` or `src/helpers/`. Whatever the CrashLog output path is called, it is not that, and it was not found writing directly to `Serial` on this path.
2. **A `[clock] set by contacts-bootstrap` line** — this string does not appear anywhere in the tree.
3. The mirror gate is at **`main.cpp:391`**, not 403.

[...]

PR #909, opened 2026-08-20
fix(#899): route optional NVS string reads through a type-checked helper

### LINKED ISSUE #899: bug: unconfigured broker slots spam NVS ERROR logs at 29/sec — 97% of serial output, splices CLI replies

### LINKED ISSUE #908: task: #899 testing — land the NVS guarded-read fix and re-verify from firmware-base
**Cause:** #899's fix is implemented, guarded, and hardware-verified. Per the epic workflow, the push/PR rides on a testing task that stays open and closes **with the merge** — the work task closes at testing, not at the push gate.

**Ask:** land `bug/899-nvs-error-spam` and confirm the fix holds from `firmware-base`.

## Already verified before this task was filed

Hardware, ST-P (`heltec_v4_repeater_telemetry`), same 79-command sweep before and after:

| | Before | After |
|---|---|---|
| NVS ERROR lines | **10,135** | **0** |
| Rate | 29.3/sec | **0.0/sec** |
| Share of all serial bytes | **97.1%** | **0.0%** |
| Replies with spliced log output | **79/79** | **0/79** |

`cli_sweep analyse` → **79 commands, no findings.**

- Guard proven **red on the pre-fix tree** (15 findings, exit 1), clean after (260 files, exit 0)
- 17 guard assertions green
- 3 envs build SUCCESS
- Two adversarial review passes, 12 findings, all actioned or justified
- `get prv.key` confirmed **redacted on real hardware**, not just against fixtures

## Definition of done

- [ ] PR opened against `firmware-base`
- [ ] `ci-green` passes (full matrix — this touches firmware, unlike #892)
- [ ] Human review and **merge**; not done until `gh pr view --json state` reports `MERGED`
- [ ] Post-merge: re-run the sweep from `firmware-base` and confirm **0 NVS errors** still
- [ ] This task closed **with the merge**

## What this does NOT close

**#893 is not answered by this.** Time-to-first-byte was **0.014s before the fix and 0.015s after** — the device was always answering in ~14 ms. The large apparent round-trip improvement was the measuring tool's idle-detection waiting out the chatter, not device latency. UART utilisation from the spam was only ~15.8%, nowhere near saturation.

So the ~20–25 s the client sees is **still unexplained** and does not live in the CLI handler or the serial path. It remains open under #893, now finally measurable.

**#871 is not closed either.** This removes the traffic that made the splicing constant; it does not serialise the two UART writers. Any other component logging while a reply is in flight reproduces it.

Strycher:
Closes #908 on merge. Fixes #899, under Feature #891.

**Touches firmware** (unlike #892), so the full `ci-green` matrix is the real gate here.

## The defect is an API contract violated by its own implementation

Arduino's `Preferences::getString(key, default)` takes a default **precisely so a missing key is normal**, then logs at ERROR when it uses it:

```c
// Preferences.cpp:483
esp_err_t err = nvs_get_str(_handle, key, value, &len);
if(err){
    log_e("nvs_get_str len fail: %s %s", key, nvs_error(err));
    return String(defaultValue);
}
```

Any caller using that API as designed emits ERROR spam by construction. Our config makes it *continuous*: `writeBrokerConfig` deliberately **removes** empty fields rather than storing blanks (#182), so absent keys are the designed steady state, and the MQTT pool worker re-drives every slot on a 500 ms cycle.

Only `getString` is affected — the numeric getters log at `log_v`, which is why exactly the string keys appeared in the capture while `enabled`, `port` and `transport` did not.

## Why not the obvious fix

Short-circuiting `readBrokerConfig` on a disabled slot was proposed and **rejected by the owner**, for two reasons — the second worse than the first:

1. A slot can be **disabled but still hold a valid URL** — a legacy broker kept deliberately rather than cleared. Gating on `!enabled && empty-url` still reads every key for it.
2. `readBrokerConfig` is **shared with the CLI display path**. `ObserverCli.cpp:460` (`get mqtt.broker.<N>.<key>`) calls it with no enabled check and prints `cfg.username`. Any gate inside the reader makes the CLI report **empty for that legacy broker** — silently hiding config still present in NVS. Worse than log spam.

## The fix

```cpp
inline String prefStr(Preferences& p, const char* key, const char* dflt = "") {
    return p.getType(key) == PT_STR ? p.getString(key, dflt) : String(dflt);
}
```

**`getType(key) == PT_STR`, not `isKey(key)`.** `isKey()` is literally `getType(key) != PT_INVALID`, so this costs the same — but `isKey` returns true for a key holding a **non-string** type, and `getString` would then fail with `ESP_ERR_NVS_TYPE_MISMATCH` and log at ERROR anyway, reintroducing the defect. Found by review.

15 call sites converted (`ConfigSchema.cpp` 11, `WifiBootstrap.cpp` 2, `WifiConfigProvider.cpp` 2). Behaviour is otherwise **identical** — a missing key still yields the default — so every caller including the CLI display path is unchanged.

## Hardware verification — ST-P, same 79-command sweep before and after

| | Before | After |
|---|---|---|
| NVS ERROR lines | **10,135** | **0** |
| Rate | 29.3/sec | **0.0/sec** |
| Share of all serial bytes | **97.1%** | **0.0%** |
| Replies with spliced log output | **79/79** | **0/79** |

`cli_sweep analyse` → **79 commands, no findings.** Replies that were previously corrupted mid-word are intact (`get txdelay -> > 0.5`), and `get prv.key` is confirmed **redacted on real hardware**.

## Read the numbers correctly — this is not a latency fix

**Time-to-first-byte: 0.014 s before, 0.015 s after.** The device was always answering in ~14 ms. The large apparent round-trip improvement was the measuring tool's idle-detection waiting out the chatter, not device speed. UART utilisation from the spam was only **~15.8%** — nowhere near saturation.

So **#893's ~20–25 s is not explained by this** and does not live in the CLI handler or the serial path. It stays open, now finally measurable.

**#871 also stays open.** This removes the traffic that made splicing constant; it does not serialise the two UART writers. Any other component logging while a reply is in flight reproduces it.

## The guard

`scripts/check_nvs_guarded_reads.py` bans bare `getString` **repo-wide**, with the helper as the single exemption — granted only while the helper still contains its type check, so renaming a function to `prefStr` cannot disable it.

**Proven red on the pre-fix tree (15 findings, exit 1) and clean after (260 files, exit 0)** — per #712, a guard must be demonstrated catching the real defect.

Scope is repo-wide because I **measured instead of assuming**: my first version was scoped to one directory on the reasoning that a wider net would flag too much existing code. The tree contains three bare call sites in total. The narrow guard bought nothing and guaranteed recurrence.

## Adversarial review — two passes, 12 findings

**Fixed:** type mismatch (`isKey` → `getType == PT_STR`); scope too narrow (now repo-wide); `->` pointer receivers; calls split across lines; macros hiding the call (now *reported*, not missed); a fake helper forging the exemption via a comment; and — a **silent miss**, the worst kind — a `//` inside a string literal (`"http://…"`) blanking the rest of the line and swallowing a real violation.

**Justified, not fixed:**
- **TOCTOU** — a key erased between the type check and the read yields **one** stray line, against 29.3/sec continuously before. The Arduino API exposes no atomic "read if present".
- **Cost** — `getType()` probes up to ten `nvs_get_*` calls for a missing key. We trade one failing probe plus a ~60-byte UART write for ten index lookups. Worth it, but a real cost on a hot path.
- **`log_v` assumption** — if a future core release promotes the numeric getters to `log_e`, this guard must widen with it. An assumption about third-party code, recorded as such.

## Why no unit test on the firmware path

`readBrokerConfig` lives inside `#ifdef ARDUINO` and needs the real `Preferences` class; **no host shim exists in this tree** (the "host round-trip test" named in `ConfigSchema.cpp`'s comments is aspirational — nothing references it). Extracting a seam is the #714 testable-surface problem, not this task. Proven instead by the guard above plus the hardware table.

## Evidence

- **17 guard assertions green**
- Builds SUCCESS: `heltec_v4_repeater_telemetry`, `heltec_v4_companion_observer_wifi`, `Heltec_v3_companion_observer_wifi`
- Rebased onto current `firmware-base`, re-verified after rebase

### DIFF
--- a/scripts/check_nvs_guarded_reads.py
+++ b/scripts/check_nvs_guarded_reads.py
@@ -1,0 +1,273 @@
+#!/usr/bin/env python3
+r"""Refuse an unguarded `Preferences::getString()` on an optional pref (#899).
+
+WHY THIS EXISTS
+
+Arduino's `Preferences::getString(key, default)` takes a default parameter
+precisely so that a MISSING key is a normal outcome -- and then logs at ERROR
+when it uses that default:
+
+    // framework-arduinoespressif32/libraries/Preferences/src/Preferences.cpp:483
+    esp_err_t err = nvs_get_str(_handle, key, value, &len);
+    if(err){
+        log_e("nvs_get_str len fail: %s %s", key, nvs_error(err));
+        return String(defaultValue);
+    }
+
+Any caller using that API as designed emits ERROR spam by construction. Our
+broker config makes it continuous: `writeBrokerConfig` deliberately REMOVES
+empty fields rather than storing blanks (#182), so absent keys are the designed
+steady state, and the MQTT pool worker re-drives every slot on a 500 ms cycle.
+
+Measured on ST-P (`heltec_v4_repeater_telemetry`, brokers unconfigured):
+
+    10,135 ERROR lines in 346 s  =  29.3/sec, median gap 9 ms
+    884,912 of 911,485 serial bytes  =  97.1% of ALL output
+
+That is not merely noise. The log path and the CLI reply path write the same
+UART with no mutual exclusion and interleave MID-WRITE, so replies came back
+spliced inside a word (`not _supportesd`, `??: lonXil`) -- mojibake on a CLI
+reply, and the same corruption defeated secret redaction in `pio-flash`.
+
+WHAT THIS CHECKS
+
+Only `getString` is guarded, and that is deliberate rather than an oversight:
+the numeric getters log at `log_v` (VERBOSE), not `log_e` --
+
+    // Preferences.cpp, getUChar
+    log_v("nvs_get_u8 fail: %s %s", key, nvs_error(err));
+
+-- which is why exactly the string keys appeared in the capture while
+`enabled`, `port` and `transport` did not. Widening to `getUChar` would flag
+code that emits nothing. (If a future core release promotes those to `log_e`,
+this check must widen with it -- noted because it is an assumption about
+third-party code, not a fact about ours.)
+
+The rule is a bare-call ban rather than "an isKey must appear nearby":
+proximity checks need data-flow analysis to be sound, and a regex pretending to
+do one is how a guard develops false positives and gets switched off. One
+helper, one place to audit, mechanically checkable.
+
+    BAD:   String u = p.getString(kKeyBrokerUsername, "");
+    GOOD:  String u = prefStr(p, kKeyBrokerUsername);
+
+The helper is allowed exactly one `getString`, inside its own type check.
+
+SCOPE
+
+Repo-wide. An earlier version scoped this to `src/helpers/wifi_observer/` on the
+assumption that a wider net would flag too much existing code and get switched
+off. That assumption was never measured, and the measurement contradicted it:
+after the fix the whole tree contains THREE bare call sites. A narrow guard
+bought nothing and guaranteed the defect would recur elsewhere.
+"""
+from __future__ import annotations
+
+import argparse
+import json
+import re
+import sys
+from pathlib import Path
+
+DEFAULT_TARGETS = ["src", "examples"]
+
+# `.` AND `->`: a pointer receiver (`pp->getString(k)`) is the same defect, and
+# matching only the dot operator meant one keystroke defeated the guard.
+#
+# Requires a NON-EMPTY argument list. That is what separates a Preferences read
+# (always passes a key) from `HTTPClient::getString()`, which takes none -- a
+# real call in examples/simple_repeater/main.cpp that a repo-wide guard would
+# otherwise flag forever. DOTALL so a call split across lines still matches.
+_GET_STRING = re.compile(
+    r"\b([A-Za-z_][A-Za-z0-9_]*)\s*(?:\.|->)\s*getString\s*\(\s*(?!\))",
+    re.DOTALL,
+)
+
+# A macro that hides the call. It cannot be resolved without a preprocessor, so
+# it is REPORTED rather than silently missed.
+_MACRO_HIDING = re.compile(r"^[ \t]*#[ \t]*define\b[^\n]*getString[ \t]*\(", re.MULTILINE)
+
+HELPER_NAME = "prefStr"
+_HELPER_DEF = re.compile(r"\b" + HELPER_NAME + r"\s*\(")
+
+# Skip directories that are not ours to police.
+_SKIP_PARTS = {".pio", "build", "node_modules", ".git"}
+
+
+def _strip_comments(text: str) -> str:
+    """Blank `//` and `/* */` comments, preserving length and newlines.
+
+    Length preservation matters: findings are located by counting newlines up to
+    a match offset, so deleting characters would shift every reported line.
+
+    Blanking also closes two holes at once -- a commented-out call can no longer
+    read as a violation, and a comment containing the guard keyword can no
+    longer forge the helper exemption.
+    """
+    out = list(text)
+    i, n = 0, len(text)
+    while i < n:
+        c = text[i]
+        nxt = text[i + 1] if i + 1 < n else ""
+        # String and char literals are skipped WHOLE. Without this, the `//` in
+        #     const char* url = "http://example.com"; auto v = p.getString(k);
+        # starts a comment and blanks the rest of the line -- silently swallowing
+        # a real violation. A guard that misses quietly is worse than no guard.
+        if c in ('"', "'"):
+            quote = c
+            i += 1
+            while i < n:
+                if text[i] == "\\":
+                    i += 2
+                    continue
+                if text[i] == quote:
+                    i += 1
+                    break
+                i += 1
+            continue
+        if c == "/" and nxt == "/":
+            while i < n and text[i] != "\n":
+                out[i] = " "
+                i += 1
+        elif c == "/" and nxt == "*":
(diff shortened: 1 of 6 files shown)


Strycher:

Strycher:
Human approval received from Ben in session 2d2b37f1 ("Merge #909").

Re-validated against the remote immediately prior: 1 commit, 6 files, `mergeable: MERGEABLE`, `mergeStateStatus: CLEAN`, **48/48 checks SUCCESS** including the full `ci-green` matrix.

Auto-merge enabled with `--rebase`. Testing task #908 closes with the merge.

Issue #907, opened 2026-08-20
task(#177): move terminal-Failed client teardown off the rotation worker — resource-class-aware


Strycher:
**Design/PLAN task** — spun out of the [#848](https://github.com/OffbandMesh/meshcore-firmware/issues/848) fix and its Gemini review ([#906](https://github.com/OffbandMesh/meshcore-firmware/issues/906)). Umbrella: [#177](https://github.com/OffbandMesh/meshcore-firmware/issues/177). **Not urgent — the #848 hotfix already removes the crash.** This tracks the debt so the mitigation doesn't masquerade as the root fix.

## Root cause

When a broker reaches terminal `Failed`, the pool worker reaps it via `MqttBroker::releaseClient()`, which calls `esp_mqtt_client_stop()` + `esp_mqtt_client_destroy()` on a **mid-connect-failure** client **synchronously, on the `mqtt_worker` task's own stack**. That teardown:

- **is deep** — esp-tls/mbedTLS teardown depth varies with TLS state, cipher, and cert-chain; it drove `mqtt_worker` high-water to **340 bytes free of 6144** under one measured failure mode, and
- **blocks the worker ~3–4 s** while it joins the stuck mqtt task.

The #848 hotfix grew the worker stack 6144 → 10240 (~4.4 KB headroom). That is a mitigation, not a root fix: the deep, blocking path still runs in the rotation worker's context, and a broker with a longer cert chain / different cipher / different failure point could consume more (Gemini review, #906, finding 1).

## The real axis: resources = cores × internal-SRAM

The right teardown strategy is **resource-class-dependent**, and resources are two independent axes — number of cores, and RAM (dominated by internal SRAM, since FreeRTOS task stacks want internal SRAM, not PSRAM; PSRAM only relieves pressure *indirectly* by letting other allocations move off internal RAM). PSRAM does **not** directly pay for a teardown task's stack.

| Class | Example board | Teardown strategy that fits |
|---|---|---|
| **single-core / no-PSRAM** | RCC6 (ESP32-C6) | Deterministic worker-stack sizing (measured worst-case). No extra task — can't spare the stack; no 2nd core to hide the block behind. **This is the binding/floor case.** |
| **single-core / PSRAM** | ESP32-S2 + PSRAM | Persistent pre-allocated teardown task becomes affordable; block still time-slices on one core. |
[...]

Issue #906, opened 2026-08-20
task(#177): observer terminal-Failed broker health — integration test record


Strycher:
**Consolidated integration-test record for the observer broker-health terminal-`Failed` fix** ([#838](https://github.com/OffbandMesh/meshcore-firmware/issues/838) + [#848](https://github.com/OffbandMesh/meshcore-firmware/issues/848), which folds [#823](https://github.com/OffbandMesh/meshcore-firmware/issues/823)). Umbrella: [#177](https://github.com/OffbandMesh/meshcore-firmware/issues/177).

This task exists so the full test process + results live in one durable place instead of scattered across the fix-task comments. The PR for the fix rides this task; it closes on merge.

## What was tested

The change set (lands as one PR): escalation choke point + connect-layer escalation (#838), terminal-`Failed` reaper that preserves state and doesn't panic (#848), worker-stack bump 6144→10240.

- Code: `src/helpers/wifi_observer/MqttBroker.{h,cpp}`, `src/helpers/wifi_observer/MqttBrokerPool.cpp`
- Docs: `docs/architecture/2026-08-20-broker-health-terminal-failed.md` (tech spec), `docs/observer-cli-commands.md` (user-facing `Failed`-state section)
- Build under test: `Heltec_v3_companion_observer_wifi`, artifact 2026-08-19 23:23 (the image every hardware result below ran on)

### 1. Host unit tests — `test/test_broker_health/` (7 cases)
`BrokerHealth` sticky-penalty policy: single-success-does-not-rehab, success-bar (3 dwells), flaky-alternating-goes-terminal, terminal-threshold, backoff schedule. **All passing.**

### 2. #838 — connect-layer escalation (hardware, hv3-bench)
- **Repro:** blocked slot-1 broker IP at the router (REJECT tcp-reset). Pre-fix: ~50 `[mqtt-err]` events over ~30 min (`type=1 TCP_TRANSPORT tls_esp_err=0x8006`), **never** escalated, never reached `Failed`. `[verified]`
- **First fix refuted on hardware:** the boolean-guard-reset-on-`MQTT_EVENT_BEFORE_CONNECT` design left the penalty pinned at 1 — connect-refused retries fire repeated ERRORs with no `BEFORE_CONNECT` between them (pN trace: constant ~71 s error spacing, no backoff growth). `[verified]`
[...]

Issue #823, opened 2026-08-18
bug: terminal Failed broker is revived by reconcileSlot — #739 give-up does not stick


Strycher:
Follow-up to #739 (merged, `925b8126`). The sticky-penalty core of #739 works, but the **terminal `Failed` state does not stick** — a broker that reaches `Failed` is revived by the pool and re-enters rotation, contradicting the design ("operator must re-enable/reconfigure to clear it") and the PR's Gemini review (which certified "no path out of Failed through normal operation").

## Evidence (hardware soak, hv3-bench, ~21h, the merged build)

- s1 timeline: `…BK → 17×FL → HB → CO → UP` (recovers), runs healthy, then `19×FL → HB → UP` again. It **oscillates** FL↔healthy.
- **15 `ROTATE-OUT s1` events after s1 first went `Failed`** — it kept re-entering rotation.
- Heap jumps to ~87K on FL lines, so the reaper *does* release the client — but the slot is re-begun right after.
- Notably: s2/s3 had *more* TLS errors (s3: 50) yet never went terminal — the sticky penalty correctly rehabbed them via clean dwells. Only s1's failures ratcheted straight to penalty 8. `live=1/1` held 96% throughout, so **#739's actual budget-theft fix is working** — this issue is purely the terminal-give-up refinement.

## Root cause (pinned at the code layer)

`rt_.state` can only leave `Failed` via `MqttBroker::begin()` (which does `rt_ = BrokerRuntimeState{}`). `begin()`'s only automatic caller is `MqttBrokerPool::reconcileSlot()`. The `FL → HB` transition is the proof: `HeldBudget` is set *only* inside `begin()` (MqttBroker.cpp:217), so `reconcileSlot→begin` demonstrably runs on a `Failed` slot.

The worker's reaper (MqttBrokerPool.cpp:745) *intends* Failed to be terminal — its comment says "never re-driven (Failed is absent from the condition below)" — and the candidate-build loop does exclude Failed. But that invariant is enforced by **scattered caller-side exclusions**, and something still reaches `begin()` on a Failed slot (a promotion-path race and/or a subtle candidate-selection hole; the exact trigger is not yet pinned by static analysis — every obvious automatic path *looks* guarded). `reloadSlot()` is operator-CLI-only, so it is not the soak vector.

## Fix (enforce the invariant at the choke point, not via scattered exclusions)

[...]

PR #917, opened 2026-08-20
bug: terminal-Failed broker reap sticks and stops panicking (#838, #848)

### LINKED ISSUE #838: bug: connect-layer broker failures never escalate — dead/unreachable broker never reaches terminal Failed
Follow-up to #739 / sibling of #823. Discovered while hardware-validating #823 by blocking a broker's IP at the router.

## Symptom

A broker that is **unreachable at the TCP-connect layer** (server down, connection refused, host unreachable, connect timeout) **never escalates its health penalty and never reaches terminal `Failed`.** It retries forever at the backoff floor, permanently consuming a rotation turn — exactly what #739's `Failed` state was meant to prevent ("a persistently-unreachable configured broker stops consuming a rotation turn every backoff cycle").

## Evidence (hv3-bench, #823 build)

Blocked slot 1's broker IP at the router (REJECT, tcp-reset). Over ~30 min, s1 logged **50 `[mqtt-err]` events** (`type=1` TCP_TRANSPORT, `tls_esp_err=0x8006` — a connect-layer failure), yet **never entered Backoff, never escalated, never reached `Failed`**, with zero penalty resets. `live=1/1` held on the healthy brokers throughout.

## Root cause `[verified: MqttBroker.cpp]`

Penalty escalation (`noteFailure()`) is called only from `onDisconnected()` (line 555) and two `tryConnect` client-start paths — **never from `onError()`** (line 567, comment: *"Errors typically precede a DISCONNECTED"*).

That assumption is false for **connect-layer** failures: esp-mqtt fires `MQTT_EVENT_ERROR` **only** when the initial TCP/TLS connect fails (nothing was established, so no `MQTT_EVENT_DISCONNECTED` follows). `onError` records the reason but does not escalate → the penalty never climbs.

The organic `Failed` seen earlier only happened because that broker was **flaky-but-reachable** — it connected, then dropped → `DISCONNECTED` → escalated. A genuinely dead broker never gets there.

## Fix

Escalate on a connect-layer failure too, **exactly once per attempt** (no double-count when a mid-session drop legitimately fires ERROR+DISCONNECTED):

- Reset a per-attempt guard on `MQTT_EVENT_BEFORE_CONNECT`.
- `onError` and `onDisconnected` both route through an `escalateFailureOnce()` that no-ops if the attempt already escalated.

This catches all three failure shapes once each: ERROR-only (connect refused/unreachable), ERROR+DISCONNECTED (TLS/auth after connect), DISCONNECTED-only (clean drop of an established session).

Stacks on #823 (which keeps `Failed` terminal once reached). Together they finally deliver #739's promise for the common "server down" case. Validation: re-run the router-block test → s1 climbs to `Failed` and stays out while healthy brokers hold `live=1/1`.

### LINKED ISSUE #848: bug: terminal Failed handling is broken — reaper clears the Failed state (never sticks) AND panics the device on teardown
**Rescoped 2026-08-19 to cover the full terminal-`Failed` reaper defect; folds in #823.**

When an MQTT broker reaches terminal `Failed` (penalty ≥ `OFFBAND_BROKER_TERMINAL_PENALTY`), the pool's reaper (`MqttBrokerPool` worker → `MqttBroker::releaseClient()`) mishandles it in two ways that share one code path.

## Facet A — the `Failed` state is cleared, so it never sticks (was #823)

`[verified: MqttBroker.cpp releaseClient() ~line 164; hv3-bench pN trace shows DNp8]`

`releaseClient()` ends with `rt_ = BrokerRuntimeState{}`, which resets `state` to `Down`. So when the reaper reaps a `Failed` broker's client it **erases the Failed state** (penalty retained → observed as `DNp8`: Down, penalty 8). The broker is then plain `Down`, gets re-promoted, fails, re-terminals, and the cycle repeats.

This is the true root cause of #823. The original #823 framing ("revived by `reconcileSlot`") was **wrong** — the `reconcileSlot` Failed-guard I wrote never fires, because the state is already `Down` by the time `reconcileSlot` runs. Correct fix: the reaper must free the client **without** clearing `Failed`.

## Facet B — the teardown panics the device (the original #848)

`[verified: 2x Guru Meditation panics at terminal on hv3-bench]`
```
Guru Meditation Error: Core 0 panic'ed (Unhandled debug exception).
Debug exception reason: Stack canary watchpoint triggered (IDLE0)
Backtrace: ... 0xa5a5a5a5 |<-CORRUPTED   (FreeRTOS context-switch; app culprit lost to stack-fill)
Rebooting...  rst:0xc (RTC_SW_CPU_RST)
```
Reaching terminal reliably panics + reboots (stack overflow). On reboot the penalty resets, so it **crash-loops** for any unreachable broker.

`[hypothesis: untested]` Prime suspect: `esp_mqtt_client_stop`+`destroy` on a client that is **mid-connect-failure** (auto-reconnect ON) — the one thing that differs from the thousands of clean `releaseClient` calls the 7h #739 soak ran on *healthy* rotated-out clients. **Not yet confirmed to a line:** coredump-to-flash is not enabled on this build, so the exact overflowing task is unconfirmed. Next step: enable coredump, re-trigger, decode.

## Sizing

Both facets are the **same code path** (the terminal-`Failed` reap/teardown in `releaseClient`/the reaper), and **one coherent fix** addresses both — preserve `Failed`, and quiesce-then-destroy. They are two symptoms of one broken teardown, not separate concerns, so this is one issue. #838 (escalation) stays a **separate** issue — different code (`MqttBroker` event handlers), independently validated.

## Gating

- **#838 must not land without this fix.** #838 is what makes brokers reach terminal; shipping it alone converts this dormant crash into an active crash-loop.
- Latent since #739 (`2fef4368`); never triggered because escalation never reached terminal until #838.

Folds in #823 (closed as superseded).

Strycher:
Fixes the observer MQTT broker pool's terminal-`Failed` lifecycle end-to-end. Rides testing task #906; root-fix debt tracked in #907.

## What changed and why

Three coupled defects on the broker reliability line (#177):

### #838 — connect-layer failures never escalated
Penalty escalation lived only in `onDisconnected()`, but a broker that is unreachable at the TCP-connect layer fires `MQTT_EVENT_ERROR` with **no** following `DISCONNECTED`. So a genuinely dead broker never escalated and never reached terminal `Failed` — it retried forever at the backoff floor, permanently consuming a rotation turn (exactly what #739's `Failed` state was meant to prevent).

**Fix:** all failure paths (`onError`, `onDisconnected`, both synchronous `tryConnect` failures) route through one `escalateFailureOnce(now_ms)`. The dedup decision lives in a new dependency-free, host-tested seam — `offband::FailEscalateWindow` (`FailEscalateWindow.h`) — which collapses the up-to-three esp-mqtt events of one attempt into a single escalation while counting genuinely separate attempts. (An earlier BEFORE_CONNECT-reset boolean was refuted on hardware — connect-refused retries fire repeated ERRORs with no BEFORE_CONNECT between them, pinning the penalty at 1.)

### #848 Facet A — terminal `Failed` did not stick (folds #823)
`releaseClient()` ended with `rt_ = {}`, resetting state to `Down`. So when the reaper freed a `Failed` broker's client, it **erased the Failed state** (penalty retained → `DNp8`), the broker was re-promoted, re-failed, and churned. This is the true root cause of #823 (whose "revived by reconcileSlot" framing was wrong — the state was already `Down` before reconcile ran).

**Fix:** `releaseClient(keep_failed=true)` on the reap path preserves `state=Failed` while freeing the ~60 KB client; revive paths (`tryConnect`/`reconcileSlot`/`onConnected`) early-return on `Failed`; `clearTerminalFailure()` clears it on operator reconfigure.

### #848 Facet B — the teardown panicked the device
`esp_mqtt_client_stop`+`destroy` on a **mid-connect-failure** client drove the `mqtt_worker` stack high-water to **340 bytes free of 6144**, intermittently overflowing into the adjacent IDLE0 task's stack canary → panic + reboot → penalty reset → crash-loop. (The coredump writer then double-faulted on the exhausted stack, which is why the raw backtrace was corrupted; pinned via stack-HWM instrumentation.)

**Fix:** `mqtt_worker` stack 6144 → 10240 (~4.4 KB headroom, ~13× the measured floor). This is a mitigation; the root fix (moving the deep teardown off the rotation worker, resource-class-aware) is tracked in #907.

## Gemini adversarial gate

`scripts/llm-consult.py --backend gemini --model gemini-2.5-pro`, hostile prompt, four findings:

- **F2 (med), F3 (low) — fixed.** Escalation window never reset on a clean connect (a failure after a genuine reconnect was masked); and a `millis()==0` escalation was double-counted. Both dissolved by the `FailEscalateWindow` seam (explicit `active_` flag, `reset()` on `onConnected`).
- **F1 (high) — justified + debt filed (#907).** +4 KB worker stack is a mitigation, not a root fix. `MAX_LIVE_TLS=1` → one teardown at a time; ~13× headroom; the soak drove two different wss brokers (different cert chains) to terminal with 0 crashes.
- **F4 — withdrawn by Gemini itself** (the time-window guard serialises the counter).

## Evidence

- **Host tests: 13/13** — `test/test_broker_health/` (7) + new `test/test_fail_escalate_window/` (6: paired-collapse, separate-attempts, `millis()==0` no-double-count, reset-unmasks-next, flapping-keeps-escalating). Re-run green post-rebase.
- **Firmware build** `Heltec_v3_companion_observer_wifi` **SUCCESS** (RAM 35.1%, Flash 44.5%), post-rebase.
- **Hardware (hv3-bench)** — full record in #906: connect-layer escalation p1→p8 (#838); ~45-min soak with two independent terminal reaps against two blocked wss brokers — **0 panics**, both `Failed` held (257/247 samples, 0 revivals), `live=1/1` throughout, no heap leak.

## Docs

- `docs/architecture/2026-08-20-broker-health-terminal-failed.md` — tech spec (states, sticky penalty, escalation choke point, terminal mechanics, auto-reconnect trade-off #746, constants, validation).
- `docs/observer-cli-commands.md` — user-facing "Broker health & the `Failed` state" section.

Closes #838
Closes #848

### DIFF
--- a/docs/architecture/2026-08-20-broker-health-terminal-failed.md
+++ b/docs/architecture/2026-08-20-broker-health-terminal-failed.md
@@ -1,0 +1,169 @@
+# Broker Health, Backoff Escalation, and the Terminal `Failed` State
+
+**Date:** 2026-08-20
+**Status:** As-built (lands with #838 + #848)
+**Area:** `src/helpers/wifi_observer/` — `MqttBroker`, `MqttBrokerPool`, `BrokerHealth`
+**Issues:** #739 (penalty + terminal state), #838 (connect-layer escalation), #848 (Failed sticks + no crash; subsumes #823), #746 (auto-reconnect trade-off)
+
+## Why this exists
+
+The observer keeps a rotating set of TLS/wss MQTT broker connections, at most
+`OFFBAND_MAX_LIVE_TLS = 1` live at a time (heap on the ESP32 fits ~2 mbedTLS
+contexts before an OOM reboot). A configured broker can be healthy, flaky, or
+dead. This document is the design-of-record for how the pool tracks a broker's
+reliability, how it escalates backoff, and how a persistently-dead broker is
+given up on **without** starving the healthy ones, crashing the device, or
+churning a rotation slot forever.
+
+## Broker states (`BrokerState`)
+
+| State | Meaning |
+|---|---|
+| `Down` | disabled, or not yet attempted |
+| `Connecting` | TCP connect / TLS handshake in progress |
+| `Up` | connected and serving |
+| `Backoff` | failed; waiting out the (penalty-scaled) backoff window |
+| `HeldNoClock` | wss/TLS deferred until the wall clock is sane (NTP/GPS). Not a failure |
+| `HeldNoHeap` | TLS bring-up deferred: free heap below `OFFBAND_TLS_HEAP_FLOOR_BYTES`. Not a failure |
+| `HeldBudget` | TLS bring-up deferred: the pool already holds `OFFBAND_MAX_LIVE_TLS` live contexts — normal rotation waiting its turn. Not a failure |
+| `Failed` | **terminal** — given up on after sustained failure. Dropped from rotation, client reaped, logged loudly. Cleared only by operator reconfigure |
+
+## Health model — the sticky penalty (`BrokerHealth`)
+
+`BrokerHealth` is a dependency-free, host-testable policy object
+(`test/test_broker_health/`). It owns *how a broker's reliability maps to
+backoff, rehabilitation, and a terminal give-up*.
+
+- `fail_penalty` escalates on each failed connection attempt and is **sticky in
+  both directions**: a single success does **not** clear it (a flaky broker
+  cannot cheaply reset), and the broker must earn its way back with
+  `OFFBAND_BROKER_SUCCESS_BAR` (3) **consecutive** clean full-dwells.
+- Backoff length is `brokerBackoffMs(fail_penalty)` (5 / 15 / 30 / 60 / 120 s,
+  then 120 s capped) — it tracks the sticky penalty, **not** a per-attempt
+  counter that a lone success would zero.
+- Past `OFFBAND_BROKER_TERMINAL_PENALTY` (8) the broker is declared `Failed`.
+
+### Why a sticky penalty (the #739 problem)
+
+Before #739 the only backoff state was `retry_count`, reset to 0 on **any**
+successful connect. That is fine for a *dead* broker but wrong for a *flaky*
+one: an intermittent broker connects just often enough to zero its penalty, so
+it oscillates at the 5 s floor forever and keeps claiming a rotation turn equal
+to a reliable broker. Measured on hv3-bench: two ~50%-failing remote JWT brokers
+held 46 % of a soak with **no** TLS broker up.
+
+## Escalation — one choke point, time-window deduped (#838)
+
+Every failure routes through a single `MqttBroker::escalateFailureOnce(now_ms)`,
+which calls `BrokerHealth::onFailure()` at most once per
+`OFFBAND_FAIL_ESCALATE_WINDOW_MS` (5 s). The dedup decision itself lives in a
+dependency-free, host-tested seam — `offband::FailEscalateWindow`
+(`FailEscalateWindow.h`, `test/test_fail_escalate_window/`) — so the escalation
+logic that was hardware-only in #838 now has unit coverage. Callers:
+
+- `onDisconnected` — an established connection dropped.
+- `onError` — a connection error (**including a connect-layer refusal /
+  unreachable / timeout that fires `MQTT_EVENT_ERROR` with no following
+  `DISCONNECTED`**).
+- the two synchronous `tryConnect` failures (JWT re-apply, `client_start`).
+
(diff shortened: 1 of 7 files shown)


Strycher:

Strycher:
Human approval received from Ben in session 82e07da7. Re-validated against the remote immediately prior: 1 commit, 7 files, mergeable=MERGEABLE, mergeStateStatus=CLEAN, checks=48/48 SUCCESS (incl. ci-green).

PR #918, opened 2026-08-20
bug: broker dwell-reset can block rotation — evict on a reconnect-proof budget-hold clock (#720)

### LINKED ISSUE #720: bug: a broker reconnecting faster than the dwell resets went_up_ms and blocks rotation entirely
## Cause

`rotateTlsIfDue()` picks its victim as the live TLS broker with the greatest age, and refuses to rotate if that age is under one dwell:

```cpp
uint32_t age = now_ms - b.runtime().went_up_ms;
if (victim == 0xFF || age > best_age) { best_age = age; victim = s; }
...
if (best_age < MQTT_ROTATE_DWELL_MS) return;   // "do not thrash"
```

`went_up_ms` is re-stamped on **every** transition into `Up` (`MqttBroker.cpp:531`), including an esp-mqtt auto-reconnect after a dropped connection.

With `OFFBAND_MAX_LIVE_TLS = 1` there is exactly one `Up` broker at a time. If that broker's connection flaps more often than `MQTT_ROTATE_DWELL_MS` (60 s), its age is reset before it ever reaches the threshold, `rotateTlsIfDue()` returns early every pass, and **rotation never fires**. Every other enabled TLS broker stays queued indefinitely.

## Impact

A single flaky broker — bad DNS, lossy uplink, a server that drops idle connections — silently holds the only TLS budget slot forever and starves every other feed. The observer looks healthy: one broker `Up`, others `waiting-turn` (#715), no errors.

This is the same failure class as #708 (rotation liveness depending on a state that can be reset) but a different mechanism, and #708's fix does not address it: #708 decides *who* gets promoted once the budget frees, while this prevents the budget from ever freeing.

## Not introduced by #708

Pre-existing since `3d125440 feat(#175): dwell-timer TLS rotation with per-slot cooldown`. Surfaced while dispositioning a round-2 Gemini finding about `last_known_state_` re-stamping on reconnect (#708) -- the same reconnect path resets `went_up_ms`.

## Not yet observed

This is a code-path finding, not a field observation. hv3-bench has not shown a flapping broker. Filed so it is tracked rather than lost; needs a repro before a fix is designed.

## Candidate directions (not a decision)

- Track a separate "budget held since" timestamp that a reconnect does **not** reset, and drive eviction from that rather than `went_up_ms`.
- Cap total budget occupancy regardless of connection state: if a slot has held the client for N dwells, it rotates out even if it keeps reconnecting.
- Count reconnects and treat a flapping broker as a rotation candidate sooner rather than later.

The first is probably the smallest correct change: `went_up_ms` legitimately means "connected since" for display/age purposes, and eviction wants a different clock.

## Acceptance

- A broker reconnecting every 30 s does not prevent other enabled TLS brokers from being promoted.
- Regression test in `test/test_broker_rotation/` covering the flapping case, alongside the existing fairness tests.

Strycher:
Fixes a silent rotation-starvation bug in the observer's TLS broker pool. Under epic #177.

## What changed and why

`rotateTlsIfDue()` chose its eviction victim's age from `runtime().went_up_ms`, which `onConnected()` re-stamps on **every** transition to `Up` — including esp-mqtt's silent auto-reconnect. With `OFFBAND_MAX_LIVE_TLS = 1` there is one live broker at a time; if it flaps more often than the 60 s dwell, its age resets below the eviction threshold every pass, `rotateTlsIfDue()` returns early forever, and **every other enabled TLS broker is starved** — while the observer looks healthy (one broker `Up`, no errors).

**Fix:** a reconnect-proof clock.

- **`BudgetHoldClock.h`** (new, dependency-free seam — same pattern as `BrokerHealth` / `FailEscalateWindow` / `BrokerRotationSelect`): records the budget-hold start **once per occupancy** (`held_` bool, `millis()==0`-safe), not reset by a reconnect, cleared by `releaseClient()`.
- `MqttBroker` holds a `BudgetHoldClock budget_hold_`: `onConnected()` stamps it, `releaseClient()` clears it, `budgetHeldMs()` exposes it.
- `rotateTlsIfDue()` ages the victim by `budgetHeldMs()`, not `went_up_ms`. `went_up_ms` is kept for display only.

### Design note (deviation from the filed plan)
The issue suggested extracting a `selectEvictionVictim` seam. In the code that proved the wrong target: the eviction *selection* (oldest-held + dwell-gate) is unchanged by this fix and takes the age as input, so a selection-seam test couldn't catch the clock-choice regression. The **clock** is where the defect lives, so `BudgetHoldClock` is the extracted, host-tested seam instead — its `ReconnectDoesNotResetTheEvictionClock` test is the actual #720 guard.

## Gemini adversarial gate

`scripts/llm-consult.py --backend gemini --model gemini-2.5-pro`, hostile prompt, 3 findings:

- **F1 (raised HIGH) — fixed.** A stale `MQTT_EVENT_CONNECTED` dispatched after `releaseClient()` could set `state=Up` + stamp `budget_hold_` with no client behind it (a "zombie" holding the budget slot). Guarded: `onConnected()` now ignores a connect event when `!hasClient()`, matching the existing #848 `Failed` guard. (Gemini called it permanent starvation; it was actually self-healing — the zombie becomes the oldest eviction candidate and rotates out within a dwell — but the inconsistency is closed regardless.)
- **F2 (LOW) — fixed.** The `[rot]` diagnostic log showed `went_up_ms` age while eviction uses `budgetHeldMs`; the log now shows the budget-hold age the logic keys off, so telemetry matches the decision.
- **F3 (LOW) — justified.** Full `rt_` reset on rotation "laundering" failure history is pre-existing and intentional (a rotated-out broker was just credited a clean dwell; the sticky `health_` penalty preserves the real reliability signal). Out of #720 scope.

## Evidence

- **Host tests: 14/14** in `test/test_broker_rotation/` — the existing promotion/fairness suite plus 5 new `BudgetHoldClock` cases: reconnect-doesn't-reset (the #720 regression), fresh-hold-not-yet-evictable (anti-thrash), release-resets-for-next-occupancy, `millis()==0` genuine-hold, held-time-advances.
- **Firmware build** `Heltec_v3_companion_observer_wifi` **SUCCESS** (post-Gemini-fixes).
- **Hardware bench-repro:** in progress — an intermittent ~30 s router block to make a broker actually flap, confirming rotation resumes (queued brokers get promoted) where pre-fix it stalls. Results will be posted here before merge.

Closes #720

### DIFF
--- a/src/helpers/wifi_observer/BudgetHoldClock.h
+++ b/src/helpers/wifi_observer/BudgetHoldClock.h
@@ -1,0 +1,51 @@
+#pragma once
+#include <cstdint>
+
+// Reconnect-proof clock for how long a broker has held the single TLS budget
+// slot, used to decide rotation eviction (#720). Dependency-free (no
+// Arduino/ESP headers) so it builds in [env:native] -- same pattern as
+// BrokerHealth / FailEscalateWindow / BrokerRotationSelect.
+//
+// WHY THIS EXISTS: rotation eviction used `went_up_ms` ("connected since") for
+// the victim's age. `went_up_ms` is re-stamped on EVERY transition to Up,
+// including esp-mqtt's silent auto-reconnect. With OFFBAND_MAX_LIVE_TLS = 1
+// there is one Up broker at a time; if it flaps faster than the dwell, its age
+// resets below the eviction threshold every pass, rotation never fires, and
+// every other enabled TLS broker is starved forever (the observer looks healthy:
+// one broker Up, no errors). Eviction needs a clock a reconnect does NOT reset.
+//
+// A `held_` flag -- not `since_ms_ == 0` -- marks occupancy, so a real
+// acquisition at millis()==0 is unambiguous (same sentinel discipline as
+// FailEscalateWindow). onConnected() stamps the start ONCE per occupancy;
+// reconnects (auto or pool-driven Backoff->retry, which keep the client and do
+// not release the budget) leave it. onReleased() clears it when the client is
+// destroyed (rotation eviction or terminal reap), so the next acquisition
+// stamps fresh.
+
+namespace offband {
+
+class BudgetHoldClock {
+public:
+    // Call on each transition to Up. Records the hold start once per occupancy;
+    // a reconnect (held_ already true) does not move it.
+    void onConnected(uint32_t now_ms) {
+        if (!held_) { held_ = true; since_ms_ = now_ms; }
+    }
+
+    // Call when the budget is released -- the client is destroyed (rotation
+    // eviction or terminal-Failed reap). The next onConnected() starts fresh.
+    void onReleased() { held_ = false; since_ms_ = 0; }
+
+    bool held() const { return held_; }
+
+    // How long the budget has been continuously held, wrap-safe. 0 if not held.
+    uint32_t heldMs(uint32_t now_ms) const {
+        return held_ ? (uint32_t)(now_ms - since_ms_) : 0;
+    }
+
+private:
+    bool     held_     = false;
+    uint32_t since_ms_ = 0;
+};
+
+}  // namespace offband
--- a/src/helpers/wifi_observer/MqttBroker.cpp
+++ b/src/helpers/wifi_observer/MqttBroker.cpp
@@ -175,6 +175,8 @@ void MqttBroker::releaseClient(bool keep_failed) {
     } else {
         rt_ = BrokerRuntimeState{};
     }
+    budget_hold_.onReleased();  // #720: the budget slot is freed on either path;
+                                // the next acquisition stamps a fresh hold clock.
     // slot_ and cfg_ deliberately RETAINED -- that is the whole difference from
     // shutdown(). A rotated-out broker must stay configured so the pool can see
     // it, honor its cooldown, and promote it back via reconcileSlot->tryConnect.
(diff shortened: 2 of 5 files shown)


Strycher:

Strycher:
## Hardware bench-repro — fix confirmed (hv3-bench)

Flashed the #720 build (\`3dec010f\`) to hv3-bench (MAC \`3c:0f:02:...\`, NVS-preserved), then flapped the live okimesh broker with a 20 s-block / 20 s-unblock router cycle (reconnect ~every 40 s, faster than the 60 s dwell) while capturing \`[rot]\`.

**Result — the flapped broker ages out instead of holding the slot forever:**

```
[rot] ROTATE-OUT s1 age=81s ...     <- flapped okimesh broker, evicted at 81s
[rot] ROTATE-OUT s2 age=60s ...     <- non-flapped brokers, normal dwell
[rot] ROTATE-OUT s3 age=60s ...
(6 ROTATE-OUT events across the run)
```

- **s1 (the flapped broker) reached age=81 s and was evicted.** Pre-fix, `went_up_ms` resets on every reconnect, so a broker reconnecting every ~40 s would cap at ~40 s and **never** reach the 60 s dwell — the exact starvation #720 describes. Reaching 81 s proves the reconnect-proof `budgetHeldMs()` clock persisted across the flaps.
- **Rotation never stalled** — 6 evictions during the flap, queued brokers kept being promoted (`live` recovered to 1/1 after each block window).
- **No crash** — `boot=` constant, 0 panic/reboot markers across the run.

Router blocks removed after the test (rules cleaned = 0). Confirms the fix on hardware in addition to the host regression (`ReconnectDoesNotResetTheEvictionClock`).

Strycher:
Human approval received from Ben in session 82e07da7. Re-validated against the remote immediately prior: 1 commit, 5 files, mergeable=MERGEABLE, mergeStateStatus=CLEAN, checks=48/48 SUCCESS (incl. ci-green). Bench-repro confirmed on hv3-bench.

Issue #921, opened 2026-08-21
epic: commands reported SENT are never transmitted — the send path cannot report failure


Strycher:
**Not filed as a bug.** No single defect has been identified, and three candidate mechanisms have already been proposed and two disproved. What *is* established is a structural property of the send path that makes this class of failure invisible — and that needs investigation, not a patch.

**BACKLOG. Unclaimed. Do not start without owner direction.**

## The observation

`get tx` / `get radio` from a companion to a repeater intermittently never arrives — roughly **1 in 25**. The client reports the command as **sent**, then times out at its own constant (~28.7 s).

## What is established

`[verified: 28-minute live capture on ST-P, heltec_v4_repeater_telemetry, caplog level=packet, 2026-08-20]`

| | |
|---|---|
| CLI requests ST-P **received** | **20** |
| CLI replies ST-P **sent** | **20** |
| Unpaired (heard but unanswered) | **0** |
| Round-trip, every one | **0.605 s** (max 0.648 s) |
| SNR on every received packet | **+11 to +12 dB** |
| Noise floor | **−101 dBm**, 852 samples, stable |
| **Link margin** | **~63 dB** |
| Loop heartbeat | median **2001 ms**, max 4041 ms — never stalled |
| `packet pool empty` | **0** |

**The failed command produced no RX at all.** The repeater is healthy and answers everything it hears.

**Ruled out with evidence:** the repeater's CLI handler; firmware blocking on the loop task; packet-pool starvation; RF marginality at the receiver (a 63 dB margin on a quiet channel does not drop 1 in 25 — received packets sit in a 1 dB band, so packets arrive perfectly or not at all); and WiFi telemetry (#910's causal claim was retracted — the same failure reproduces with zero WTEL cycles).

## The structural finding — the reason this is an epic

`[verified: source, 2026-08-20]` The send path **reports success at allocation and is silent about every failure after it.**

```
CMD_SEND_TXT_MSG
  └─ sendCommandData()      returns MSG_SEND_SENT_DIRECT unconditionally
       └─ sendDirect()      void
            └─ sendPacket() void  ── drops on invalid path_len / oversized payload
                 └─ queueOutbound()  void  ── drops when the send queue is full
```

- `RESP_CODE_SENT` is emitted because a **packet object was created**, not because anything was transmitted
- The **only** client-visible failure is `createDatagram()` returning `NULL`
[...]

Issue #914, opened 2026-08-20
task: adapt wifi_telemetry_collect_and_publish() to the async lifecycle


Strycher:
**Cause:** `wifi_telemetry_collect_and_publish()` (`examples/simple_repeater/main.cpp:307-316`) is a **synchronous** connect → publish → teardown cycle that consumes `begin()`'s return value directly and maintains attempt/failure counters.

`[verified: main.cpp:308-316]`
```cpp
bool transport_up = g_tel_transport->begin();
if (!transport_up) { g_tel_wifi_fails++; g_tel_transport->end(); return; }
```

**Ask:** adapt this caller to the agreed lifecycle without losing what it does today.

## Must be preserved

- `g_tel_wifi_attempts` / `g_tel_wifi_fails` / `g_tel_publish_attempts` stay meaningful — a cycle that now spans multiple `loop()` passes must not double-count or under-count
- `getMqttState()` capture for diagnostics
- Clean teardown on failure — the current code guarantees `end()` on a failed `begin()`

## Definition of done

- [ ] Caller matches the agreed lifecycle
- [ ] Counters verified against a real cycle, not assumed
- [ ] Telemetry still publishes on a healthy cycle
- [ ] Gemini review run **on this task before it closes**

Depends on the transport lifecycle task.

Strycher:
Closed at testing. Epic #911's integration test (#915) passed both sections on hardware — results recorded there. Gemini review ran on this task before closing, per the per-task rule.

Issue #913, opened 2026-08-20
task: implement the agreed non-blocking WiFi/MQTT transport lifecycle


Strycher:
**Cause:** #910 — the WiFi/MQTT lifecycle blocks the loop task.

**Ask:** implement the option agreed in the design record.

Deliberately not specifying the approach here: the design record task decides it, and an implementation task that pre-commits to a design makes the AGREE gate decorative.

## Definition of done

- [ ] No blocking wait on the loop task in the transport lifecycle
- [ ] `TelemetryTransport` contract updated **and documented** if the agreed option changes it
- [ ] Builds green on every env carrying `ENABLE_WIFI_TELEMETRY`
- [ ] Gemini review run **on this task before it closes** — not deferred to epic end

Depends on the design record task. Sibling: the caller-adaptation task (same files may be touched — serialise per the File-Convergence Rule).

Strycher:
**Scope set: Option 2** (owner, 2026-08-20). The shared-state audit gate fired and Option 1 was abandoned — see design record §11-12, commits `befe649a` / `fef4d001`.

**Audit findings that decided it:**
- `g_tel_transport->loop()` (PubSubClient keepalive, shared `WiFiClient`) runs on `loopTask` **every iteration**; a mutex is *disqualified* because `loopTask` would block on it through a 15 s connect, recreating the defect
- OTA is fed by **AsyncTCP** while `loopTask` extends the hold deadline — Option 1 moved teardown to a third task and could kill a firmware update mid-flight

**This task now delivers:**
1. `LinkStateMachine` — a **pure**, Arduino-free state machine (inputs → next state) in its own translation unit, **host-testable**
2. `WifiMqttTransport` driven by it; the `delay()` spins in `connectWifi()`/`connectMqtt()` removed
3. Contract: `begin()` = *start or advance, true iff ready now*, plus `LinkState{Idle,Connecting,Ready,Failed}`. `isReady()` retained as `linkState()==Ready` so the five existing guards are untouched

Caller adaptation stays in #914 and now covers **both** triggers.

Strycher:
Closed at testing. Epic #911's integration test (#915) passed both sections on hardware — results recorded there. Gemini review ran on this task before closing, per the per-task rule.

Issue #911, opened 2026-08-20
epic: non-blocking WiFi telemetry lifecycle — get connect/teardown off the loop task


Strycher:
**Cause:** `WifiMqttTransport::connectWifi()` (15 s) and `connectMqtt()` (5 s) spin-wait with `delay()` **on the Arduino loop task**, so `mesh::Mesh::loop()` and `Dispatcher::checkRecv()` never run and the **LoRa receiver is unserviced** for up to ~20 s per telemetry cycle. Defect record and hardware evidence: #910.

**Ask:** get the WiFi/MQTT lifecycle off the loop task, without regressing telemetry.

## Why this is an epic and not a one-line fix

`[verified: 2026-08-20, 36 references enumerated repo-wide]` — the change reaches an abstract interface and **two callers with incompatible expectations**:

| Caller | Shape | Tolerates async? |
|---|---|---|
| `WifiTelemetry.cpp:75-79` | discards `begin()`'s return, re-checks `isReady()` | **yes** |
| `examples/simple_repeater/main.cpp:308-316` | `bool up = begin();` then publishes and tears down **in the same call** | **no** |

`[verified: WifiTelemetry.h:138-152]` — `begin()` / `end()` / `isReady()` are pure virtuals on the abstract `TelemetryTransport`, so their semantics are a contract, not a local detail.

`wifi_telemetry_collect_and_publish()` is a **synchronous connect → publish → teardown cycle** that also maintains attempt/failure counters. Any option that makes `begin()` asynchronous forces that function to be restructured.

## Options — NOT yet decided

Deliberately unresolved here. Choosing before the design record exists is how this epic gets the wrong shape.

1. **Move the whole cycle to a worker task.** Blocking is preserved but off `loopTask`; **no interface change, no caller redesign.** Follows the in-tree #53 `MqttBrokerPool` lifecycle-worker precedent. Costs: a task + stack, shared-state protection, WiFi bring-up heap spike on a second stack.
2. **Non-blocking state machine.** The blocking ceases to exist rather than relocating, and introduces no concurrency. Costs: changes the `TelemetryTransport` contract and forces `wifi_telemetry_collect_and_publish()` to be redesigned.
3. **Shrink the timeouts.** Mitigation only — narrows the deaf window, does not remove it. Not a fix.

The trade-off is real in both directions, and the first task exists to resolve it with evidence rather than preference.

## Children

[...]

PR #922, opened 2026-08-21
epic(#911): non-blocking WiFi telemetry lifecycle — 31.3s loop stall to 4.0s, proven on hardware

### LINKED ISSUE #915: task: epic #911 integration test — prove the loop no longer stalls on a FAILED WiFi connect
**Re-scoped 2026-08-20** after the causal claim in #910 was retracted. The original criterion — *"no RX gap at WTEL boundaries"* — assumed a stall that measurement showed does not occur. This issue now tests **what the change actually delivers**.

## What epic #911 actually delivers

| Delivered | Not delivered |
|---|---|
| WiFi bring-up no longer blocks the loop task (`WiFi.begin()` is async; the 15 s wait is gone) | **Any fix for the 28.7 s CLI timeouts** — those reproduce with zero WTEL cycles (#893) |
| MQTT connect bounded to ~2 s (`setSocketTimeout`), was an unbounded ~15 s default | |
| `delay(3000)` removed from the first telemetry cycle | |
| Per-cycle deadline so a flapping AP cannot stall telemetry until reboot | |
| Burst-mode teardown preserved when triggers overlap | |

**The honest problem with testing this:** on this network WiFi associates in ~1 s at RSSI −20, so the old blocking code exited almost immediately. **In normal operation the fix is invisible.** A test that only exercises the happy path proves nothing.

## The measurement to use

`noise_floor` heartbeats — nominal **2000 ms**, emitted from the loop. A blocked loop shows as a gap. This is the metric that should have been used from the start; it is what disproved the original diagnosis.

Baseline from the pre-fix capture: **245 samples, median 2001 ms, largest stall in the entire run 4.04 s.**

## A — decisive test: make WiFi FAIL

This is the only test that exercises the path the change fixes. Requires deliberately breaking WiFi association (bad PSK, or the AP taken down), then restoring — **owner-approved volatile change, not something to do casually.**

- [ ] With WiFi unable to associate, capture across **2+ telemetry cycles**
- [ ] **Loop heartbeat stays ~2000 ms** — no stall approaching 15 s
- [ ] CLI still answers at **~0.6 s** while WiFi is failing to connect
- [ ] `g_tel_wifi_fails` increments **once per cycle**, not once per loop pass
- [ ] The cycle recovers when WiFi is restored — no permanent stall (the deadline path)

**Pre-fix, this scenario stalls the loop ~15 s per cycle.** That contrast is the proof.

## B — regression checks, normal operation

- [ ] Telemetry still publishes; MQTT still connects (#914 must not have traded the fix for a broken cycle)
- [ ] **First cycle after boot shows no ~3 s heartbeat gap** — the `delay(3000)` is gone
- [ ] Burst mode still tears WiFi down when both cycles are idle — link not left up indefinitely
- [ ] Counters sane: attempts/fails/publishes consistent with cycles run
- [ ] No `packet pool empty`; free heap not materially below ~204 KB (pre-fix ~199–204 KB)

## Explicitly NOT a criterion

- ~~No RX gap at WTEL boundaries~~ — retracted. The loop was never blocked at those boundaries; the WTEL cycle costs ~0.7 s over nominal.
- ~~CLI timeouts resolved~~ — belongs to #893, which is open and unexplained. **If a 28.7 s timeout occurs during this test, it is not a failure of this epic.**

## Done

- [ ] Captures attached, before/after, with heartbeat statistics
- [ ] Result stated in terms of what was proven, and what was not
- [ ] Human sign-off — agents cannot close the epic

Strycher:
Closes #915 on merge. Epic #911, under Feature #891.

## What this delivers — and what it does not

**Delivers:** the WiFi/MQTT lifecycle no longer blocks the Arduino loop task.

**Does NOT deliver:** any fix for the intermittent CLI timeouts. Those reproduce with **zero** WiFi-telemetry cycles and are tracked separately under #921. **#910's causal claim was retracted** — the retraction is in this branch's design record, and this PR must not be read as fixing #893.

Both statements are true and neither cancels the other: the defect is real and severe under failure, *and* it was not the cause of the reported symptom.

## Hardware proof — two-arm A/B on ST-P, WiFi deliberately unreachable

SSID overridden at **build time** to a non-existent network. No device config touched, no secret read. Override verified present in each flashed binary and absent in the restore image.

Measured on the `noise_floor` loop heartbeat (nominal 2000 ms) — the metric that disproved the original diagnosis.

| | **Arm 1** pre-fix (`1dc95c43`) | **Arm 2** post-fix |
|---|---|---|
| Heartbeat samples | 179 | 191 |
| Median | 2001 ms | 2001 ms |
| **Worst loop stall** | **31,327 ms** | **4,003 ms** |
| Gaps > 6 s | 1 | **0** |
| `connectWifi: timeout` | **2** (`after 15037ms, status=1`) | **0** |
| Connect attempts in 6½ min | 2 | **9** |
| Failures actually reported | 0 | **6** |

The 4.003 ms residual matches healthy-firmware runs (4041 / 4013 ms) — normal jitter, not a remnant.

**A correction this test forced:** `status=1` is `WL_NO_SSID_AVAIL`, and the pre-fix spin does **not** exit on it — the loop only tests `WL_CONNECTED`, so every failure state reads as "keep waiting". Exposure is therefore **wider** than #910 described: any **absent** AP does it, not just one that fails to associate.

## The change

**`LinkStateMachine.h`** — a pure, Arduino-free state machine. Inputs → next state + action. Never calls WiFi, never sleeps, never reads a clock: **elapsed time is an input**, so both timeout boundaries are tested in ~2 s on the host. `WifiMqttTransport` needs the real `WiFi`/`PubSubClient` and is absent from the native `build_src_filter`, which is exactly why a 31 s stall lived in it unnoticed.

**Transport** — `begin()` advances one step and returns; both `while (…) { delay(N); }` loops are gone. `TelemetryTransport` gains `linkState()`, because a bool cannot separate "still connecting" from "failed" and `main.cpp` treats false as failure.

**Callers** — both triggers (publish and HTTP cmd-poll) are multi-pass, each with a phase so counters are touched **once per cycle**. A naive retry would have counted thousands of attempts per cycle.

**`delay(3000)` removed** from the first telemetry cycle — a 3 s loop block after every boot, in the publish path, and not in my original catalogue of this defect. Replaced with a deferred second sample. **There is now no `delay()` anywhere on the loop path**; the only two left are in `setup()`.

`setSocketTimeout(2)` set explicitly — `[verified: PubSubClient.h:34-36]` the default is 15 s and was never overridden, so one `connect()` could block ~15 s and the old loop's nominal 5 s budget could not bound it.

## Review — three passes, and what they caught

**#913, two passes.** Three CRITICALs, all real:
- **Shared phase timer** charged WiFi's elapsed time against MQTT's budget — a 6 s WiFi connect exhausted the 5 s MQTT budget before one attempt. *My test for this passed while the property did not exist*: it hand-fed the post-reset value. Replaced with a simulated caller on a monotonic clock, and **proven by re-introducing the bug** (test FAILS with it, passes without).
- **`Failed` was a permanent hang** — `end()` never reset `_link`, so after one timeout telemetry was dead until power cycle.
- **The `None` branch hammered the broker** — a blocking TCP connect on every loop pass. Retry cadence moved into the machine (500 ms, matching the old `delay(500)`).

**#914.** A flapping AP could stall telemetry until reboot: the machine times each *phase*, and every AwaitingWifi↔AwaitingMqtt transition restarts the timer, so no budget expires. Per-cycle deadline added. Also fixed a power regression I introduced — overlapping triggers meant neither tore down, leaving WiFi on indefinitely in burst mode.

**Rejected, with evidence:** a claimed `millis()` rollover failure (uint32 subtraction wraps correctly: `(10 - 4294967290) mod 2^32 == 16`); a "double-counted publish attempts" CRITICAL (the original did the same — two samples are genuinely published); and "rescheduling from completion is a regression" (the original did that too).

## Evidence

- **39/39 CI matrix** on the rebased branch — including every nRF52 and non-telemetry env that compiles the changed interface
- **19 native assertions**, one proven red against the defect
- Section B regression, 28-min real-network capture: **10 cycles / 10 teardowns** (burst power preserved), CLI **20/20 at 0.605 s**, no `pool empty`, heap ~199,948

## Open, recorded rather than smoothed over

**5 of 10 cycles reached MQTT** on the real network, alternating. The pre-fix sample was 2 cycles — **no baseline**, so this is not called a regression and is not explained. Worth its own investigation.

**Residual:** `end()`'s `WiFi.disconnect(true)` can still block; this change addresses bring-up. `isReady()` (hardware truth) and `linkState()` (machine belief) can disagree between ticks — `isReady()` remains the guard used by `publish()`.

### DIFF
--- a/docs/architecture/2026-08-20-wifi-telemetry-lifecycle-design.md
+++ b/docs/architecture/2026-08-20-wifi-telemetry-lifecycle-design.md
@@ -2,7 +2,7 @@
 
 **Issue:** #912 (design) under epic #911. Defect record: #910. Answers #893.
 **Date:** 2026-08-20
-**Status:** awaiting human AGREE. **No implementation until agreed.**
+**Status:** **AGREED 2026-08-20** — owner selected **Option 1** (lifecycle worker), with **Option 2 planned as a follow-up**.
 
 ---
 
@@ -154,6 +154,15 @@ Option 1 gets the radio listening again while changing **no contract and no call
 
 **This is a sequencing judgement, not a claim that Option 1 is the better end state.** If the concurrency audit in §7 turns up shared state that cannot be cleanly isolated, Option 2 becomes correct and this recommendation should be revisited rather than forced.
 
+### Decision — owner, 2026-08-20
+
+**Option 1 agreed for this epic. Option 2 planned as a follow-up**, tracked separately so it does not gate #911's completion.
+
+That makes Option 1 explicitly a **staging step**, not the end state. Two consequences bind the implementation:
+
+- The worker must not entrench assumptions that make Option 2 harder later. Concretely: keep the blocking confined to `WifiMqttTransport`, and do **not** spread `begin()`-is-synchronous assumptions into new code.
+- The shared-state audit (§7.1) is still blocking. It is now doubly load-bearing: it decides whether Option 1 is safe, and its findings are the main input to scoping Option 2.
+
 ---
 
 ## 7. Open risks to resolve during implementation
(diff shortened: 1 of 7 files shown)


Strycher:

Strycher:
Human approval received from Ben in session 2d2b37f1 ("Merge it on green").

- 11 commits, 7 files
- `mergeable: MERGEABLE`
- **46/47 checks SUCCESS**, 1 still running (`build (meshsmith_photon_esp32c6_e22p_30dbm_companion_radio_ble)`)

Auto-merge enabled with `--rebase`. It will land **only** once `ci-green` passes; a red matrix leaves this PR sitting unmerged rather than merging. Testing task #915 closes with the merge.

Issue #454, opened 2026-07-29
task: transport maxFrameSize()/maxFramePayload() + cap companion drains (#453)


Strycher:
Push vehicle for #453 on feat/453-transport-maxframe. Adds BaseSerialInterface::maxFrameSize() (BLE override = negotiated MTU-3) + guarded maxFramePayload(header); caps caplog/GPS/config drains against it; retires the #450 hardcode. Builds ESP32/nRF52 BLE + observer; Gemini-clean. Parent #453.

Strycher:
Landed in feat/453-transport-maxframe @ e596f99c.

Issue #450, opened 2026-07-29
bug: caplog CHUNK frame (176B) exceeds BLE ATT-deliverable payload — near-full downloads lose 3 bytes/chunk over BLE


Strycher:
Parent epic #384.

**Symptom (Ben, Android over BLE):** near-full caplog download reports the right chunk count but loses ~279 bytes. buffer 16336 → `received 16057 in 94 chunks`; every chunk arrived (94 = expected count) but each **full** chunk delivered **171 data bytes instead of 174** — 3 short. 93 full × 3 = **279** = exact deficit, reproduced twice. Serial/TCP downloads are complete (no ATT header there).

**Root cause [verified via Ben's app MTU log, QuietSnow]:** the caplog CHUNK frame is `[0xC4][0x02][174 data]` = **176 B = MAX_FRAME_SIZE** (`caplogDrain`, `cap = MAX_FRAME_SIZE-2`). Over BLE a notification carries **MTU − 3** (ATT header). The device negotiates **MTU 176** (client requests 185, device grants 176) → deliverable payload **173 B**. The device's own BLE stack clips every 176-B frame to 173, dropping 3 data bytes per full chunk. Caplog is the first feature to emit full `MAX_FRAME_SIZE` frames, so nothing exposed it before.

**Fix:** cap the caplog chunk so the CHUNK frame fits one BLE notification — `cap = MAX_FRAME_SIZE - 6 = 170` data bytes (172-B frame; ≤173 BLE with margin; consistent with the '+4 for transport codes' note on MAX_FRAME_SIZE). Works on BLE/TCP/serial; the client handles any chunk size. Scope: `caplogDrain()` in MyMesh.cpp, one line.

**Broader note (separate, not this issue):** any future full-`MAX_FRAME_SIZE` feature hits the same BLE-173 wall unless the device BLE MTU cap is raised above 176.

Builds ESP32 + nRF52; Gemini; hardware re-verify on hv4-bench-1 (BLE).

Strycher:
[...]

PR #937, opened 2026-08-22
bug(#711): derive the BLE frame ceiling from our own preference, not a reported MTU

### LINKED ISSUE #711: bug: REGRESSION of #450 — BLE CHUNK frames lose 3 bytes each again; #454 traded the safe constant for a cap that trusts a stale MTU

Strycher:
## What this changes

Re-fixes #711. **The first fix shipped in beta2/3/4 and does not work** — beta4 still clips 3 bytes off every full BLE frame. Affects every BLE companion role, ESP32 and nRF52.

Addresses #711 — **stays OPEN until a bench V4.3 confirms it.**

### The field evidence

Tester `schill`, Heltec **V4.3 (ESP32-S3)**, on **beta4, which contains the first fix**:

```
Capture truncated: received 12751 of 12973 bytes in 75 chunks
Buffer 12973 / 16384 bytes
```

`shortfall 222 = 3 × 74 full chunks` — the same signature as the original report (`147 = 3 × 49`).

12973 bytes needs **76** chunks at a 171-byte payload and **75** at 174. He saw 75, and `74 × 174 + 97 = 12973` exactly. So `maxFrameSize()` returned **176 — the pre-fix value — on a build carrying the fix.**

### Why the first fix was a no-op

`[verified: NimBLE source]` `NimBLEDevice::getMTU()` returns `ble_att_preferred_mtu_val`, whose **NimBLE default is 256** (`MYNEWT_VAL_BLE_ATT_PREFERRED_MTU`, set in `ble_att_init()` ← `ble_hs_init()`). Where our `setMTU(176)` did not take effect:

```
effectiveMtu(517, 256)     -> 256      (min, as designed)
deliverableFrame(256, 176) -> 176      (min(253, 176))
maxFramePayload(2)         -> 174      <- exactly what shipped
```

**The clamp was a no-op for any local value >= 179.** I made the fix depend on a value I never verified — which is the same trade #454 made, and the reason this issue exists.

### The fix

`deliverableFrame()` now returns `min(effective_mtu - 3, max_frame - 3)`. The second term is a **hard ceiling derived from our own pinned preference and from nothing the stack reports**. Clamping to `max_frame` — the *buffer* size — was always one ATT header too generous.

This restores the property #450's constant had and every computed replacement since has lost: **safe by construction.**

- **nRF52 carve-out REMOVED.** Covered by the same ceiling, with no invented local MTU. Leaving it unclamped was wrong: it is a whole board class (RAK4631, RAK3401, T1000-E, Wio Tracker, T114, T096, XIAO nRF52) that may have been unable to deliver a caplog at all, and unlike ESP32 nobody had a tester capture that would have surfaced it. The #718 Gemini review called it a landmine; that judgement did not survive contact.
- **`setMTU()`'s bool return is checked and logged.** It was discarded, so a refusal was invisible — inferable only from a tester's truncated download.

**Cost:** at most 3 bytes per frame on links that could carry more (nRF52 at MTU 247 now caps at 173 rather than 176). 1.7% throughput for correctness on every board and every stack, including ones that misreport.

## Tests

**Shown failing first: 4 of 13 cases fail against the pre-re-fix code, 12/12 pass with it.**

Three new cases exercise `local = 256`, every `(reported, local)` pair, and schill's exact arithmetic (76 chunks at the correct cap, 75 and −222 at the broken one).

**Why the existing suite went green on a fix that did not work:** every case pinned `local` at 176 and **never exercised the value that breaks it**. That is the #712 lesson landing on its own author. All cases now sweep `local` as well as `reported`.

One pre-existing test was **renamed and corrected, not deleted**: `SerialAndTcpAreUnaffected` asserted `deliverableFrame()` returns the *full frame* for a generous MTU — the defect written down as an expectation — and did not test its own name. `[verified]` `deliverableFrame()` is called only from the two BLE interfaces; serial/TCP use the `BaseSerialInterface` default and never reach it.

## Verification

- [x] `pio test` green for every native env: **266 / 8 / 7**
- [x] Builds: `heltec_v4_companion_radio_ble`, `RAK_4631_companion_radio_ble` (nRF52 path), `Heltec_v2_companion_radio_ble` (DRAM-tight, not in the CI matrix)
- [ ] **NOT hardware-verified. Owner has directed that a bench V4.3 (`hv4-bench-1`) tests this before beta5 is cut.** Pass = announced == received on a caplog download.

## Review

- [x] Gemini-reviewed (`gemini-2.5-pro`) — verdict "ship it".

Its Q6 finding is adopted as a **documented limit, not a claimed fix**: the ceiling stops an MTU over-estimate *above* `max_frame`, but not a medium one. A stack reporting 247 on a link that really negotiated 100 would still clip. That cannot be closed from reported values alone — it needs an integrity signal from the wire. Recorded in the commit and here rather than left implied by "ship it".

It also argued the corrected test was justified rather than self-serving, having been asked explicitly to argue the other side.

## Notes

**Third attempt at this invariant** (#450 → #454 → #711 → here). The pattern each time has been replacing a value that was safe by construction with one that is safe only if something reported is truthful. This change goes the other way.

**On-device discriminator, already shipped in beta2+:** a diag build prints
`setEffectiveMtu(): reported=<n> local=<n> -> effective=<n> (deliverable frame=<n>)`.
`local=256` on a V4.3 confirms the diagnosis above outright.

### DIFF
--- a/src/helpers/BleFrameSizing.h
+++ b/src/helpers/BleFrameSizing.h
@@ -50,10 +50,34 @@ inline uint16_t effectiveMtu(uint16_t reported, uint16_t local) {
 /// raise the result above what the link carries, which is what the previous
 /// `min(mtu - 3, MAX_FRAME_SIZE)` form did once the cached MTU was too large.
 inline size_t deliverableFrame(uint16_t effective_mtu, size_t max_frame) {
-  size_t payload = effective_mtu > ATT_NOTIFY_HEADER
-                     ? (size_t)(effective_mtu - ATT_NOTIFY_HEADER)
-                     : 0;
-  return payload < max_frame ? payload : max_frame;
+  // What the link says it can carry, from the (possibly untrustworthy) MTU.
+  const size_t link = effective_mtu > ATT_NOTIFY_HEADER
+                        ? (size_t)(effective_mtu - ATT_NOTIFY_HEADER)
+                        : 0;
+
+  // #711 RE-FIX: the HARD ceiling, derived from OUR OWN pinned preference and from
+  // nothing the stack reported. begin() pins the local preferred MTU to max_frame,
+  // so a link at that preference delivers at most max_frame - 3 -- clamping to
+  // max_frame (the BUFFER size) was always one ATT header too generous.
+  //
+  // The first fix computed min(reported_mtu - 3, max_frame) and was a no-op whenever
+  // the reported value was >= max_frame + 3. It shipped in beta4 and STILL clipped:
+  // NimBLEDevice::getMTU() returns ble_att_preferred_mtu_val, whose NimBLE default is
+  // 256, so a device where our setMTU() did not take effect computed
+  // min(253, 176) = 176 -- exactly the pre-fix size. Tester schill, Heltec V4.3:
+  // 12751 of 12973 bytes in 75 chunks, shortfall 222 = 3 x 74 full chunks.
+  //
+  // Taking the min of both bounds is safe BY CONSTRUCTION, which is the property
+  // #450's hardcoded constant had and every computed replacement since has lost:
+  //   - setMTU() took effect  -> link is the real bound, and own == link. Exact.
+  //   - setMTU() did NOT take effect -> we may under-use a larger link by 3 bytes.
+  //     Conservative, never clipped. A 1.7% throughput cost buys correctness on
+  //     every board and every stack, including ones that misreport.
+  const size_t own = max_frame > ATT_NOTIFY_HEADER
+                       ? max_frame - ATT_NOTIFY_HEADER
+                       : 0;
+
+  return link < own ? link : own;
 }
 
 }  // namespace ble_frame
--- a/src/helpers/esp32/SerialBLEInterface.cpp
+++ b/src/helpers/esp32/SerialBLEInterface.cpp
@@ -25,7 +25,18 @@ void SerialBLEInterface::begin(const char* prefix, char* name, uint32_t pin_code
 
   // Create the BLE Device
   NimBLEDevice::init(dev_name);
-  NimBLEDevice::setMTU(MAX_FRAME_SIZE);
+  // #711: setMTU() returns false if the stack refused it, and the return was
+  // previously discarded -- a silent failure (SAFELANE §6). It matters because
+  // getMTU() then reports NimBLE's DEFAULT preferred MTU (256), not ours, and any
+  // sizing that trusts it will oversize. The frame ceiling no longer depends on
+  // this succeeding (see BleFrameSizing::deliverableFrame), but a failure here is
+  // still a real degradation and must be visible rather than inferred from a
+  // tester's truncated download.
+  if (!NimBLEDevice::setMTU(MAX_FRAME_SIZE)) {
+    BLE_DEBUG_PRINTLN("begin(): setMTU(%d) REFUSED -- getMTU() will report the "
+                      "NimBLE default (%d), not ours", (int)MAX_FRAME_SIZE,
+                      (int)NimBLEDevice::getMTU());
+  }
 
   // Security/pairing init. Bluedroid's setSecurityCallbacks(this) + BLESecurity
   // setStaticPIN/setAuthenticationMode combo becomes three NimBLE static calls
--- a/src/helpers/nrf52/SerialBLEInterface.cpp
+++ b/src/helpers/nrf52/SerialBLEInterface.cpp
@@ -283,14 +283,19 @@ size_t SerialBLEInterface::maxFrameSize() const {
   // divergent copy is what let the ESP32 side regress unnoticed. The shared function is
   // unit-tested in test/test_frame_size.
   //
-  // NOTE the asymmetry, which is intentional: ESP32 must clamp with
-  // ble_frame::effectiveMtu() because NimBLE reports the PEER's MTU and begin() pins our
-  // local side via NimBLEDevice::setMTU(). Here, Bluefruit's conn->getMtu() is the
-  // CONNECTION's negotiated MTU (already min(local, peer)), and this port sets no explicit
-  // local MTU, so there is no second value to clamp against. Do NOT invent one: passing a
-  // guessed local MTU into effectiveMtu() would silently shrink frames on healthy links.
-  // If conn->getMtu() is ever found to report a peer-preferred value instead, THAT is the
-  // bug to fix, and it should be fixed by clamping against a real configured local MTU.
+  // #711 RE-FIX: this path is now protected by the same hard ceiling as ESP32.
+  // deliverableFrame() caps at MAX_FRAME_SIZE - 3 regardless of what getMtu() reports,
+  // so a stack that over-reports here can no longer produce a clipped frame.
+  //
+  // The previous revision left this path deliberately unclamped, reasoning that
+  // Bluefruit's conn->getMtu() is already the negotiated value and that inventing a
+  // local MTU would shrink frames on healthy links. The Gemini review of #718 called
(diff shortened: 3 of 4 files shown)


Strycher:
Owner approved merge (2026-08-22). 48/48 green including `native-tests`, which ran this PR's three new regression cases (`RegressionSevenElevenB_*`) — the ones proven to fail 4/13 against the pre-re-fix code.

Branch is 0 behind base, `mergeState=CLEAN`. Enabling auto-merge.

**Still NOT hardware-verified.** Owner has directed that bench V4.3 (`hv4-bench-1`) tests this before beta5 is cut, so #711 stays OPEN on merge.

Issue #668, opened 2026-08-13
bug: nRF52 BLE companions crash-boot loop on the 1.17.0 base (RAK4631 AND T1000-E) — USB companions on the same boards are fine


Strycher:
Parent: Epic #614 · Related: #628 (1.17.0 integration), #663 (migration matrix — nRF52 column), #350 (CrashLog)

# RAK4631 BLE companion crash-boot loops on the 1.17.0 base, ~4.19s in, immediately after radio init

Found while attempting the nRF52 leg of the #663 migration matrix. **This blocks the entire nRF52
column** — the device cannot stay up long enough to verify anything.

## Reproduction

`[verified]` Device `WSMJ898-RAK4631` (RAK4631, nRF52840), DFU-flashed with
`RAK_4631_companion_radio_ble` built from the #628 integration branch (`-D OFFBAND_FORCE_CAPLOG`),
artifact sha `5562a02a`. `Device programmed.` reported by adafruit-nrfutil.

**6 of 6 boot cycles crash at the same point.** Captured with a reconnecting serial capture across
resets:

```
---- attach #3 ----
[4193] DEBUG: Set _preambleMillis=91
[4193] DEBUG: Set _maxPayloadMillis=1213
[4194] DEBUG: RX Boosted Gain Mode: Enabled
** port dropped (device reset) **
```

Timing is tight and consistent across cycles: 4193–4195 ms. The last line is always the end of
radio initialisation.

## What was running before this

`[verified]` Prior firmware `offband-v1.3.0-5-g75c40f8` (MeshCore v1.16.0 base) ran fine on this
board — it is the owner's working companion. Backup at
`flash-backups/rak4631-DFU-backup-20260812-224907/` (`CURRENT.UF2`, sha `9a8dadae`), which covers
SoftDevice + application (`0x1000`..`0xEA000`, verified from the UF2 block headers).

So this is a **1.17.0-base regression on this board**, not a pre-existing fault.

## Why there is no crash dump yet

`[verified]` USB CDC on this board does not enumerate until ~4.1 s. The crash is at ~4.19 s, so a
host has roughly **0.1 s** of visibility. Everything earlier — CrashLog's recovery dump, the
`[prefs]` migration line, boot diagnostics — is emitted before the host can attach and is simply
not seen on the live mirror.

The capture ring holds it all, but this is a **BLE** companion: the console is not shared with the
framed protocol, so there is no serial path to `caplog download`, and BLE never comes up because
[...]

PR #939, opened 2026-08-23
bug(#711): MultiSerialInterface never overrode maxFrameSize — delegate to it

Strycher:
## What this changes

**This is the actual defect.** #450, #454, and both previous #711 attempts all modified code that **is not reached on the companion path**.

`MultiSerialInterface` overrides nine `BaseSerialInterface` methods — `enable`, `disable`, `isEnabled`, `isConnected`, `loop`, `isWriteBusy`, `writeFrame`, `checkRecvFrame` — and **not `maxFrameSize()`**.

`companion_radio` holds a `MultiSerialInterface` as its `_serial` (`main.cpp:49`, `interface_manager`). So `caplogDrain()`'s `maxFramePayload(2)` returned `BaseSerialInterface`'s default `MAX_FRAME_SIZE - 2 = 174` on every board and every link, and the BLE interfaces' MTU-aware `maxFrameSize()` was never consulted at all.

Introduced by the 1.17.0 MultiSerialInterface port (#668) — `main.cpp:382` already records that the legacy single-`serial_interface` wiring was not removed in that port.

Addresses #711 — **stays OPEN until `hv4-bench-1` re-tests.**

### The evidence that settled it

Bench client debug log:

```
20:55:13  MTU set to: 176            <- negotiated AND reported honestly
21:03:34  14246 of 14495 in 84 chunks
measured chunk payloads: 83 x 171 received + 1 x 53
```

MTU 176 delivers 173. The firmware wrote 176-byte frames. **3 bytes lost per full frame.** All three field reports share that signature exactly:

| | | |
|---|---|---|
| madmax_2069 | 8461 of 8608, 50 chunks | 147 = 3 × 49 |
| schill | 12751 of 12973, 75 chunks | 222 = 3 × 74 |
| hv4-bench-1 | 14246 of 14495, 84 chunks | 249 = 3 × 83 |

### The fix

`maxFrameSize()` returns the **minimum across interfaces where `instance && instance->isEnabled()`** — the same predicate `writeFrame()` uses to fan a frame out, so the two cannot disagree. Nothing enabled → `MAX_FRAME_SIZE`, since `writeFrame` sends nothing in that state and returning 0 would surprise callers for no gain.

### Hard ceiling reverted (owner-directed)

The previous attempt made `deliverableFrame()` return `min(mtu-3, max_frame-3)` on the theory that a stack might over-report its MTU. **That theory is refuted** — the MTU was reported accurately — and the ceiling cost 3 bytes/frame on nRF52 (MTU 247, delivers 244) for no benefit.

`min(link, max_frame)` is correct on both: ESP32 at MTU 176 → 173, nRF52 at 247 → 176. The `setMTU()` return check is **kept** — an independent silent-failure fix.

### A correction to the record

I previously claimed the nRF52 board class "may have been unable to deliver a caplog at all." **That was wrong.** A 176-byte frame fits a 244-byte deliverable, so nRF52 never clipped — which is precisely why every field report is ESP32. The defect is only fatal where the deliverable is **below** `MAX_FRAME_SIZE`, i.e. the ESP32 case, because `begin()` pins our preferred MTU to `MAX_FRAME_SIZE`.

## Tests

**`test/test_multiserial` drives the real `MultiSerialInterface`** with fake transports and is **verified to FAIL 3/6 with the override removed.**

It pins the exact field case (an interface reporting 173 must yield payload **171**, not 174), the minimum-across-interfaces rule, the disabled-interface predicate, and that a generous interface cannot exceed the frame buffer.

The frame-size suite is updated to the reverted invariant and now asserts the field chunk counts directly: **8608 → 51, 12973 → 76, 14495 → 85**, nothing lost.

One of my own test premises was wrong and **the test was corrected, not the code**: `MultiSerialInterface::enable()` enables *every* registered interface, so a fake constructed disabled comes back up; the realistic state is disabled *after* `enable()`.

## Verification

- [x] `pio test` green for every native env: **271 / 8 / 7**
- [x] Builds: `heltec_v4_companion_radio_ble`, `RAK_4631_companion_radio_ble`, `Heltec_v2_companion_radio_ble`, `heltec_rc32_companion_radio_usb`
- [ ] **NOT device-verified.** `hv4-bench-1` re-test gates this. Expect **85 chunks at 171 payload** for a 14495-byte buffer, announced == received.

## Review

- [x] Gemini-reviewed (`gemini-2.5-pro`) — "approve with comments".

Three of its points, handled explicitly:

- It argued to **keep** the hard ceiling as defence-in-depth ("this part of the code is cursed"). Recorded; the owner ruled to revert, and that is not re-litigated here.
- It called the wrapper's missing `isConsoleSharedWithProtocol()` override a **critical live bug**. **Checked: it is not.** `main.cpp:491` calls it on `usb_serial_interface`, the concrete `ArduinoSerialInterface`, which *does* override it — never on the wrapper. Latent hazard, recorded against #871 rather than widened into this change.
- Its note on the pre-existing by-value `for(auto iface : _interfaces)` loops is correct; left as out of scope, new code uses `const auto&`.

## Notes

**Fourth attempt at this invariant.** The first three each changed a value's *derivation* without checking whether the function computing it was *called*. The test added here exercises the wrapper that was skipped, which is the gap all three previous test suites shared — they tested the leaf, and the leaf was never reached.

### DIFF
--- a/src/helpers/BleFrameSizing.h
+++ b/src/helpers/BleFrameSizing.h
@@ -50,34 +50,24 @@ inline uint16_t effectiveMtu(uint16_t reported, uint16_t local) {
 /// raise the result above what the link carries, which is what the previous
 /// `min(mtu - 3, MAX_FRAME_SIZE)` form did once the cached MTU was too large.
 inline size_t deliverableFrame(uint16_t effective_mtu, size_t max_frame) {
-  // What the link says it can carry, from the (possibly untrustworthy) MTU.
+  // Bound 1: what one ATT notification carries on this link.
   const size_t link = effective_mtu > ATT_NOTIFY_HEADER
                         ? (size_t)(effective_mtu - ATT_NOTIFY_HEADER)
                         : 0;
-
-  // #711 RE-FIX: the HARD ceiling, derived from OUR OWN pinned preference and from
-  // nothing the stack reported. begin() pins the local preferred MTU to max_frame,
-  // so a link at that preference delivers at most max_frame - 3 -- clamping to
-  // max_frame (the BUFFER size) was always one ATT header too generous.
+  // Bound 2: the frame BUFFER (uint8_t buf[MAX_FRAME_SIZE]) -- a link that can
+  // carry more cannot be exploited without changing MAX_FRAME_SIZE itself, which
+  // is a protocol constant the client shares. An nRF52 at MTU 247 delivers 244 and
+  // still sends 176-byte frames for this reason.
   //
-  // The first fix computed min(reported_mtu - 3, max_frame) and was a no-op whenever
-  // the reported value was >= max_frame + 3. It shipped in beta4 and STILL clipped:
-  // NimBLEDevice::getMTU() returns ble_att_preferred_mtu_val, whose NimBLE default is
-  // 256, so a device where our setMTU() did not take effect computed
-  // min(253, 176) = 176 -- exactly the pre-fix size. Tester schill, Heltec V4.3:
-  // 12751 of 12973 bytes in 75 chunks, shortfall 222 = 3 x 74 full chunks.
-  //
-  // Taking the min of both bounds is safe BY CONSTRUCTION, which is the property
-  // #450's hardcoded constant had and every computed replacement since has lost:
-  //   - setMTU() took effect  -> link is the real bound, and own == link. Exact.
-  //   - setMTU() did NOT take effect -> we may under-use a larger link by 3 bytes.
-  //     Conservative, never clipped. A 1.7% throughput cost buys correctness on
-  //     every board and every stack, including ones that misreport.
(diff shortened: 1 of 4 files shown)


Strycher:
Owner approved merge (2026-08-22), **after a hardware pass** — the first in four attempts at this invariant.

```
[INFO/Caplog] Caplog download complete: 16381 bytes in 96 chunks
```

`hv4-bench-1` (Heltec V4.3), running `offband-v1.5.0-beta4-2-ge259b12`, near-full 16 KB buffer. **complete**, not truncated.

The chunk count is the independent proof: 16381 bytes needs **96** chunks at a 171-byte payload and **95** at 174. 96 observed → 171-byte payloads → the delegation is live on the real path.

Same signature and cause as madmax_2069 (8461/8608) and schill (12751/12973), so this covers both.

48/48 CI green. Enabling auto-merge.

Issue #938, opened 2026-08-22
chore: Feather sniffer listens on both UART header positions — RC52 reverses RX/TX


Strycher:
The Feather sniffer rig listens on one pin, and Heltec reversed the UART header positions on the RC52. Swapping the harness between boards currently means resoldering.

## The problem

| Board | header pin 11 | header pin 12 |
|---|---|---|
| RC32 / RCC6 | board **RX** | board **TX** |
| **RC52** | board **TX** | board **RX** |

The rig's single listen wire is soldered to the pin-12 position, which is correct for RC32/RCC6 and lands on **RX↔RX** on the RC52 — no data, and it looks exactly like a dead board.

## Fix — listen on both positions at once, no switch

The owner has no dipswitch to put in the chain, so the sketch stops choosing and listens to both.

- The Feather is an **Adafruit Feather ESP32-S3**, which has a GPIO matrix — any pin can carry a UART RX, and the core's `TX` name is just an alias for GPIO39. `[verified: framework-arduinoespressif32/variants/adafruit_feather_esp32s3/pins_arduino.h:30 — static const uint8_t TX = 39]`
- The sketch's **TX pad is already unused** — its own header says *"sniffer TX → NOT CONNECTED (either board)."*
- So: second jumper from the other header position to the Feather's TX pad, second UART opened on it, both drained into the same stream.

Whichever board is attached, one wire carries data and the other sits on that board's RX. Both stay soldered permanently.

### Why it is electrically safe in both orientations

Both Feather pins are **inputs**, and the pin each one meets on the idle side is the target's **RX**, also an input. Two inputs on a net have nothing to fight.

⚠ **This holds only while both UARTs are opened with `TX = -1`.** Giving GPIO39 a real TX role would drive into a target's TX output on one of the two boards. It is a one-character mistake with hardware consequences, and the code says so at the call site.

### Which wire is live, visibly

Separate byte counters per pin, both in the heartbeat. A glance at the capture says which header position the attached board transmits on — no inference.

## Scope

- `#define` for the alternate RX pin, defaulting to the Feather's TX pad.
- Second UART opened RX-only; both relayed to USB in the same loop, relay-first as now.
- Per-pin `rx_bytes` counters surfaced in the heartbeat.
- Build ID `SNIFFER-v3` → `SNIFFER-v4`. Nothing parses that string — `capture.py` only mentions it in help text `[verified: grep, 2026-08-22]` — but it must change so a capture says which sketch produced it.
[...]

Issue #857, opened 2026-08-19
task: RC52 battery runtime burn — measured discharge to auto-shutdown


Strycher:
Measure RC52 runtime on battery, from full charge to firmware auto-shutdown. Third board in the RadioCore series, directly comparable to #833 (RC32, **21 h 34 m**) and #903 (RCC6).

## Scope correction

This issue previously asked for a **multimeter check of the ADC polarity and multiplier before the run**. That was invented work and is removed. It was wrong on three counts:

1. **The constants are not in doubt.** `R17 390K / R18 100K` → **4.90**, and `ADC_CTRL_ENABLED HIGH`, from this board's own schematic *and* Heltec's own BSP (`BAT_AMPLIFY 4.9F`). Two authoritative sources for this board, not a sibling.
2. **The same divider is already validated on two boards.** #833's cross-check: *"A consistent ~28 mV offset with no drift over 21 hours … independently validates the RC32 sense chain — R36 390k / R38 100k, multiplier 4.9 — to under 1% across a 700 mV span."* Same values. The RCC6 run confirms it a third time (~27 mV, matching RC32's 27.7).
3. **The run verifies both for free, and better.** A plausible non-zero `[pwr]` reading *is* the polarity proof — wrong-polarity gating reads **zero** (the RC32 #766 case exactly). And `[pwr]` vs `[gauge]` across a full discharge validates the multiplier across the whole span, which a single meter reading cannot.

The earlier justification cited #602. That failure was inheriting polarity **from a sibling variant without checking**. Reading this board's own schematic and its own vendor header is not that.

## Prerequisites — both of these were gating on the RCC6 and cost time mid-run

**1. `AUTO_SHUTDOWN_MILLIVOLTS=3400` on the RC52 BLE companion env.**
Not a nicety — the shutdown block sits inside `#ifdef AUTO_SHUTDOWN_MILLIVOLTS`, so with the macro undefined the path **compiles out entirely** and the run has *no endpoint*. It would discharge to brownout instead of shutting down, which is both a different measurement and worse for the cell. Scope it to the BLE companion env only, as `heltec_rc32` and `heltec_rcc6` both do.

**2. Board-side battery telemetry — the nRF52 equivalent of `OFFBAND_POWER_TELEMETRY`.**
[...]

Issue #833, opened 2026-08-19
task: RC32 battery runtime characterisation — measured discharge to auto-shutdown (Heltec beta Q01)


Strycher:
Measured runtime of an RC32 companion on battery, from full charge to firmware auto-shutdown.
Answers **Heltec RadioCore beta survey Q01** (power consumption / runtime), and establishes the
baseline for the role and display comparisons that follow.

## Why a measured run rather than an estimate

The board reports **voltage only** -- no fuel gauge, no current sense. Converting voltage to
runtime requires a discharge curve nobody had for this cell, and the two estimation methods in
`tools/diag/rc32-boot-740/scripts/battery_runtime.py` disagreed by as much as **seven hours**
mid-run. Every projection made during the run **undershot the measured result**, several by
hours. Only running to cutoff produces a number that owes nothing to a model.

## Device under test

| | |
|---|---|
| Board | `rc32-bench-1` — Heltec RadioCore **RC32-L62** (ESP32-S3 + HT-RA62A) |
| Firmware | `heltec_rc32_companion_radio_ble_diag` |
| Build | `offband-v1.5.0-beta2-26-g0a2373a` (post-#704, post-#769) |
| Role | BLE companion, display active, GPS off, `DISPLAY_ROTATION=0` |
| Cutoff | `AUTO_SHUTDOWN_MILLIVOLTS=3400` |
| Cell | 2500 mAh single-cell LiPo, wired to VBAT (header pin 2 / GND pin 20) |

## Instrumentation

Deliberately **external** -- the RC32's own USB console is generated by the chip, dies with the
chip, and power-cycles the board on attach (#702). Nothing about a battery run can be trusted if
observing it changes it.

- **UART0 sniffer** — Adafruit Feather ESP32-S3 (`feather-sniffer`, COM16) on RC32 pin 12 (U0TXD)
  and pin 20 (GND), running `rc32_uart_sniffer_v3`.
- **MAX17048 fuel gauge** on the sniffer's `Wire1` (0x36), wired between cell and RC32 —
  independent modelled charge, and an independent voltage cross-check.
- **Host capture** — `capture.py`, detached, appending with UTC timestamps.
- **Device telemetry** — `OFFBAND_POWER_TELEMETRY` emits `[pwr] mv=… up=…s` every 30 s via
  `mesh_log_line()`. Since #769 the UART0 mirror emits **regardless of caplog state**, so no CLI
  setup is needed.

## Procedure

[...]

Issue #740, opened 2026-08-16
task: instrument the RC32 boot path end-to-end -- verbose bootloader + raw-UART0 phase beacon + Feather remote reset


Strycher:
Parent: #702. Instrument work only — this task adds visibility, it does not attempt a fix.

## Why

Three stages of the RC32 boot have never been observed, and #702's four failed fixes all targeted
code downstream of them. Established in
[#702 comment](https://github.com/OffbandMesh/meshcore-firmware/issues/702#issuecomment-5306695089):

- `entry 0x403c98d0` is the **second-stage bootloader's** entry point (`readelf` on
  `bootloader_qio_80m.elf`), not the application's.
- That bootloader is built `CONFIG_BOOTLOADER_LOG_LEVEL_NONE=y` — it prints nothing on success or
  failure, so the existing capture cannot distinguish "bootloader hung" from "app hung".
- `CONFIG_LOG_MAXIMUM_LEVEL=1` in the prebuilt arduino-esp32 libs compiles out every IDF
  `INFO`/`DEBUG` line; only `ERROR` survives, and it cannot be raised at runtime.

Owner directive (2026-08-16): robust logging at every step, no blind spots, before any further
hypothesis work. Beta 1 covers four RadioCore boards (RC32, RCC6, RC52/nRF52, HaLow) — RC32 is the
first, so this instrument is built to carry to the other three.

## Deliverables

**1. `OFFBAND_BOOT_BEACON` — app-side raw UART0 phase markers**
Direct UART0 FIFO register writes. No Arduino, no FreeRTOS, no logging framework, so it works from
the first executable instruction of the app.
- `__attribute__((constructor(101)))` → `APP:CTOR`. This single marker exonerates or indicts the
  bootloader + IDF startup in one RST press.
- Markers at top of `setup()`, after `board.begin()`, and at every existing `CW_PHASE` through
  `display.begin()` to screen-up.
- MeshLog/`CW_PHASE` routed to **UART0** for this build, not USB-CDC (USB-CDC dies with the chip,
  so it can never witness a failure).
- `ENV_INCLUDE_GPS=0` retained so `Serial1.setPins(44, 43)` cannot steal GPIO43 — U0TXD, the
  sniffer wire — partway through boot.

**2. `tools/diag/bootloader-verbose/` — IDF 4.4.7 bootloader at `LOG_LEVEL_DEBUG`**
[...]

PR #942, opened 2026-08-23
chore(#938): sniffer listens on both UART header positions, and reads the INA219

Strycher:
Makes the Feather bench sniffer work on both UART header positions, and turns the INA current monitor from a thing it names into a thing it reads.

## Why

Heltec **reversed the UART header positions** between board revisions:

| | header pin 11 | header pin 12 |
|---|---|---|
| RC32 / RCC6 | board **RX** | board **TX** |
| **RC52** | board **TX** | board **RX** |

A rig soldered for one lands RX-to-RX on the other and receives nothing — indistinguishable from a dead board. Swapping meant resoldering, and there is no switch in the chain to select with.

## What changed

- **Relocated** to `tools/diag/sniffer/rc_uart_sniffer_v3/` — board-neutral, since it sniffs RC32, RCC6 and RC52. The old `rc32-boot-740/` folder stays: `scripts/capture.py` lives there and is in active use.
- **`SNIFFER_RC52` selects the listening pad** — the owner's design, kept as written. The pad labelled `TX` receives because the ESP32-S3 routes UART through the GPIO matrix and direction follows the argument `begin()` gets.
- **INA identification fixed** — see below.
- **INA219 read path** — current, bus voltage, and a cross-check against the fuel gauge.
- **Wiring header rewritten** to match the code.

## Hardware verification — this was not built and hoped for

Captured on `feather-sniffer` against a live RC52. Full record: #938.

**RX on the TX pad:**
```
=== listening on GPIO39, 115200 8N1
[hb] SNIFFER-v3 alive  n=3  up=5s  rx_bytes=53
```
416 KB received across one capture — 6 513 `noise_floor` lines, 656 dispatcher lines, **54 valid adverts**.

**INA219 read, on battery:**
```
03:00:54  ma=-0.2   raw_shunt=0xFFFE     USB connected — noise floor
03:01:24  ma=14.0   raw_shunt=0x8C       USB pulled — running on battery
```
**37 on-battery samples: 14.0–14.6 mA, mean 14.25, σ 0.14 mA (1.0%)** against a 0.7% quantisation floor.

**Decode validated against an independent instrument** — INA bus 4156 mV vs MAX17048 cell 4157 mV. Different chips, different registers, different maths, 1 mV apart. Signed path proven by a real `0xFFFE` → −0.2 mA sample.

**Same-board proof:** the target's millis counter never reset across the whole capture (continuous to 21,504,731 ms) and the stream ran uninterrupted through the USB removal that produced 14 mA on the shunt. A USB-powered board dies when USB is pulled; this one kept running on the measured battery.

## The decode that would have been silently wrong

A datasheet summary reported the shunt register as "bits 15-3", the same shift as the bus register. **Rejected**, by computing from the config register's own reset value rather than reading the annotation: `0x399F` has `PG = 0b11` → PGA ÷8 → ±320 mV, which at 10 µV/LSB needs ±32000 counts; a 13-bit field holds ±4096. So the shunt is a **full signed 16-bit value, unshifted**.

Shifting it would have reported **one eighth of the real current** — in range, plausible, and caught by nothing. The reviewer independently confirmed the reasoning.

## Gemini adversarial gate — run twice

`scripts/llm-consult.py --backend gemini --model gemini-2.5-pro`. Audit logs: `docs/llm-consultations/2026-08-22-938-sniffer-ina-*` and `-938-sniffer-rereview-*`.

**First pass — 7 findings. 5 fixed:**

| Finding | Fix |
|---|---|
| Any non-TI device at 0x40–0x4F was bucketed as INA219 and then **read** as one — fictitious current from an unrelated chip | Identification is now **positive**: config register must read its documented reset value `0x399F`. Anything else prints `UNIDENTIFIED -- not read` |
| `ina_addr` latched from the first device that merely ACKed, before any identification | Set only inside branches that positively identify a part |
| A stalled I2C device could silently cost captured UART bytes — 256-byte buffer fills in ~22 ms, Wire timeouts are 100 ms+ | `Serial1.setRxBufferSize(4096)` before `begin()` |
| `ma_x10 = -ma_x10` is undefined for `INT32_MIN` | Negation through an unsigned magnitude |
| Cross-check comment named the wrong comparison | Corrected |

**1 justified, not fixed.** The reviewer rated using the `TX`-labelled pad as an RX hardware-damaging, on the grounds a future edit could re-enable TX or someone could mis-wire it. The exposure is real but is not created by this change — any clip onto a target header carries it, and the existing RX-pad wire has the identical exposure on the other revision. The alternative means re-wiring hardware the owner built and flashed deliberately. **Surfaced to him rather than silently overridden.** Both `begin()` calls still pass `-1` for TX and say why at the call site.

**1 withdrawn** by the reviewer inside its own report.

**Re-review — and it earned its keep.** It flagged that the wiring header contradicted the code: still said *"sniffer TX → NOT CONNECTED (either board)"* while the code listens on that pad. **Mine** — I rewrote that block for an earlier design, then reset the branch when the sketch moved to the owner's copy, and the correction went with it. Fixed.

It reached that by a wrong route — it asserted the Feather's `RX` is GPIO44. It is not: all three plain variants define `RX = 38`, `TX = 39`, and the bench banner reads `listening on GPIO39`. **The pin claim is refuted; the defect it led to was real.**

**And it surfaced a trap neither of us had:** `TX`/`RX` are **variant-dependent aliases**. Plain Feather ESP32-S3, `_nopsram` and `_reversetft` → GPIO39/38; `_tft` → **GPIO1/GPIO2**. Wrong board in the IDE silently moves the listening pin. The header now points at the boot banner's resolved GPIO as ground truth.

⚠ **The re-review response was truncated** (226 output tokens, cut mid-sentence) after one run failed outright with a `KeyError` in the shared LLM client. Recorded so the second pass is not read as more complete than it was.

## Not verified

- **Compilation by CI or by an agent.** This is an Arduino IDE sketch — its own header records two PlatformIO attempts that flashed cleanly and printed nothing. It **was** compiled and flashed by the owner, and the new output appearing in the capture is the proof. Checked mechanically here instead: 8 `#if` / 8 `#endif`, braces balanced, every symbol declared ahead of first use.
- **The cross-check above ~40 mA.** Below that the true shunt drop is under one bus LSB.
- **The noise floor discrepancy** — −120 in this capture against −98/−99 from the RC52 capture earlier the same day. Unexplained; contradicts nothing, recorded rather than dismissed.

## Known limits, recorded as limits

`bus_mv` cannot resolve the discharge (4 mV LSB); sleep current is below the R100 shunt's 100 µA floor; 30-second sampling misses TX bursts, which is what the INA228's charge accumulator fixes (#857).

## Issues

Refs #938, #625, #740, #833, #857. Closes nothing on merge — #938 closes on owner authorization.

Agent: QuietBarn (session 5ef417e8)

### DIFF
--- a/tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino
+++ b/tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino
@@ -145,6 +145,37 @@
   #define PIN_GAUGE_SCL A3
 #endif
 
+// INA219 REGISTERS -- #938. [verified: TI INA219 datasheet (sbos448), fetched
+// 2026-08-22]
+//
+//   0x00 Config (reset 0x399F)   0x01 Shunt V   0x02 Bus V
+//   0x03 Power                   0x04 Current   0x05 Calibration
+//
+// !! ONE DECODE DETAIL THE DATASHEET SUMMARY GOT WRONG, AND WHY WE DO NOT
+// !! FOLLOW IT. A summary of the datasheet reported the SHUNT register as
+// !! "bits 15-3", the same shift as the bus register. That cannot be right at
+// !! the default configuration, and the config register proves it:
+// !!
+// !!     reset value 0x399F = 0011 1001 1001 1111
+// !!                            ^^  PG = 0b11 -> PGA /8 -> +/-320 mV range
+// !!
+// !! At 10 uV per LSB, +/-320 mV needs +/-32000 counts. A 13-bit field (bits
+// !! 15-3) holds only +/-4096 counts = +/-40.96 mV, which is the PGA /1 range,
+// !! not the default. So the shunt register is a FULL signed 16-bit value and
+// !! must NOT be shifted. Computed from the mechanism rather than read off the
+// !! annotation -- shifting it would have quietly reported 1/8th of the real
+// !! current, a plausible number that no sanity check would catch.
+//
+// The BUS register genuinely is bits 15-3, with CNVR at bit 1 and OVF at bit 0.
+#define INA219_REG_SHUNT   0x01
+#define INA219_REG_BUS     0x02
+
+// Bench shunt, owner-confirmed from the silkscreen: R100 = 0.100 ohm.
+// Shunt LSB 10 uV / 0.1 ohm  ->  exactly 100 uA per count. Range +/-3.2 A.
+// Blind to single-digit-uA sleep current; that is a property of the shunt, not
+// the chip, and no averaging setting recovers it.
+#define INA219_SHUNT_MILLIOHM  100
+
 #define MAX1704X_ADDR    0x36
 #define MAX1704X_VCELL   0x02
 #define MAX1704X_SOC     0x04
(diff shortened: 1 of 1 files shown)


Strycher:
Agent: QuietBarn (session 5ef417e8)

Strycher:
Human approval received from Ben in session 5ef417e8.

Re-validated against the remote immediately prior to enabling auto-merge: **4 commits, 1 file** (`tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino`, +689/-0), `mergeable=MERGEABLE`, base `firmware-base`. Checks at that moment: **34 SUCCESS, 13 pending, 0 failures** — `mergeStateStatus=BLOCKED` solely because `ci-green` has not aggregated yet.

Unchanged from what he approved. Auto-merge enabled with `--rebase`; GitHub performs the merge once the matrix completes.

Agent: QuietBarn (session 5ef417e8)

Issue #973, opened 2026-08-23
bug: companion UI screens are allocated with unchecked new — a null screen silently stops all rendering on 60 variants


Strycher:
## Summary

`UITask::begin()` allocates the three companion screens with `new` and **checks none of them**:

```cpp
splash = new SplashScreen(this);
home = new HomeScreen(this, &rtc_clock, sensors, node_prefs);
msg_preview = new MsgPreviewScreen(this, &rtc_clock);
```

A null screen is **completely silent**. `gotoHomeScreen()` sets `curr = nullptr`, the render loop's guard `if (millis() >= _next_refresh && curr)` stops firing, and nothing is ever drawn again. The board keeps running perfectly — radio, mesh, BLE, heartbeat all normal — while the display shows the last frame that happened to be blitted, forever.

There is no error, no log line, and no degraded mode. *"Silent | PROHIBITED | This tier does not exist."*

## Blast radius

`[verified: grep of variants/*/platformio.ini, 2026-08-23]` **60 variants compile `examples/companion_radio/ui-new/`.** This is shared code, not one board's problem.

## Confirmed live, not theoretical

`[verified: RC52 hardware, heltec_rc52_companion_radio_ble_diag, 2026-08-23 20:19Z]`

```
[817]  [ui] screens alloc: splash=1 home=0 msg_preview=1
[822]  [ui] setCurrScreen NULL -> SPLASH
[3817] [ui] setCurrScreen SPLASH -> NULL
```

`new HomeScreen(...)` returned nullptr on RC52's BLE companion. Note `msg_preview` allocated **successfully afterwards** — so this is not blunt exhaustion in order; HomeScreen is simply the largest of the three and needed a block that was not available.

Symptoms as the owner experienced them, all of which follow from the null:
- panel appears "stuck on the splash screen" indefinitely;
- after auto-off, a button wake shows a flash of the old frame then a blank backlit panel;
- board otherwise completely healthy, so every instinct says the display driver is broken.

Diagnosing it took a hardware flash cycle, a purpose-built driver instrument, and a UART sniffer — for a failure a two-line null check would have named instantly.

## A second, compounding defect — the trace lied

The screen-transition logging tested the pointers in the wrong order:

```cpp
if      (c == splash)         to = "SPLASH";
else if (c == home)           to = "HOME";      // matches when home is nullptr AND c is nullptr
else if (c == nullptr)        to = "NULL";      // unreachable in exactly that case
```

[...]

PR #976, opened 2026-08-24
task(#972): ungate companion UI tracing so nRF52 boards can emit it

### LINKED ISSUE #972: task: UI state-machine tracing is gated on OFFBAND_OBSERVER, so no nRF52 board can ever emit it
The companion UI's state-machine tracing is gated on `OFFBAND_OBSERVER`:

```cpp
#ifdef OFFBAND_OBSERVER
  offband::crashLogf("[ui] setCurrScreen %s -> %s", from, to);
#endif
```

`OFFBAND_OBSERVER` implies WiFi. **nRF52 has no WiFi**, so on every nRF52 board — RC52, RAK4631, T1000-E, Wio, T096, XIAO — this instrumentation is compiled out entirely.

That is backwards. The tracing exists *because* of a UI bug: its own comment in `UITask.cpp` reads

> SplashScreen ↔ HomeScreen oscillation observed on hv3-bench is a real state-machine bug.

So the purpose-built instrument for a known screen-state defect is unavailable on a whole class of boards, and available only on the ones with WiFi — which has nothing to do with the UI.

## Found by

RC52 (#951) shows the splash and then does not visibly progress. The board is healthy — `display.begin(OK)`, `setup:DONE`, radio up, heartbeat running, no errors — so the question is purely which screen the UI thinks it is on and whether it is rendering. That is exactly what these four `crashLogf` calls answer, and not one of them is compiled in.

`[verified: grep of examples/companion_radio/ui-new/UITask.cpp, 2026-08-23 — 5 gates at lines 9, 55, 725, 763, 892; 4 crashLogf calls at 56, 726, 777, 893]`

## Change

Gate on a purpose-named macro derived from both, rather than moving the tracing from one unrelated flag to another:

```cpp
#if defined(OFFBAND_OBSERVER) || defined(OFFBAND_BOOT_BEACON)
  #define OFFBAND_UI_TRACE 1
#endif
```

- **Observer boards are unchanged** — they still get it, so no regression anywhere it works today.
- **`OFFBAND_BOOT_BEACON` brings in the RadioCore `_diag` envs** (RC32, RCC6, RC52 all set it), which matches the standing directive that every env built during bring-up is a full diagnostic build.
- The name says what it is. `OFFBAND_BOOT_BEACON` is about *boot*; this is *UI runtime* tracing, and piggybacking the semantics would mislead the next reader.

`crashLogf` itself is already unconditional in `CrashLog.h` — only the call sites are gated, so nothing else has to move.

## Scope

`examples/companion_radio/ui-new/UITask.cpp` only. Five `#ifdef` lines, no logic change, no behavior change on any board that has the tracing today.

Owner-directed 2026-08-23.

Strycher:
Closes #972

## What and why

The companion UI's four `crashLogf()` tracing call sites were gated on `OFFBAND_OBSERVER`. That flag implies WiFi, and **nRF52 has none** — so on every nRF52 board (RC52, RAK4631, T1000-E, Wio, T096, XIAO) this instrumentation compiled out entirely.

That is backwards. The tracing exists *because* of a UI bug — its own comment records a `SplashScreen ↔ HomeScreen` oscillation seen on hv3-bench as a real state-machine defect. Gating it on an unrelated radio capability removed it from a whole class of boards.

Gated now on `OFFBAND_UI_TRACE`, defined when either `OFFBAND_OBSERVER` **or** `OFFBAND_BOOT_BEACON` is set. Derived from both rather than moved onto the beacon flag: observer boards keep it byte-identically, `OFFBAND_BOOT_BEACON` brings in the RadioCore `_diag` envs, and the name says what it controls — `OFFBAND_BOOT_BEACON` is about *boot*, this is *UI runtime* tracing.

## It paid for itself immediately

RC52 showed its splash and never visibly progressed, on a board that was otherwise perfectly healthy — `display.begin(OK)`, `setup:DONE`, radio up, heartbeat running, no errors. With the tracing compiled in:

```
[ui] screens alloc: splash=1 home=0 msg_preview=1
[ui] setCurrScreen SPLASH -> NULL
```

`new HomeScreen(...)` returns **null**. `gotoHomeScreen()` sets `curr = nullptr`, the render loop's `&& curr` guard stops firing, and the panel holds the last blitted frame forever. Filed as #973 — it affects **60 variants**, since `new` is unchecked at all three screen allocations.

**And the trace was actively lying.** It tested `c == home` before `c == nullptr`, so with `home` null a transition to *nothing* printed `"HOME"`. The NULL arm was unreachable in exactly the case it existed to catch. That misdirected the first round of diagnosis at the display driver. Now null-first, with a comment saying not to reorder it.

## Also here

- **which screens got memory**, plus `sizeof()` of each — the unchecked `new` itself is #973 and deliberately not fixed here
- **an opt-in heap probe** reporting the largest contiguous allocatable block, which is what `new` actually has to find

## Gemini review — findings and dispositions

Run twice: once on the original, once on the rewrite, because a review of code I then changed is not a review of what ships.

| Finding | Severity | Disposition |
|---|---|---|
| Probe could raise the heap ceiling it measures | High | **Fixed.** Was 128 malloc calls stepping down from 64 KB; on a roomy board that request *succeeds* and newlib keeps the sbrk extension. Now a 9-probe binary search, ceiling 24 KB. |
| Probe should be gated separately from tracing | Medium | **Fixed.** Needs `OFFBAND_UI_TRACE` **and** `OFFBAND_UI_TRACE_HEAP`; off by default. Verified in the ELF both ways. |
| `if/else` screen-name chain is brittle | Low | **Noted, not changed.** A lookup table centralises names but costs storage on every board for a diagnostic most never enable. Documented in place. |
| "Probe reports 256 when nothing succeeds" | — | **Refuted.** The old loop decremented 65536 by 512 and landed exactly on 0. The re-review accepted the arithmetic and withdrew it. |
| `if (mid < PROBE_FLOOR) break` abandons the interval early | Low | **Fixed** (found by the re-review, in my own rewrite). With 80 B free the search reaches `lo=0/hi=96`, computes `mid=48`, breaks, reports `0` — the same misleading zero the rewrite existed to remove. |

**One defect the review did not raise, found while fixing it:** the old probe stepped by 512 B, so its resolution floor was 512 — and `sizeof(HomeScreen)` is **480**. It could not resolve the allocation it existed to measure, and reported "0" for anything under 512 B. `PROBE_FLOOR` is now 64.

## Verification

`[verified: pio run]`

| Env | Result |
|---|---|
| `Heltec_v3_companion_observer_wifi` | SUCCESS — observer boards unchanged |
| `heltec_rc32_companion_radio_usb_diag` | SUCCESS — newly traced |

Gate scope checked against the built ELF rather than asserted:

```
'[ui] largest_free_block=' default            → 0   (probe off)
'[ui] largest_free_block=' -D ..._TRACE_HEAP  → 1   (probe on)
'[ui] sizeof' / '[ui] screens alloc'          → 1   (tracing on)
```

**No behaviour change on any board.** Everything added compiles out unless its flag is set, and the observer boards that already had this tracing resolve exactly what they did before.

Hardware: the tracing itself ran for hours on `rc52-bench-1` this session via the UART mirror, and is what located #973.

### DIFF
--- a/examples/companion_radio/ui-new/UITask.cpp
+++ b/examples/companion_radio/ui-new/UITask.cpp
@@ -658,6 +658,35 @@ void UITask::begin(DisplayDriver* display, SensorManager* sensors, NodePrefs* no
   splash = new SplashScreen(this);
   home = new HomeScreen(this, &rtc_clock, sensors, node_prefs);
   msg_preview = new MsgPreviewScreen(this, &rtc_clock);
+#ifdef OFFBAND_UI_TRACE
+  // #951: these three are new'd with no null check, and a null `home` is SILENT --
+  // gotoHomeScreen() sets curr = nullptr, the render loop's `&& curr` guard stops
+  // firing, and the panel simply holds whatever was last blitted. On a memory-tight
+  // board that presents as "stuck on the splash screen" with a fully healthy board
+  // behind it. Log which of them actually got memory.
+  offband::crashLogf("[ui] screens alloc: splash=%d home=%d msg_preview=%d",
+                     (int)(splash != NULL), (int)(home != NULL), (int)(msg_preview != NULL));
+
+  // #856: how much did they actually need, and how much was there?
+  //
+  // sizeof() is the exact object size. The second figure is the LARGEST
+  // CONTIGUOUS BLOCK still allocatable, probed by walking malloc down until one
+  // succeeds and immediately freeing it -- that, not total free bytes, is what
+  // `new HomeScreen` has to find. Freeing 56 KB of back buffer was known to be
+  // sufficient; this says what is actually REQUIRED, so the buffer can be sized
+  // on evidence instead of on a guess.
+  {
+    size_t probe = 64u * 1024u;
+    while (probe > 256u) {
+      void* p = malloc(probe);
+      if (p) { free(p); break; }
+      probe -= 512u;
+    }
+    offband::crashLogf("[ui] sizeof: splash=%u home=%u msg=%u | largest_free_block=%u",
+                       (unsigned)sizeof(SplashScreen), (unsigned)sizeof(HomeScreen),
+                       (unsigned)sizeof(MsgPreviewScreen), (unsigned)probe);
+  }
+#endif
   setCurrScreen(splash);
 }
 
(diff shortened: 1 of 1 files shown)


Strycher:

Issue #977, opened 2026-08-24
bug: log-mirror putc() is not reentrant — latent, unreachable until something logs from an ISR


Strycher:
Recorded from the Gemini adversarial gate on the #944 epic (finding 1.1). **Deliberately not fixed** — filed so it is not rediscovered from scratch if it ever becomes reachable.

## The defect

`offband_log_mirror_putc()` in `src/helpers/LogMirrorUart.h` is not reentrant, on both back ends:

- **nRF52** — a single `static volatile uint8_t byte_buf` plus the shared `TXD.PTR` / `TXD.MAXCNT` / `TASKS_STARTTX` registers. An interrupt landing between the buffer store and the `EVENTS_ENDTX` wait can overwrite the byte and restart the transfer, aborting the one in flight.
- **ESP32** — writes the FIFO register with no mutual exclusion, same exposure.

Consequence if triggered: the preempted character is lost or garbled. The diagnostic channel produces unreliable output exactly when something unusual is happening.

## Why it is not being fixed now

**It is not reachable in the tree as it stands** [verified 2026-08-24]:

- the only `attachInterrupt()` under `src/` is `setPmuFlag()` in `src/helpers/esp32/TBeamBoard.cpp:173`, which sets a `bool` and does not log;
- there are no `IRAM_ATTR` handlers, no bare `*_Handler(void)`, and no `__attribute__((interrupt))` anywhere under `src/`;
- nothing in the tree logs from ISR context.

The one live preemption path is a fault handler firing while `loop()` is mid-`putc`. That costs one mirrored character *before* the crash dump is emitted, which is the correct trade in a fault.

The obvious fix is also actively harmful here. Wrapping `putc` in `__disable_irq()` places the bounded spin loop inside a critical section, trading an unreachable race for a reachable multi-millisecond stall with interrupts off — on the same fault path that finding 2.1 (the oversized guard, fixed in this epic) was about.

## What makes this live again

Any ISR that logs. If one is added, this stops being theoretical and must be fixed then — by making the wait short enough to sit safely in a critical section, or by queueing from ISR context, **not** by wrapping the loop in its current form.

The constraint is recorded in-place as a comment above both `putc` implementations so it is visible at the point where someone would violate it.

Strycher:
**Cross-reference: this is part of the diagnostic-log-integrity body of work tracked by #1091.**

[...]

Issue #953, opened 2026-08-23
task: route CrashLog to the mirror UART so reset reasons survive USB re-enumeration


Strycher:
## Problem

CrashLog writes only to `Serial`. On a board whose `Serial` is native USB CDC, that transport **disappears at the moment of the reset CrashLog exists to report**.

`src/helpers/diagnostics/CrashLog.cpp:198-209`:

```c
static size_t crashLogSerialWrite(const char* data, size_t len) {
    if (len == 0) return 0;
    if (!Serial) return 0;                       // no host -> drop, never wait
```

Meanwhile `OFFBAND_LOG_MIRROR_UART` carries **MeshLog** out a physical UART pin -- a wire that is unaffected by USB enumeration and keeps recording straight through a reset. CrashLog does not use it.

The result on the bench is precise, and was hit directly: a build with `OFFBAND_FORCE_CAPLOG`, the boot beacon, and a working UART mirror **still could not report why the board reset**. The sniffer recorded the reset happening; the one line naming its cause went to a transport that had just vanished.

That is at odds with the standing directive that every environment built during bring-up is a full diagnostic build. The most valuable line a diag build emits -- `[boot] <role> up; reset=<reason>`, written by `crashLogStandardInit` on every boot -- never reaches the diagnostic channel.

## Proposal

In builds that define `OFFBAND_LOG_MIRROR_UART`, route CrashLog output to the mirror in addition to `Serial`.

The natural seam is `crashLogSerialWrite`, since every path funnels through it:

```c
static size_t crashLogSerialWrite(const char* data, size_t len) {
    if (len == 0) return 0;
#if defined(OFFBAND_LOG_MIRROR_UART)
    offbandLogMirrorWrite(data, len);   // survives the reset; never gates on a host
#endif
    if (!Serial) return 0;
    ...
```

Constraints that must hold, and they are not negotiable given the history in #741/#756:

- **Never block.** The mirror write must be bounded and non-blocking, same contract as the CDC path. A truncated crash log beats a device that will not boot.
- **Zero cost when the mirror is off.** Non-diag builds must be byte-identical.
- **No recursion.** MeshLog must not route back into CrashLog.
- Whether the mirror UART is initialized early enough for `crashLogBegin()` needs checking -- the boot beacon already writes to that pin before `board.begin()` (verified on RC52 in #888/#936), which is encouraging but is a different code path.

## Why it matters beyond one board

[...]

Issue #952, opened 2026-08-23
bug: deferred crash-log re-dump is consumed at T+5s even when no serial host is attached


Strycher:
## What happens

The deferred crash-log re-dump is consumed **at T+5 s whether or not a host is attached**, so the one chance to recover the previous boot's crash log is silently spent into a closed transport.

`src/helpers/diagnostics/CrashLog.cpp:404-408`:

```c
void crashLogTick(uint32_t now_ms) {
    if (s_prev_pending && now_ms >= 5000u) {
        s_prev_pending = false;                 // cleared BEFORE the attempt
        emitPreviousBootDump("deferred re-dump for late serial connect");
    }
}
```

`emitPreviousBootDump` (`CrashLog.cpp:243-247`) then gives up if nothing is listening:

```c
static void emitPreviousBootDump(const char* why) {
    if (s_prev_len == 0) { return; }
    if (!Serial) return;                        // nobody listening -> silently nothing
```

The flag has already been cleared by the caller, so the early return discards the only opportunity. **A monitor attached at T+6 s gets nothing, forever.**

## Why this is a defect and not a design choice

The comment at `CrashLog.cpp:261-263`, inside the very function that drops the dump, states the opposite:

> The deferred re-dump (#378/#463) still gets a second chance later once a monitor has attached.

There is no second chance. The flag is one-shot.

#378 exists specifically to serve monitors that attach **after** the reset. As written, the effective window is "attached before T+5 s" — which on a native-USB board that re-enumerates after every reset is a race no human wins.

## How it was found

`rc52-bench-1` (RC52, nRF52840) reset itself once, 29.8 s into a room_server boot. Chasing the reset reason:

1. Serial monitor attached to the USB CDC. Uptime kept climbing, so the attach itself was clean.
2. Owner pressed RST. Board reset at 06:25:32 -- witnessed on the mirror UART by an independent sniffer.
3. The CDC vanished at the reset; PlatformIO's monitor does not reconnect; its log froze mid-stream.
4. At 06:25:37 the deferred dump fired into the dead port and cleared its flag.

The reset reason is now unrecoverable. Full trace: #889.

**The data is still in the device's RAM.** Only `s_prev_pending` was consumed -- `s_prev_snapshot` and `s_prev_len` are untouched. Nothing can ask for it; no CLI command reaches `emitPreviousBootDump`.

## Fix

Gate the flag on the transport, so it is spent only when the dump can actually land:

```c
[...]

Issue #946, opened 2026-08-23
task: Epic verification — nRF52 diagnostic channel


Strycher:
Final task of Epic #944. Proves the diagnostic channel works as one unit in the shared code.

## Acceptance

1. **`APP:CTOR` and the `setup:` markers captured on a NON-companion role** — repeater, room_server or sensor. A companion-only capture is precisely what let #936 survive undetected.
2. **RC32 and RCC6 build and behave identically.** Production consumers of this code. Evidence required, not assertion.
3. `heltec_rc32_repeater_testdiag` — which already sets `OFFBAND_BOOT_BEACON` and emits nothing today — now emits. Confirmed, and recorded as an intended change to a shipping diag artifact.
4. A non-diag build is byte-identical.
5. #927's false platform text no longer appears on nRF52.

Test target is RC52: the only nRF52 on this bench that can be wired to a sniffer. RC52 is not what is being validated — the shared code is.

**Epic closure is a HUMAN ONLY gate.**

Parent: #944.

PR #978, opened 2026-08-24
epic: nRF52 diagnostic channel — portable UART mirror, boot beacon, CrashLog routing

Strycher:
Epic [#944](https://github.com/OffbandMesh/meshcore-firmware/issues/944) — the nRF52 arm of the diagnostic channel, in shared code.

## Why

On most boards in this fleet `Serial` is USB, and USB dies with the chip — so every log ever captured that way came from a boot that **succeeded**. A UART mirror on a physical wire witnesses the boots that fail. That capability existed but was ESP32-only, and `BootBeacon.h` raised a hard `#error` on any other target. This ports both to nRF52 (UARTE/EasyDMA) and routes the crash log through the mirror.

This is **fleet-shared code — roughly 60 variants compile it**, which is why the review gate below was aimed at regression risk rather than style.

## What changed

| Area | Change |
|---|---|
| `src/helpers/LogMirrorUart.h` (new) | Portable raw-UART mirror. ESP32 arm lifted from the previous in-`BootBeacon` implementation; new nRF52 UARTE/EasyDMA arm. |
| `src/helpers/BootBeacon.h` | Was ESP32-only with a hard `#error` elsewhere. Now portable — the one genuine rewrite here. |
| `src/helpers/diagnostics/CrashLog.{h,cpp}` | Crash log routed out the mirror as well as USB ([#953](https://github.com/OffbandMesh/meshcore-firmware/issues/953)); deferred re-dump no longer consumed when no transport is attached ([#952](https://github.com/OffbandMesh/meshcore-firmware/issues/952)). |
| `src/MeshLog.cpp` | Mirror hookup. |
| `simple_repeater`, `simple_room_server`, `simple_sensor` | Boot beacon was silently inert on these roles ([#936](https://github.com/OffbandMesh/meshcore-firmware/issues/936)); wired in. |
| 8 variant `platformio.ini` | `OFFBAND_MESHLOG_UART0` → `OFFBAND_LOG_MIRROR_UART`. |
| tests | Two new native suites: `test_crashlog_ring`, `test_crashlog_deferred`. |

Six areas attacked. **Two cleared outright**: the deferred-dump fix (no path loses or double-delivers a dump) and zero-cost-when-disabled (genuinely zero flash/RAM/boot cost for a board setting neither flag).

**Fixed in `8b5c227c`:**

- **2.1 (Medium) — oversized per-byte bound.** The wait was capped at 2,000,000 spins (~125 ms) to bound a wait whose legitimate duration is one character time — 87 µs at 115200, 1.04 ms at 9600, the slowest baud the table supports. Over 1000× headroom on a path `CrashLog` runs from fault and shutdown handlers, where a stall that long can let a watchdog turn a recoverable fault into a hard reset and lose the dump. Now `OFFBAND_LOG_MIRROR_BYTE_GUARD` (2¹⁷): ~8× headroom over a 9600-baud byte, worst-case stall near 8 ms. The ESP32 FIFO-drain guards deliberately keep the large bound — they wait for up to a full 128-byte hardware FIFO (~11 ms at 115200), so shrinking them would drop bytes on a healthy board. Reasoning recorded beside the constant.
- **5.1 (Medium) — silent loss on a half-renamed variant.** A variant left on `OFFBAND_MESHLOG_UART0` would still compile and simply have no mirror: the diagnostic that exists to catch silent failures failing silently. Now a hard `#error`. **Negative-tested** — building with the old flag produces the refusal. Four stale doc references updated in `.github/release-envs.txt` and `docs/rc32-beta-checklist.md`; `CHANGELOG.md` deliberately left, as it records the flag under the name it shipped with.

**Recorded, deliberately not fixed:**

- **1.1 (rated High) — `putc()` is not reentrant** on both back ends. Filed as [#977](https://github.com/OffbandMesh/meshcore-firmware/issues/977) and documented above both `putc` implementations. It is **not reachable in the tree as it stands** [verified 2026-08-24]: the only `attachInterrupt()` under `src/` is `setPmuFlag()` in `src/helpers/esp32/TBeamBoard.cpp:173`, which sets a `bool` and does not log, and there are no `IRAM_ATTR` or bare interrupt handlers anywhere — nothing logs from ISR context. The proposed fix is also worse than the defect: wrapping `putc` in `__disable_irq()` puts the spin loop inside a critical section, trading an unreachable race for a reachable multi-millisecond stall with interrupts off, on the same fault path 2.1 is about. It becomes live the moment any ISR logs, and #977 says so.

**Declined — the gate's own non-findings, no change proposed for either:**

- **1.2** — `g_ADigitalPinMap` read from `constructor(101)`. It lives in `.data`, which the C runtime copies from flash before any constructor runs; guaranteed by the startup sequence, not luck.
- **3.1** — new-target fallthrough. The gate concluded the existing `#error` already handles this correctly.

## Verification

- `pio run` **SUCCESS**: `heltec_rc32_repeater`, `Heltec_v3_companion_observer_wifi`, `RAK_4631_companion_radio_ble` — re-run after rebase onto `firmware-base`.
- **The nRF52 arm was compiled explicitly**, with `OFFBAND_LOG_MIRROR_UART` + `OFFBAND_BOOT_BEACON` forced on. No variant enables it yet — RC52's `_diag` envs land on a later branch — so the three-env build alone would **not** have exercised the changed code. Force-flagged artifacts were deleted afterward so nothing stale can be flashed.
- Rename tripwire negative-tested (build refuses with the old flag name).
- Native tests **285/285**, including the two new crash-log suites.
- Secrets/private-IP scan clean.

## Not verified here

Hardware behavior of the nRF52 mirror on a live board. No variant ships the flags yet; that verification belongs to [#946](https://github.com/OffbandMesh/meshcore-firmware/issues/946) (epic verification), which stays open through merge.

## Merge

**Ben merges.** Not auto-merge — `ci-green` is a required gate and this is a fleet-shared change.


Strycher:

Issue #985, opened 2026-08-25
bug: RC52 powerOff() drops the FEM rail before the radio is quiesced


Strycher:
Raised by the Gemini gate on epic [#879](https://github.com/OffbandMesh/meshcore-firmware/issues/879). **Pre-existing, not introduced by that branch** — owner ruled 2026-08-25 to record it and deal with it later.

## What the gate claimed, and why that specific claim is wrong

The gate rated this High: *"`RC52Board::powerOff()` cuts power to the entire radio module before `NRF52Board::powerOff()` → `shutdownPeripherals()` sends SPI commands to a now-unpowered chip."*

**The premise is refuted.** `[verified: rc52.pdf, vendor schematic]` Module U6 has **two independent supply pins**:

| Pin | Net | Gated by |
|---|---|---|
| 10 | `VDD_3V3` | nothing — always powered |
| 25 | `VDD_FEM` | U14 (`TLV75733PDBVR`), enable = `VFEM_Ctrl` |

`setLoRaFemLnaEnabled(false)` drives `VFEM_CTRL` low, which drops **`VDD_FEM` only**. The SX1262 runs from `VDD_3V3` and stays powered. So `shutdownPeripherals()` talks SPI to a live chip, and there is no "SPI to an unpowered radio" defect.

## The residual that DOES survive, and is worth fixing

`shutdownPeripherals()` (`src/helpers/NRF52Board.cpp`) does more than SPI. It:

1. hardware-resets the radio via `P_LORA_RESET` (low ≥100 µs, then high),
2. calls `SPI.begin()`,
3. reads `P_LORA_BUSY`.

On RC52 that sequence runs **after** `VDD_FEM` has already been dropped. Resetting the SX1262 puts its DIO2 rf-switch output into its reset state, and on this board **DIO2 is internal to U6 and drives the module's internal T/R switch** — which is on the now-unpowered FEM rail. That is the same failure class as [#931](https://github.com/OffbandMesh/meshcore-firmware/issues/931): a control line driven into an unpowered load, forward-biasing an ESD structure and back-powering through the signal pin.

Severity is lower than #931's because it happens once on the way into SYSTEMOFF rather than every boot, and the duration is short. But it is the same defect shape, and #931's whole point was that "nothing defends leaving a control line asserted into a dead rail."

[...]

Issue #983, opened 2026-08-24
task: wire OFFBAND_CAP_FEM_LNA runtime control on RC52 (module-power semantic)


Strycher:
Owner-directed 2026-08-24. RC52 was the only FEM-bearing board in the fleet **not** exposing the stock `OFFBAND_CAP_FEM_LNA` runtime-control capability that HV4 and T096 expose. The owner asked for it explicitly — parity with every other FEM control board — and its absence forced a throwaway `-D` build request just to A/B the FEM for the [#963](https://github.com/OffbandMesh/meshcore-firmware/issues/963) TX-deficit measurement. Wiring the capability removes that build and closes the gap.

## Why it was missing (the reasoning was wrong)

`RC52Board.h` recorded that overriding `canControlLoRaFemLna()` "would mean taking RXEN away from RadioLib's rf-switch." That is true **only if the capability maps to the LNA/RXEN pin** (P1.07), which RadioLib does own. But RC52 has three FEM lines, and two of them — `RADIOCORE_FEM_EN` (P0.26) and `RADIOCORE_VFEM_CTRL` (P0.16), the whole-module power — are board-owned and untouched by RadioLib. The capability maps to those. The decline read the comment, not the three pins.

## Decision: same framework, divergent semantic (owner-approved)

Adopt the **exact stock scaffold** — the three `MainBoard` virtuals, the `OFFBAND_CAP_FEM_LNA` bit, the `fem on/off` CLI, the persisted `radio_fem_rxgain` pref — but with an RC52-specific meaning under the hood, because the hardware differs:

| | T096 / HV4 (KCT8103L) | RC52 (HT-RA62A) |
|---|---|---|
| `setLoRaFemLnaEnabled(false)` | LNA out of the **RX** path; TX and module power untouched | **whole module unpowered** (FEM_EN + VFEM_CTRL low); TX+RX both affected |
| RXEN/LNA line | board-controlled | **stays with RadioLib's rf-switch — untouched** |

Same label, different physical effect. The divergence is deliberate and documented in the board header and CLI-adjacent comment so nobody reads RC52 `fem off` as "RX gain only."

## Implementation

[...]

Issue #982, opened 2026-08-24
bug: single-sample boot-voltage check can shut down a viable cell — and RC52 cannot self-recover from it


Strycher:
Raised by the Gemini review gate on epic [#879](https://github.com/OffbandMesh/meshcore-firmware/issues/879) (finding 2, rated High). Filed rather than fixed in that epic by owner decision 2026-08-24 — it is a **pre-existing shared behavior**, not something #879 introduces, and fixing it there would widen a variant epic into shared power-management work.

## The defect

`NRF52Board::checkBootVoltage()` takes a **single** battery sample at boot and compares it against a fixed `PWRMGT_VOLTAGE_BOOTLOCK` threshold. If that one sample reads low, the board enters SYSTEMOFF to protect the cell.

A healthy cell can produce a low sample transiently:

- a cold battery has elevated internal resistance and sags under load
- inrush at boot — the radio and, on RC52, the FEM rail coming up — pulls current before the sample is taken
- a bench supply with a slow rise looks identical to a depleted cell at the sampling instant

The result is a protective shutdown on a viable battery.

## Why RC52 makes it worse

On most nRF52 boards a false trip is recoverable: LPCOMP is armed for voltage wake, so the board brings itself back when the cell recovers.

**RC52 deliberately has no LPCOMP wake.** Its battery divider is FET-gated, so feeding LPCOMP through SYSTEMOFF means holding the gate transistor on — a board that shut down to protect a depleted cell would drain it faster while off than it could recover. RC52 therefore arms USB-attach wake only (`configureUsbWake()`, added in #879).

Consequence: **on RC52 a false trip leaves the board inert until physically plugged into USB.** For a deployed node that means retrieval. That amplification is RC52-specific and new; the sampling weakness itself is shared.

## Scope

`PWRMGT_VOLTAGE_BOOTLOCK` is 3300 mV on RC52, matching eight other nRF52 variants (two use 3100). So any change here is a **fleet-wide power-management change** and needs to be treated as one — this is explicitly not a variant-local fix.

## Options

1. **Sample, delay, re-sample.** Reject a single low reading; only shut down if it persists. Costs a few ms of boot and nothing else. Most likely correct.
2. **Sample after inrush settles** rather than before, so the reading is taken under steady-state load.
3. **Lower the threshold** for headroom. Cheapest, but trades away the protection the check exists to provide, and does not remove the failure mode.
4. **Document only.** Honest but leaves a field-inert-board path open on RC52.

[...]

Issue #965, opened 2026-08-23
bug: RC52 room server watchdog-resets at ~29 s and renders nothing on the first boot


Strycher:
## Summary

On `rc52-bench-1`, the room server role **watchdog-resets ~29 s after boot**, once. On that first boot the **TFT does not render**, although the board is otherwise fully functional. After the watchdog reset the board comes up again and the display works, and it stays up.

This is the "29.8 s room_server reset" recorded as unexplained in the RC52 bring-up handoff. It now has a reset reason and a reproduction. **It predates the display work** — it was first seen on a headless room-server build, so the display is not its cause.

## Evidence

`[verified: Feather sniffer on COM16 reading the UART mirror off carrier header pin 11, 2026-08-23]`
Board: `rc52-bench-1`, env `heltec_rc52_room_server_diag`, branch `epic/948-rc52-display` @ `8ea3ace3`, artifact sha256 `8b14237b54ae33add70753e8a7961c25a5b528959d9b2a982ff48e4b743cfd0e`.

```
18:15:24.812  [1061] [boot] room-server up; reset=Soft Reset     <- post-DFU. NO DISPLAY.
18:15:54.601  [1071] [boot] room-server up; reset=Watchdog       <- 29.8 s later. DISPLAY WORKS.
```

Owner confirmed both halves visually: dark panel on the first boot, then an audible reset and the display came up.

**The failing boot was not hung.** It completed `setup()`, initialised the radio and entered the main loop:

```
18:15:24.815  [BEACON] setup:post crashLogStandardInit
18:15:25.551  DEBUG: Set _preambleMillis=91          <- 736 ms after the last beacon
18:15:25.583  DEBUG: loop - skipping busy (or evicted) client 4F
18:15:25.660  DEBUG: RadioLibWrapper: noise_floor = -119
```

That 736 ms gap is the window `display.begin()` occupies, and it matches the **739 ms** measured on a repeater boot whose panel rendered correctly. So the display driver executed and consumed its normal init time on the failing boot too — it is not being skipped and it is not hanging.

The board then ran normally for ~29 s, serving client `4F` at ~156 ms intervals, before the watchdog fired.

## Not a loop

One reset only. The board has been up continuously since 18:15:54 with no further boots.

## What is NOT the cause

- **Not the display driver.** The same driver renders correctly on the repeater role on this board, first boot, every time.
[...]

PR #986, opened 2026-08-25
epic: RC52 headless bring-up — nRF52840 variant, FEM capability, diagnostics, all roles

Strycher:
Epic [#879](https://github.com/OffbandMesh/meshcore-firmware/issues/879) — RC52 headless bring-up. Third of the four RadioCore boards, and the first nRF52 one.

## What this is

Heltec RadioCore RC52: nRF52840 + HT-RA62A (SX1262 **with a FEM inside the module**). RC32 and RCC6 are shipping vetted boards; this brings the nRF52 sibling up headless — repeater, room server, companion USB/BLE — plus a `_diag` twin per role.

## What changed

| Area | Change |
|---|---|
| `variants/heltec_rc52/` (new) | variant, board class, targets, `platformio.ini` — 5 shipped envs + `_diag` twins |
| `boards/heltec_rc52.json` (new) | board definition |
| `src/helpers/NRF52Board.{cpp,h}` | **shared** — VBUS wake extracted from `configureVoltageWake()` into `configureUsbWake()` |
| `.github/workflows/ci.yml` | RC52 envs into the matrix |

**The only fleet-shared change is the `configureUsbWake()` extraction**, and it is a pure refactor: the block moved verbatim, called from exactly where it sat. RC52 needs it standalone because it **cannot use LPCOMP** — its battery divider is FET-gated, so holding the divider connected through SYSTEMOFF costs ~2.6 mA continuously. A board that shut down to protect a depleted cell would drain it faster than it could recover. USB-attach wake costs nothing standing.

## Hardware-verified

- **Display, repeater and room-server roles all render on the panel** (owner-confirmed).
- **Boot traced on the wire** via the UART mirror (P0.08, carrier header pin 11): `APP:CTOR` → `setup:ENTRY` → SafeBoot → `board.begin` → `crashLogStandardInit` → radio.
- **`fem on` / `fem off` exercised on hardware** — reports `can=yes` on RC52 and `can=no` on RC32/RCC6/RAK4631, which is the correct capability gating.
- **CrashLog init does not hang here** — the RC32 #740/#741 fault does not reproduce on this board.

## The FEM capability (#983), and its deliberate divergence

RC52 now exposes the stock `OFFBAND_CAP_FEM_LNA` capability, so `fem on/off` works on the CLI and the client shows the toggle as it does for HV4.3 and T096.

⚠ **It means something different on this board, and that is owner-approved.** On T096/HV4 (KCT8103L) "off" only pulls the LNA out of the **RX** path — `setTxModeEnable()` forces `CTX` HIGH unconditionally, so the flag has no TX effect. On RC52 (HT-RA62A) "off" cuts **whole-module power**: `FEM_EN` (P0.26) + `VFEM_CTRL` (P0.16) both low, TX and RX. It does **not** touch `SX126X_RXEN`/P1.07, which stays with RadioLib's rf-switch.

The prior code declined this capability on the grounds that claiming it "would mean taking RXEN away from RadioLib." That reasoning was wrong: it mapped the capability to the LNA pin, when two board-owned power pins were available that RadioLib never touches.

## ⚠ The TX deficit this epic chased does not exist

`[verified: owner-operated four-board bench sweep, 2026-08-25]`

| tx 22 | RC52 | RC32 | RCC6 | RAK4631 |
|---|---|---|---|---|
| | +20.30 | +20.30 | +20.30 | +20.30 |

All four ceiling identically — one FEM board and three without, three MCU families, two vendors. The **−1.70 dB is the measurement chain, not the radio**.

Both the original **−6.9 dB** and its first "correction" to **−2.04 dB** were sampling artifacts: a 2 MHz sweep for a 62.5 kHz signal parked the analyser on the carrier ~3% of each sweep, so most 150 ms bursts were never sampled. Resolved by zero-span gating.

Commit `45a794e1` strips those numbers from the source comment that cited them as evidence. **The FEM power-up reorder (#931) stands on general power-sequencing grounds only — it fixed nothing, because nothing was broken.** The comment says so explicitly, so the investigation is not resurrected from retracted figures.

Also recorded: the instrument quantises to **0.5 dB**, so board-to-board differences below that are unmeasurable on this bench.

**Fixed** — finding 4 (Low): `RC52Board.cpp` used `#include "../../src/helpers/diagnostics/CrashLog.h"`, the only file in the tree using that fragile relative form. Now `<helpers/diagnostics/CrashLog.h>`, matching how `src/` includes it.

**Found while checking the gate's premise, which the gate missed** — the comment above `checkBootVoltage()` read *"bootlock is 0, so this returns immediately."* False since #857 set the threshold; `PWRMGT_VOLTAGE_BOOTLOCK` is **3300 mV and armed**. Anyone reading it would have concluded a low-voltage trip was impossible here. Corrected, with the RC52-specific consequence recorded: a false trip has no automatic escape, because `initiateShutdown()` arms USB wake only.

**Refuted** — finding 1 (rated High): *"`powerOff()` cuts power to the entire radio module before SPI shutdown."* `[verified: rc52.pdf]` U6 has **two independent supplies** — `VDD_3V3` (pin 10, ungated) and `VDD_FEM` (pin 25, gated by `VFEM_Ctrl`). The SX1262 stays powered; there is no SPI-to-a-dead-chip path. A **residual does survive** — the radio reset drives internal DIO2 into an unpowered FEM rail — filed as [#985](https://github.com/OffbandMesh/meshcore-firmware/issues/985). Pre-existing; this branch preserved the existing ordering rather than introducing it.

**Deferred, owner-ruled** — finding 2 (High): the single-sample boot-voltage check can trip on a transient sag, and RC52 cannot self-recover. Tracked fleet-wide on [#982](https://github.com/OffbandMesh/meshcore-firmware/issues/982) (now raised by two independent gates). The gate's proposed fix — zeroing the threshold — was **rejected**: it trades a false trip for a brownout boot loop on a genuinely flat cell.

**Non-finding** — finding 3: the gate assessed its own fleet-regression concern about the `NRF52Board` extraction as *"purely theoretical… no tangible fleet risk"*, matching the independent [#944](https://github.com/OffbandMesh/meshcore-firmware/issues/944) gate verdict on the same code.

## Verification

`pio run` **8/8 SUCCESS**, re-run after rebase onto `firmware-base`:

- all five RC52 envs (`repeater`, `repeater_diag`, `room_server`, `companion_radio_ble`, `companion_radio_usb`)
- `RAK_4631_repeater`, `Heltec_t096_repeater`, `t1000e_repeater`, `heltec_v4_repeater_telemetry` — four fleet boards that compile the shared `NRF52Board` change

Secrets/private-IP scan clean over the diff and commit messages. (`ADMIN_PASSWORD="password"` / `ROOM_PASSWORD="hello"` are upstream env defaults present in 88 and 68 of 91 variants respectively, overridden at runtime on bench units.)

## Not verified here

Epic verification on hardware end-to-end is [#876](https://github.com/OffbandMesh/meshcore-firmware/issues/876), which stays open through merge. [#965](https://github.com/OffbandMesh/meshcore-firmware/issues/965) (room-server watchdog reset on first boot) is open and unfixed.

## Merge

**Ben merges.** `ci-green` is a required gate and this carries a fleet-shared change.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -253,6 +253,17 @@ jobs:
           - Xiao_S3_WIO_companion_radio_usb
           - Xiao_S3_WIO_companion_radio_wifi
           - Xiao_C6_repeater_
+          # #854 (epic #879): Heltec RadioCore RC52 (nRF52840 + HT-RA62A/SX1262).
+          # Owner-directed 2026-08-21: RC52 goes into CI as it is built out, so the
+          # build is validated, but deliberately NOT into .github/release-envs.txt
+          # until #880 (display) lands and it has had further testing. CI-only here.
+          # (No observer env exists: observer is ESP32/WiFi-only, and nRF52840 has
+          # no WiFi -- same reason Heltec_t096 has none above.)
+          # Envs carry the RC32 _without_display_ convention: this board has no
+          # display support yet (#880), so every shipping env today is headless.
+          - heltec_rc52_without_display_repeater
+          - heltec_rc52_without_display_companion_radio_ble
+          - heltec_rc52_without_display_companion_radio_usb
     steps:
       - uses: actions/checkout@v5
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -270,7 +270,6 @@ jobs:
           - heltec_rc52_without_display_companion_radio_ble_diag
           - heltec_rc52_without_display_companion_radio_usb_diag
           - heltec_rc52_without_display_room_server_diag
-          - heltec_rc52_without_display_sensor_diag
     steps:
       - uses: actions/checkout@v5
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -261,10 +261,15 @@ jobs:
           # no WiFi -- same reason Heltec_t096 has none above.)
           # Envs carry the RC32 _without_display_ convention: this board has no
           # display support yet (#880), so every shipping env today is headless.
-          - heltec_rc52_without_display_repeater
-          - heltec_rc52_without_display_companion_radio_ble
-          - heltec_rc52_without_display_companion_radio_usb
-          - heltec_rc52_without_display_room_server
+          # OWNER DIRECTIVE 2026-08-22 (#889): until a near-release beta, every env
+          # BUILT is a full diagnostic build -- forced caplog + boot beacon + UART
+          # mirror. CI therefore builds the _diag envs, not their plain parents.
+          # The parents still exist so the release-time diag/non-diag pairs are
+          # already there, correctly named, when we get to beta.
+          - heltec_rc52_without_display_repeater_diag
+          - heltec_rc52_without_display_companion_radio_ble_diag
+          - heltec_rc52_without_display_companion_radio_usb_diag
+          - heltec_rc52_without_display_room_server_diag
     steps:
       - uses: actions/checkout@v5
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -270,6 +270,7 @@ jobs:
           - heltec_rc52_without_display_companion_radio_ble_diag
           - heltec_rc52_without_display_companion_radio_usb_diag
           - heltec_rc52_without_display_room_server_diag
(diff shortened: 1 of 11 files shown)


Strycher:

Issue #948, opened 2026-08-23
epic: RC52 display — variant-local driver, no shared-driver changes


Strycher:
Get the RC52's TFT rendering. **Variant-local. No shared-driver changes, no fleet work.**

## The constraint that defines this Epic

`src/helpers/ui/NV3001BDisplay.{h,cpp}` is in production on RC32 and RCC6. Porting it to nRF52 means editing a file every board compiles, regression-testing two shipping boards, and gating RC52 on all of it. **That is not required to put pixels on this panel.**

Instead: an RC52 display driver that lives in `variants/heltec_rc52/`, implements the same `DisplayDriver` interface, and drives SPI1/SPIM3 directly. The shared driver is never opened, so RC32 and RCC6 **cannot** be affected — not "shouldn't be", cannot.

Some code is duplicated from the shared driver. That is the deliberate trade: duplication inside one variant is cheap and reversible; a fleet-wide refactor blocking RC52 is neither.

## Hardware facts to build on

`[verified: vendor variant.h + RC52-L62_V1.02 schematic]`

- TFT is on **SPI1 backed by SPIM3** (`SPI_32MHZ_INTERFACE 1`); LoRa has SPI0. No contention, and **do not** port RCC6's `NV3001B_USE_SOFTWARE_SPI` — that was a workaround for having no free peripheral.
- The Arduino core already instantiates `SPIClass SPI1(_SPI1_DEV, PIN_SPI1_MISO, PIN_SPI1_SCK, PIN_SPI1_MOSI)` from the variant's own pins, and those pins ARE the panel's: `PIN_SPI1_MOSI` == `PIN_TFT_MOSI` (P1.03), `PIN_SPI1_SCK` == `PIN_TFT_SCK` (P0.30). Reference it; do not construct one.
- ⚠ **`spi.transfer(void*, size_t)` is destructive** — it is `transfer(buf, buf, count)` and overwrites the caller's buffer. Use `transfer(tx, nullptr, len)`. Getting this wrong corrupts the back buffer every flush and looks like display noise, not a bus fault.
- ⚠ **TFT occupies carrier header pins 6-10.** `TFT_SCK` P0.30 is also SWDIO, so SWD and the display are mutually exclusive. Owner's rig (RST on pin 18, USER on pin 5) is clear of all five.
- ⚠ **`TFT_EN` is active LOW, `TFT_BL` is active HIGH.** Opposite polarities; one backwards presents as a dead panel.
[...]

Issue #856, opened 2026-08-19
bug: RC52 display back buffer — 56 KB may not allocate, and the fallback path has never run on hardware


Strycher:
RC52 **has** a TFT `[verified: HelTecAutomation/Heltec_nRF52 variants/heltec_rc52/variant.h defines PIN_TFT_MOSI/SCK/VDD_CTL/LEDA_CTL on SPI1]`. That makes an existing, recorded concern live.

## The actual risk

`#747` added a display back buffer. On a 128x220 RGB565 panel that is a **~56 KB** allocation. The nRF52840 has 256 KB RAM total, so on RC52 that allocation is **doubtful** `[recorded in HARDWARE.local.md as applying to RCC6/RC52]`.

**The failure is not "no buffer".** Allocation failure degrades to the unbuffered `drawRGB565` fallback — and **that path has never executed on hardware on any board.** Every board carrying this display has always had the back buffer. RC52 would be the first hardware to run that code.

So the risk is not a missing optimisation; it is landing on an untested code path, silently, depending on heap state at boot.

## Do this

1. **Instrument the allocation.** Log whether the back buffer allocated, at boot, explicitly. Do not infer it from whether the display looks right.
2. If it fails, the fallback is now load-bearing — **verify it renders correctly on hardware**, including the colour splash, before trusting any display work on this board.
3. If it succeeds, record the free-heap headroom it leaves, because 56 KB on a 256 KB part changes what else fits.

Relevant either way: BLE stack and SoftDevice-equivalent RAM reservations on nRF52840 reduce what is actually available well below the nominal 256 KB.

Parent: #625.

Strycher:
Picked up as part of epic 625 bring-up.

Agent: QuietBarn (session 5ef417e8)

Strycher:
Cross-link: the back-buffer question is blocked behind a driver-level issue found during #625 DIAGNOSE — `NV3001BDisplay` does not compile for nRF52 in either SPI mode. Detail and citations on #854.

The narrow point for this issue: `allocFrameBuffer()` (src/helpers/ui/NV3001BDisplay.cpp:469) logs **only on failure**. A successful 56 KB allocation is silent, so "did it allocate" cannot be answered from the log as it stands — which is exactly what this issue asks for. A success-path line is the cheap half.

Agent: QuietBarn (session 5ef417e8)

Strycher:
[...]

PR #987, opened 2026-08-26
epic: RC52 display — variant-local NV3001B driver, indexed palette shipped fleet-wide

Strycher:
Epic [#948](https://github.com/OffbandMesh/meshcore-firmware/issues/948) — RC52 display. Stacks on [#986](https://github.com/OffbandMesh/meshcore-firmware/pull/986) (headless bring-up), now merged.

## What this is

The T108 / NV3001B TFT on the Heltec RadioCore RC52, driven over hardware SPI on SPI1/SPIM3 at 8 MHz. Eight display envs — repeater, room server, companion USB/BLE, each with a `_diag` twin — plus one bench env.

**Zero shared-code changes.** `git diff --name-only` against the base returns nothing outside `variants/heltec_rc52/` and `tools/`. `src/helpers/ui/` is untouched, so **RC32 and RCC6 cannot regress from this** — that was the epic's explicit constraint, not an accident.

## Hardware-verified on `rc52-bench-1`

- Splash renders, **colour art holds for the full splash**, text correct
- Home screen renders — the #973 failure does not reproduce
- Colours correct, no flicker on the repaint cycle
- Repeater, room server and companion roles all render

## Three defects fixed, all mine, all found by the gate or by chasing it

**1. The verified fix was never shipped.** `RC52_DISPLAY_BUFFER_INDEXED` existed on exactly one line — the `_diag_idx` bench env. Every *shipping* env still built the 56,320 B RGB565 buffer that measured **zero** allocatable bytes and made `new HomeScreen` (480 B) fail, which silently stops all rendering ([#973](https://github.com/OffbandMesh/meshcore-firmware/issues/973)). Trimming `MAX_CONTACTS` / `MAX_GROUP_CHANNELS` / `OFFLINE_QUEUE_SIZE` was ruled out, indexed was built instead, confirmed on hardware — and then left in a bench env. Moved to the shared `[Heltec_RC52_with_display]` base so all nine display envs inherit it.

**2. Colour art was erased by the very next blit.** The indexed path wrote art **straight to the panel**, bypassing the buffer, on the reasoning that *"the splash draws once at boot, so the repaint costs nothing."* That was wrong: `SplashScreen::render()` runs **inside** `startFrame()`/`endFrame()` and returns `1000`, so once a second for the whole boot screen — and every `endFrame()` blits the buffer over the entire panel. Observed on hardware as **"the lockup graphic flashed in and out"** while the text below it, which lives in the buffer, stayed put.

Fixed by **compositing**, not bypassing and not quantising. Quantising would posterise the art through a UI palette — the exact thing the bypass existed to avoid — and dropping to the mono lockup is not acceptable on a colour LCD. The art is recorded for the frame and substituted into the blit's **existing** per-pixel unpack loop, which only runs for indexed anyway. Art pixels pass at full RGB565; every other pixel is the same palette lookup as before. One transfer, no flash, no posterisation, and no extra RAM — the source stays in flash and only a pointer and a rect are held. Cleared by `startFrame()`, so it lives exactly one frame. Hoisted behind a null check, so **non-splash frames run the loop byte-for-byte as before**.

**3. `px_nonbg` was a lying instrument.** `endFrame()`'s beacon compared each buffer entry against a hard-coded byte-swapped `uint16_t` RGB565 background. Correct at 16bpp; wrong the moment the buffer holds 8-bit values, since a `uint8_t` index can never equal a `uint16_t` above 255 — so it printed `px_nonbg=28160` (every pixel, 128×220) **on every frame, unconditionally**, reading as a rendering anomaly. Now compares against `pxPack(window_bkg)`, which answers "what would the background be stored as" for all three formats.

## Bench envs

`_8bpp` and `_idx` **removed** (owner decision). The format A/B is finished — indexed won on hardware. With indexed now in the base, `_8bpp` would *inherit* it and silently build **as indexed**, because `pxPack()` tests INDEXED first — turning an A/B into a comparison of indexed against itself. `-U` does not survive PlatformIO (`macro names must be identifiers`, the same failure as the earlier `-U DISPLAY_CLASS` attempt), so opting out isn't available. A `#error` guard is kept as a tripwire against re-adding a conflicting format flag.

`_nobuf` **kept**: it inherits INDEXED harmlessly (`DISABLE_BACKBUFFER` short-circuits `allocFrameBuffer()` before the format is consulted) and remains the only way to exercise the unbuffered draw path, which **has never executed on any board in this fleet** ([#856](https://github.com/OffbandMesh/meshcore-firmware/issues/856)).

- **Finding 1 (High)** — *"back buffer and panel desynchronize in indexed mode."* The stated mechanism (stale blend) is **refuted**: `startFrame()` fills the entire buffer each frame, so no stale region survives. But tracing it found the **real** defect above — the art erased by the blit. Fixed, hardware-verified.
- **Finding 2 (High)** — *"memory optimization not applied to the primary target env."* **Confirmed and fixed.** This was the most valuable finding in the review.
- **Finding 3** — the `pxPack` static memo could go stale if the palette were rebuilt. **Declined**: the palette is built once (`if (s_pal_used) return; // built once`) and never rebuilt, so there is nothing to go stale.
- **Finding 4** — free the back buffer on `turnOff()`. **Declined.** `begin()` re-runs on every `turnOn()`; that is exactly the leak fixed in #856, where each off→on cycle abandoned a buffer and presented as *"8bpp didn't fix the flicker"* when 8bpp was working. The `if (frame_buf) return;` guard is the fix. Free-on-off / realloc-on-on reintroduces a window where a transient allocation failure costs the buffer permanently. Holding 28 KB for the device's life is the deliberate trade.

## Verification

`pio run` **9/9 SUCCESS**, re-run after rebase: all eight display envs plus `_nobuf`. Secrets/private-IP scan clean over the diff and commit messages (`ADMIN_PASSWORD` / `ROOM_PASSWORD` are upstream defaults in 88 and 68 of 91 variants).

## Merge

**Ben merges.** `ci-green` is a required gate.

### DIFF
--- a/tools/diag/rc52-display-951/evidence/951-boot-sessions.md
+++ b/tools/diag/rc52-display-951/evidence/951-boot-sessions.md
@@ -1,0 +1,124 @@
+# RC52 display bring-up — captured boot sessions (#951, #965)
+
+Curated from a continuous Feather-sniffer capture, 2026-08-23 17:39Z → 18:37Z, reading the
+MeshLog UART mirror off carrier header **pin 11** (P0.08, board TX), ground on **pin 20**.
+
+The raw capture was 1,920,700 B and contains binary runs from the DFU cycles. What matters is
+extracted here; the raw file was session-temporary and is not preserved.
+
+**Why the wire and not USB:** on this board `Serial` is USB-CDC, which dies with the chip. Every
+log ever captured that way necessarily comes from a boot that *succeeded*. The wire witnesses the
+boots that fail — which is the entire reason the `_diag` envs set `OFFBAND_LOG_MIRROR_UART`.
+
+---
+
+## Every boot recorded
+
+| Time (UTC) | Role | Reset reason | Display |
+|---|---|---|---|
+| 17:41:46 | repeater | `Soft Reset` (post-DFU) | **dark** — wrong pins (pre-`75ff23dc`) |
+| 17:49:37 | repeater | `Reset Pin` | dark — same image |
+| 18:01:02 | repeater | `Soft Reset` (post-DFU) | ✅ **renders** — first ever, corrected pins |
+| 18:04:07 | repeater | `Reset Pin` | ✅ renders |
+| 18:15:24 | room-server | `Soft Reset` (post-DFU) | ❌ **dark** — see #965 |
+| 18:15:54 | room-server | **`Watchdog`** | ✅ renders |
+| 18:34:35 | room-server | `Reset Pin` | ✅ renders |
+
+Two independent conclusions fall out of that table:
+
+1. **The repeater renders on its first post-DFU boot; the room server does not.** Same driver,
+   same panel, same artifact lineage. So the dark first boot is role-specific, not a driver or
+   a DFU property. (#965)
+2. **Only `Soft Reset` on the room-server role fails.** `Watchdog` and `Reset Pin` both render.
+
+---
+
+## The first successful render (repeater, 18:01:02)
+
+```
+[BEACON] APP:CTOR -- earliest static ctor; bootloader + platform startup COMPLETED
+[BEACON] setup:ENTRY -- before Serial.begin
+[BEACON] setup:before SafeBoot::checkAndMaybeSleep
+[BEACON] setup:post SafeBoot::checkAndMaybeSleep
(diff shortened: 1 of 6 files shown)


Strycher:

Issue #990, opened 2026-08-26
epic: enable the nRF52 diag channel on the bench fleet — per-board mirror pins, proven on the wire


Strycher:
The nRF52 diagnostic channel exists in shared code — [#944](https://github.com/OffbandMesh/meshcore-firmware/issues/944) / [#887](https://github.com/OffbandMesh/meshcore-firmware/issues/887) ported the UART mirror and boot beacon off ESP32-only, merged as [#978](https://github.com/OffbandMesh/meshcore-firmware/pull/978). **This Epic is the other half: actually turning it on, per board.**

Those Epics deliver the capability. Neither delivers a board that can use it — that needs a `_diag` env per board with a mirror TX pin that is free, physically reachable, and proven on the wire. That is a distinct build-and-test unit with a real verification step, which is why it is an Epic rather than more tasks under #944.

## Why this is not trivial config

Every board answers the pin question differently, and the first attempt got **three of four wrong** — each would have built clean, flashed clean, then quietly driven a peripheral (GPS RX on two boards, the I2C clock on a third). Nothing errors.

Two rules came out of that and belong to this Epic:

1. **Read the env, not just `variant.h`.** On these variants `platformio.ini` is where peripherals are actually assigned; `variant.h` only supplies definitions. A pin that looks free in the header can be the GPS UART in the env.
2. **A clean build is not evidence.** A mirror on a wrong or unrouted pin compiles, links and flashes exactly like a correct one. Only the beacon arriving on the wire proves it — and a silent mirror is indistinguishable from a board that never booted.

## Scope

Boards that can be instrumented **without opening a case** (owner constraint, 2026-08-24):

| Board | Status |
|---|---|
| **Heltec T096** | delivered — [#979](https://github.com/OffbandMesh/meshcore-firmware/issues/979) |
| **RAK4631** | written, disabled — [#989](https://github.com/OffbandMesh/meshcore-firmware/issues/989) |
| **RAK3401** | written, disabled — [#989](https://github.com/OffbandMesh/meshcore-firmware/issues/989) |
| **XIAO nRF52** | written, disabled — [#989](https://github.com/OffbandMesh/meshcore-firmware/issues/989) |
| **Heltec T114** | no free pin — split to [#981](https://github.com/OffbandMesh/meshcore-firmware/issues/981) |

Out of scope by the same owner decision: **Wio Tracker Pro** (needs the case opened) and **SenseCAP** (sealed housing).

[...]

Issue #989, opened 2026-08-26
task: validate and enable the three deferred nRF52 diag envs (RAK4631, RAK3401, XIAO nRF52)


Strycher:
Follow-up to [#979](https://github.com/OffbandMesh/meshcore-firmware/issues/979), which delivered the T096 bench diag env. Three boards were written and **deliberately left commented out** because each has an unresolved hardware question. This issue carries those three to completion.

The configuration already exists in-tree, disabled, each block documenting its intended pin, why the alternatives are wrong, and its preconditions. This is validation work, not design work.

## Why they are disabled

All three failed the same way in the first attempt: the pin was read from `variant.h` while the **env** — which is where these boards actually assign peripherals — said something different. Each would have built clean, flashed clean, and quietly driven a peripheral.

## RAK4631 — confirm the pad↔pin mapping

- **Intended:** `WB_IO1` = Arduino pin 17 (P0.17)
- **Blocker:** the RAK19007 base carries a solder pad silkscreened `IO1` `[verified: owner photo, 2026-08-26]`, and `variant.h` maps `WB_IO1` to 17 — but that these are the **same net** is inferred from the WisBlock slot convention, not confirmed against a RAK schematic or measured.
- **Not Serial1:** the env sets `PIN_GPS_RX=PIN_SERIAL1_TX`, and GPS is **active on the repeater** (`ENV_INCLUDE_GPS=1` globally, `RAK_BOARD` set, no `RAK3401_NO_GPS`; `EnvironmentSensorManager.cpp` states *"Repeater (has GPS) keeps the path"*).
- **Not Serial2 (P0.20):** not on the pad header. The pads are `BAT / IO2 / IO1 / AIN1 / TX1 / RX1 / GND / BOOT0 / SDA / SCL / GND / VDD`; everything else is inside the SLOT_C / SLOT_D connectors.
- ⚠ **Not IO2:** `WB_IO2` (P1.02, pin 34) is the **WisBlock slot power switch** — driven HIGH by an early constructor in `RAK4631Board.cpp`, reused as Ethernet power-enable, and toggled by the GPS wake probe.

**To close:** confirm `IO1` reaches P0.17 on this core; confirm no fitted slot module needs IO1; uncomment; flash, wire, and confirm the beacon actually arrives.

## RAK3401 — the pin is not merely unvalidated, it is unknown

- **Intended:** `WB_IO1` — **which this variant does not define at all.**
[...]

Issue #979, opened 2026-08-24
task: bench diag envs for the case-accessible nRF52 boards (RAK4631, RAK3401, XIAO nRF52, T096)


Strycher:
Owner directive 2026-08-24: add bench diag envs for the nRF52 boards that can be instrumented **without opening a case**.

**Bench and CI only — these do NOT ship.** `.github/release-envs.txt` states the rule directly: *"diag envs are diagnostic instruments, not products, and they are not released anywhere it is not explicitly excepted"*, and the RadioCore entries carry *"Do NOT cite these entries as precedent for releasing diag builds elsewhere."* The RadioCore exception exists for beta testers on unproven hardware; these are bench boards, so no release-matrix entry is added.

## Boards in scope

**RAK4631, RAK3401, XIAO nRF52, Heltec T096.**

Out of scope, owner's reasoning:

- **Heltec T114** — deferred to research, no clean mirror pin. See the companion issue.
- **Wio Tracker Pro** — requires opening the case; acceptable only if it becomes necessary.
- **SenseCAP** — requires destructive entry to a sealed housing.

## Dependency — CLEARED

Required the nRF52 UARTE arm from [#944](https://github.com/OffbandMesh/meshcore-firmware/issues/944). **Merged 2026-08-24 as `2d8f8149` via [#978](https://github.com/OffbandMesh/meshcore-firmware/pull/978).** No longer blocked.

## Flags per env

Following the RC52 `_diag` pattern: `OFFBAND_FORCE_CAPLOG`, `OFFBAND_BOOT_BEACON`, `OFFBAND_LOG_MIRROR_UART=1`, `OFFBAND_LOG_MIRROR_TX_PIN=<per board>`, `OFFBAND_LOG_MIRROR_BAUD=115200`.

## Mirror TX pin — per board, NOT copy-pasteable

`[verified: read from each variant.h in-tree]` — physical accessibility still needs bench confirmation.

| Board | `PIN_SERIAL1_TX` | `PIN_SERIAL2_TX` | Determination |
|---|---|---|---|
| RAK4631 | P0.16 | P0.20 | Both appear free (`PIN_GPS_1PPS` is P0.17, unrelated). Confirm which is broken out on the RAK19007 baseboard. |
[...]

PR #992, opened 2026-08-26
task: T096 bench diag env — three deferred boards written and disabled

Strycher:
Task [#979](https://github.com/OffbandMesh/meshcore-firmware/issues/979) under Epic [#990](https://github.com/OffbandMesh/meshcore-firmware/issues/990).

## What ships: one board

**`Heltec_t096_repeater_diag`** — the T096 repeater plus the diag channel (boot beacon, UART mirror, forced caplog), so a boot that fails can be witnessed on a wire instead of a USB console that dies with the chip.

`pio project config` lists **exactly one** new env. RAK4631, RAK3401 and XIAO nRF52 are written and **deliberately commented out**.

## Three of the four original pins were wrong

The first commit picked a mirror TX pin per board from `variant.h`. Three collided with peripherals, and every one of them would have **built clean, flashed clean, and then quietly driven something**:

| Board | Original pin | Collided with |
|---|---|---|
| RAK4631 | `PIN_SERIAL1_TX` | env sets `PIN_GPS_RX=PIN_SERIAL1_TX` — Serial1 **is** the GPS UART, and GPS is active on the repeater |
| T096 | `PIN_SERIAL1_TX` | env sets `PIN_GPS_RX=25`; this board has a **built-in GPS** |
| XIAO | `PIN_SERIAL1_TX` = D6 | `PIN_WIRE_SCL=D6` — the I2C clock |
| RAK3401 | `PIN_SERIAL2_TX` | free in firmware, but **not on the RAK19007 pad header** — unreachable |

**Root cause:** pins were read from `variant.h` and the **env** was never read. On these variants `platformio.ini` is where peripherals are actually assigned. The original `[verified: read from each variant.h]` tag was true about what was read and false about what it proved.

## T096's replacement pin, and why

**`PIN_SERIAL2_TX` (P0.10)**, on three independent lines:

1. **Unclaimed** — all 29 pins assigned across `variant.h` and the env were mapped; 9/10 appear nowhere else.
2. **This board's own `Heltec_t096_repeater_bridge_rs232` env already drives P0.09/P0.10 as an external RS232 UART**, which implies Heltec routes them out and they work as a UART.
3. **The NFC objection is disproven on our own hardware.** P0.09/P0.10 are NFC2/NFC1 and normally need `CONFIG_NFCT_PINS_AS_GPIOS`, defined nowhere in tree or framework — but RC52, same nRF52840 and same Adafruit core, drives its TFT reset and backlight on those pins with no UICR change. Serial2 was originally rejected on a theory our own bench had already falsified.

⚠ The comment states plainly that **P0.10 is the MCU pin, not a silkscreen label**, and that which physical pad carries it is **not established** — per owner direction, because knowing the MCU pin does not tell anyone where to land a wire.

⚠ Do **not** add the diag block to `Heltec_t096_repeater_bridge_rs232`; it owns those pins.

## The three disabled boards

Written, inert, each documenting its intended pin, why the alternatives are wrong, and its preconditions. Carried by [#989](https://github.com/OffbandMesh/meshcore-firmware/issues/989).

- **RAK4631** → `WB_IO1` (P0.17). `IO1` **is** a pad on the base `[verified: owner photo]`; that the pad and pin 17 are the same net is inferred from the WisBlock convention, not measured. Not IO2 — that is the **slot power switch**.
- **RAK3401** → `WB_IO1`, **which this variant does not define**. Left referencing the undefined symbol so the gap is visible rather than papered over with a guess. Probably 17 (same base; the two variants already agree on `WB_IO2`, spelled `PIN_3V3_EN(34)` and `WB_IO2=34`), but different cores are not obliged to match. Not IO2 — here it is the **PA's 5 V boost enable**.
- **XIAO** → `D0`, which requires the repeater to release `PIN_USER_BTN`. All 11 signal pins are assigned — five to the radio, three to SPI, two to I2C, one to the button — so D0 is the only candidate, and freeing it is a behaviour change to a shipping role. Owner decision.

## Also fixed: a pin designation that does not exist

Three RAK3401 files described `WB_IO2` / `PIN_3V3_EN` as **`P0.34`**. There is no such pin — nRF52840 is P0.00–P0.31 then P1.00–P1.15, so Arduino pin 34 is **P1.02**, exactly as our own `RAK4631Board.cpp` states for the same pin. Corrected in `variant.h`, `RAK3401Board.cpp` and the disabled diag block, each noting what it used to say. Comment-only; the pin *number* was always right.

That line is a warning about the PA's 5 V boost enable on a board with a documented GPS-vs-PA power conflict, so a wrong designation there is the kind of thing that gets acted on.

## Not released

Bench and CI instruments only. `.github/release-envs.txt` is the authority: diag envs *"are not released anywhere it is not explicitly excepted,"* and the RadioCore entries carry *"Do NOT cite these entries as precedent for releasing diag builds elsewhere."* Zero `*_diag` entries added.

## Review gate

First pass found the pin collisions — **it was right on every count** and the change was reworked around it. Re-run against the rework came back sound on all five questions asked: P0.10 safe on the repeater, the NFC argument holds, the disabled blocks are genuinely inert, the diag env differs from its sibling only in diag flags, and the disabled-block documentation is accurate. One trivial finding — the `P0.34` designation — fixed above.

## Verification

`pio run` **6/6 SUCCESS** after rebase: `Heltec_t096_repeater_diag`, `Heltec_t096_repeater`, `Heltec_t096_repeater_bridge_rs232` (owns the same pins), plus `RAK_4631_repeater`, `RAK_3401_repeater`, `Xiao_nrf52_repeater` — the three prod repeaters whose files were edited.

`envdump` confirms the diag build carries `OFFBAND_LOG_MIRROR_TX_PIN=PIN_SERIAL2_TX` and the prod build carries **zero** mirror flags. Beacon markers present in the diag ELF, so the instrumentation is compiled in and not merely flagged. Secrets/private-IP scan clean.

## Not verified here

**The T096 mirror has not been observed on the wire.** A clean build is not evidence — a mirror on a wrong or unrouted pin compiles and flashes identically to a correct one. Hardware confirmation is Epic verification under [#990](https://github.com/OffbandMesh/meshcore-firmware/issues/990).

## Merge

**Ben merges.**

### DIFF
--- a/variants/heltec_t096/platformio.ini
+++ b/variants/heltec_t096/platformio.ini
@@ -176,3 +176,35 @@ lib_deps =
 extends = Heltec_t096
 build_src_filter = ${Heltec_t096.build_src_filter}
   +<../examples/kiss_modem/>
+
+; =============================================================================
+; BENCH DIAG ENVS (#979) -- NOT RELEASED.
+;
+; Diag envs are diagnostic instruments, not products. `.github/release-envs.txt`
+; is the authority: they ship ONLY where explicitly excepted, and the RadioCore
+; entries there carry "Do NOT cite these entries as precedent for releasing diag
+; builds elsewhere." These are bench/CI only and are deliberately ABSENT from the
+; release matrix.
+;
+; Owner-directed 2026-08-24: the nRF52 boards that can be instrumented without
+; opening a case. Wio Tracker Pro (needs the case opened) and SenseCAP (sealed
+; housing) are out of scope by the same decision.
+; =============================================================================
+
+[heltec_t096_diag]
+build_flags =
+  -D OFFBAND_FORCE_CAPLOG
+  -D OFFBAND_BOOT_BEACON
+  -D OFFBAND_LOG_MIRROR_UART=1
+  ; P0.25 -- Serial1. MUST NOT be Serial2: PIN_SERIAL2_TX is P0.10, an nRF52840 NFC
+; antenna pin, which needs CONFIG_NFCT_PINS_AS_GPIOS (set by no variant in this tree).
+  -D OFFBAND_LOG_MIRROR_TX_PIN=PIN_SERIAL1_TX
+  -D OFFBAND_LOG_MIRROR_BAUD=115200
+
+; Heltec_t096_repeater_diag -- the shipped role plus the diag channel, nothing else. A diag
+; build that diverges from the shipped build diagnoses the wrong firmware.
+[env:Heltec_t096_repeater_diag]
+extends = env:Heltec_t096_repeater
+build_flags =
+  ${env:Heltec_t096_repeater.build_flags}
+  ${heltec_t096_diag.build_flags}
--- a/variants/heltec_t096/platformio.ini
+++ b/variants/heltec_t096/platformio.ini
@@ -196,13 +196,39 @@ build_flags =
   -D OFFBAND_FORCE_CAPLOG
   -D OFFBAND_BOOT_BEACON
   -D OFFBAND_LOG_MIRROR_UART=1
-  ; P0.25 -- Serial1. MUST NOT be Serial2: PIN_SERIAL2_TX is P0.10, an nRF52840 NFC
-; antenna pin, which needs CONFIG_NFCT_PINS_AS_GPIOS (set by no variant in this tree).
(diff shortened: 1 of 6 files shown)


Strycher:

Issue #997, opened 2026-08-26
task: measure true load current via the INA228 charge accumulator — 30 s sampling misses TX bursts


Strycher:
Follow-up to [#991](https://github.com/OffbandMesh/meshcore-firmware/issues/991), which delivered the INA228 decode and proved it on hardware. That decode reports **instantaneous** current sampled every 30 s. This issue is about measuring what the node actually *draws*.

## Why a 30-second sample is the wrong instrument for this load

`[verified: rc52-bench-1, 2026-08-26]` the decode reads a steady **9.7 mA** with the node idle — nRF52840 plus SX1262 in RX, display off.

**That is idle draw, not average draw.** A LoRa node's consumption is dominated by short transmit bursts: an advert is ~150 ms at a TX current many times the idle figure. Sampling every 30 s lands on a burst essentially never, so the reported number is the floor, not the mean. Any runtime or battery-life estimate built on it will be optimistic, and the error grows with traffic.

This is the same class of measurement error that produced wrong RC52 TX-power figures — a sweep that visited the carrier ~3% of the time and under-reported because it kept missing the bursts. Same failure, different instrument.

## What the part already provides

The INA228 has a **hardware CHARGE accumulator** (`0x0A`) — a 40-bit signed register that integrates current continuously in silicon. It cannot miss a burst, because it is not sampling: it accumulates between reads. Divide accumulated charge by elapsed time and the result is true average current over that window, bursts included.

There is also an `ENERGY` register (`0x09`, 40-bit unsigned) if power rather than current is wanted.

## Why this was scoped out of #991

Deliberately, not overlooked:

- **`CHARGE` requires `SHUNT_CAL` (0x02) to be written.** Unlike `VSHUNT`, it is not a raw reading — the on-chip integration depends on the calibration register being correct. #991's whole design was *no calibration write*, on the reasoning that reading raw and dividing on the host cannot be half-configured and survives a part reset. Reading CHARGE means giving that up and owning the calibration properly.
- **40-bit reads** need another helper; the existing ones are 16- and 24-bit.
- Doing it in the same commit would have been half-done work bolted onto a verified decode.

## What this needs

[...]

Issue #991, opened 2026-08-26
bug: sniffer identifies the INA228 but has no decode path — no flash reversion needed to explain it


Strycher:
An INA228 was received and wired into the Feather sniffer rig. **What is flashed is not reading it.** GreenTower's read was that a flash reversion occurred. The owner is not convinced, on the grounds that **the sketch was never tested against an INA228 in the first place** — so there is no known-good state to have reverted from.

## The owner is right, and it is provable from the source

`[verified: tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino on origin/firmware-base, read 2026-08-26]`

**`INA_228` appears in exactly two places in the whole sketch:**

1. the `InaKind` enum declaration (line ~280)
2. one identification line: `ina_kind = (part == 0x228) ? INA_228 …` (line ~381)

**There is no decode path.** `inaTick()` opens with:

```c
if (ina_kind != INA_219_CLASS || ina_addr == 0) return 0;
```

So a correctly-identified INA228 is recognised and then **deliberately not read** — the function returns before touching a register. The sketch's own header says so: *"Deliberately identification-only: the current decode is written once the part is known, rather than shipping three decoders that never run and cannot be tested."*

**No flash reversion is required to explain the symptom.** The firmware has never been able to read an INA228. That theory should not be pursued until something else supports it.

## What the live probe is actually reporting

From the capture log:

```
=== INA candidate at 0x40  raw 0xFE=0x5959 0xFF=0x2719 0x3E=0x5959 0x3F=0x2719
    -> no TI mfg ID (INA219-class)
```

Two things worth noting:

- **A genuine INA228 answers `0x3E` with `0x5449` ("TI").** This part does not. So whatever is at 0x40 is not being seen as an INA228.
- **Both register pairs return identical values** (`0x5959` / `0x2719` at 0xFE/0xFF *and* 0x3E/0x3F). Reading the same data at two unrelated addresses is the signature of a device that does not implement those registers and wraps/aliases — not of a part reporting an ID.

So the open question is not only "why is it not decoded" but **"what is answering at 0x40, and is the INA228 on the bus at all?"** Candidates: the INA228 is at a different strap address, is unpowered, is not on the `Wire1` bus the gauge uses, or the original INA219-class part is still fitted and answering.

## ⚠ Consequence if the INA228 *is* the part at 0x40

[...]

PR #998, opened 2026-08-26
task: INA228 decode for the sniffer — identified but never read

Strycher:
Task [#991](https://github.com/OffbandMesh/meshcore-firmware/issues/991) under Epic [#990](https://github.com/OffbandMesh/meshcore-firmware/issues/990). Bench tooling only — `tools/diag/sniffer/`, one Arduino sketch. **No firmware change; no shipped env is touched.**

## The problem

An INA228 replaced the INA219-class current monitor on the Feather sniffer rig. The sketch **identified it correctly and then never read it** — `inaTick()` opened with `if (ina_kind != INA_219_CLASS ...) return 0;`, so a recognised INA228 was silently ignored. No `[ina]` lines at all.

**There was no flash reversion**, and the log rules one out: the *same* firmware reported `no TI mfg ID (INA219-class)` through 2026-08-25 08:06 and `mfg=0x5449 dev=0x2281 -> INA228` at 23:59:26. The part changed, not the image.

## What this adds

- `gaugeRead24At()` — 24-bit register read, modelled on the existing 16-bit helper
- `ina228Tick()` — reads `VSHUNT` (0x04) and `VBUS` (0x05), emits the **same `[ina]` line shape** as the INA219 path so existing captures, greps and parsers keep working, plus an `adcrange=` field
- `inaTick()` dispatches by part instead of early-returning

## Three deliberate choices

**No calibration write.** `SHUNT_CAL` only feeds CURRENT/POWER/ENERGY/CHARGE. This reads `VSHUNT` and divides by a known resistance on the host: no magic constant to get wrong, cannot be half-configured, and survives a part reset that would silently invalidate a written calibration. Same reasoning the INA219 path already used.

**The shunt is a property of the part, not the rig.** `INA228_SHUNT_MILLIOHM 15` (Adafruit INA228 breakout carries `R015`), kept separate from the INA219 path's `100` (that module carried `R100`). Mixing them under-reports by 6.67× and looks entirely plausible doing it. A previous note recorded `R100` as a rig-level fact with no part attached — it was only ever confirmed for the INA219, and generalising it made a swapped part look like a firmware defect.

**ADCRANGE is read from CONFIG, not assumed** — a 4× scale difference (312.5 vs 78.125 nV/LSB). Reading it every tick means the decode follows if anything ever configures the part; a failed read falls back to the part's reset default, the coarser range, so it under-resolves rather than over-reports.

## Hardware-verified on `rc52-bench-1`

`[verified: owner-operated bench, 2026-08-26 05:30–05:52Z]`

**Bus voltage, cross-checked against an independent part on the same cell:**

```
[ina]   bus_mv=4164  raw_bus=0x53480
[gauge] mv=4163
```

Recomputed by hand: `0x53480` → `>>4` = 21,320 → × 195.3125 µV = **4164.06 mV**, exactly as printed — and within 1–2 mV of the MAX17048.

**Sign, confirmed by controlled state change rather than assumption:**

| Time | `ma` | State |
|---|---|---|
| 05:49:37 | 0.0 | USB connected, battery full |
| **05:50:07** | **10.1** | **USB unplugged** |
| 05:50:37 → 05:52:07 | 9.7 ×4 | steady on battery |

**Positive = discharge.** `bus_mv` stepped 4165 → 4161 at the same moment, gauge tracking exactly. Noise floor measured at ±7 LSB ≈ **±0.15 mA**.

## ⚠ The old rig's numbers were wrong by an order of magnitude

Idle draw measures **9.7 mA**. The previous INA219-class setup reported **117–182 mA** for the same board. 9.7 mA is what an idle companion should draw — nRF52840 plus SX1262 in RX, display off. 150 mA was never plausible.

**Treat every current figure taken before this commit as suspect**, including the "charging, constant-voltage tail" reading of 2026-08-25. The *direction* was independently corroborated by SOC rising; the magnitudes were not.

## Review gate

Register numbers and LSB values **verified against the TI datasheet (SBOS855F)** — `CONFIG` 0x00, `VSHUNT` 0x04, `VBUS` 0x05, ADCRANGE bit 4, 312.5/78.125 nV, 195.3125 µV. Also confirmed sound: the int64 intermediates, the 20-bit sign extension, the no-calibration approach, and that the dispatch change does not disturb the INA219 path.

**Fixed** — unchecked `Wire.read()`: it returns `int` and yields −1 on an empty buffer, which an implicit `uint8_t` cast turns into `0xFF`. The gate called the result "a large, fictitious current"; that is only half right, and the wrong half is the dangerous one. An all-`0xFF` VBUS decodes to ~204.8 V, obvious nonsense someone would catch — but an all-`0xFF` VSHUNT sign-extends to −1 LSB and reads as **−0.02 mA**, entirely unremarkable. Applied to `gaugeRead16At` too, which carries the identical defect and feeds every gauge reading.

**Declined** — caching ADCRANGE at identification instead of reading it per tick. Reading it is the point: a cached value would silently keep scaling by the wrong factor of 4 if anything reconfigured the part. It saves one 16-bit read per 30 s.

## Scoped out on purpose

The INA228's hardware **charge accumulator** — the feature that makes a runtime figure honest when the load is bursty, which is exactly a LoRa node's TX profile. **The 9.7 mA above is a 30-second sample, so it is idle draw, not average draw under traffic.** It needs `SHUNT_CAL` written and 40-bit reads, so it is [#997](https://github.com/OffbandMesh/meshcore-firmware/issues/997) rather than half-done here.

## Verification

`arduino-cli compile --fqbn esp32:esp32:adafruit_feather_esp32s3` — **395,472 bytes (13%), no warnings.** Secrets/private-IP scan clean. Rebased onto `firmware-base`, 0 behind.

CI's PlatformIO matrix does not compile `.ino` sketches, so it exercises nothing here — that is why the compile was run explicitly and the result stated above.

## Merge

**Ben merges.**

### DIFF
--- a/tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino
+++ b/tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino
@@ -209,6 +209,45 @@
 // the chip, and no averaging setting recovers it.
 #define INA219_SHUNT_MILLIOHM  100
 
+// ---------------------------------------------------------------------------
+// INA228 (#991). Different part, different register map, DIFFERENT SHUNT.
+//
+// THE SHUNT IS A PROPERTY OF THE FITTED PART, NOT OF THE RIG. The INA219 bench
+// module carries R100 = 0.100 ohm; the Adafruit INA228 breakout carries
+// R015 = 0.015 ohm [verified: owner photo of the fitted board, 2026-08-26].
+// Reusing the INA219 constant here would under-report by 100/15 = 6.67x and look
+// entirely plausible doing it.
+//
+// NO CALIBRATION WRITE, deliberately -- same reasoning as the INA219 path above.
+// SHUNT_CAL only feeds the CURRENT/POWER/ENERGY/CHARGE registers. Reading VSHUNT
+// and dividing by a known resistance on the host needs no magic constant, cannot
+// be half-configured, and survives a part reset that would silently invalidate a
+// written calibration.
+//
+// VSHUNT and VBUS are 24-bit registers holding 20-bit data LEFT-ALIGNED in bits
+// 23:4, so both are >>4 before scaling. VSHUNT is SIGNED (two's complement over
+// 20 bits); VBUS is unsigned.
+//
+// The VSHUNT scale depends on ADCRANGE (CONFIG bit 4), so it is READ rather than
+// assumed -- if anything ever configures this part, the decode follows instead of
+// silently scaling by 4x:
+//     ADCRANGE=0   312.5   nV/LSB  ->  +/-163.84 mV  ->  +/-10.9 A at 15 mohm
+//     ADCRANGE=1    78.125 nV/LSB  ->  +/- 40.96 mV  ->  +/- 2.73 A at 15 mohm
+// At R015 that is 20.83 uA or 5.21 uA per count. A LoRa node's draw sits well
+// inside either range, so resolution is the only thing at stake.
+// ---------------------------------------------------------------------------
+#define INA228_SHUNT_MILLIOHM  15
+
+#define INA228_REG_CONFIG   0x00
+#define INA228_REG_VSHUNT   0x04
+#define INA228_REG_VBUS     0x05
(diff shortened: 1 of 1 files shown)


Strycher:

Issue #889, opened 2026-08-20
task: every RC52 env is a _diag build — forced caplog, beacon, mirror, captured end to end


Strycher:
Bring RC52's diagnostic surface to the family bar, and prove it captures end to end against the WROOM sniffer.

**Rescoped 2026-08-22** to the owner directives recorded on [#625](https://github.com/OffbandMesh/meshcore-firmware/issues/625). The original scope — one BLE-companion diag env — was a regression below both predecessors and is superseded.

## The directives this implements

1. **Until beta:** every environment built is a full diagnostic build — caplog forced on and the boot beacon on. Not "a diag variant exists alongside"; during bring-up there is no plain build to reach for.
2. **At beta/stable:** the matrix ships diag **and** non-diag versions of every build we issue. Companion, observer, repeater, room_server, sensor — each issued role gets two.
3. **Naming:** the suffix is **`_diag`**, the RCC6 vocabulary, on every board. `_testdiag` is not acceptable.

## The bar — the union of both predecessors, not either one

`[verified: variants/heltec_rc32/platformio.ini, variants/heltec_rcc6/platformio.ini:327-330, read 2026-08-22 — resolved through `extends` / `${...}`, not a grep of direct `-D` lines]`

| Board | Diag envs | Carries |
|---|---|---|
| RC32 | 6 | boot beacon, `OFFBAND_MESHLOG_UART0`, power telemetry, a CrashLog-disabled A/B twin |
| RCC6 | 4, via a shared `[rcc6_diag]` block | `OFFBAND_MESHLOG_UART0` + **`OFFBAND_FORCE_CAPLOG`** |
| RC52 today | **1** | mirror + beacon, **no `FORCE_CAPLOG`**, no repeater, no USB |

`OFFBAND_FORCE_CAPLOG` exists because the RC32 bring-up had been blind. Its rationale, beside the flag: without it the boot path is mute on exactly the boot under investigation, because every `MESH_DEBUG_PRINTLN` is gated on `g_meshLogEnabled`, which stays false until prefs load. `caplog_enabled` defaults to `0` `[verified: src/helpers/CommonCLI.h:87]`.

## Scope

**A shared `[rc52_diag]` block**, mirroring `[rcc6_diag]`:

- `OFFBAND_FORCE_CAPLOG`
- `OFFBAND_BOOT_BEACON`
[...]

Issue #879, opened 2026-08-20
epic: RC52 bring-up — variant, FEM, battery, diagnostics, display, all roles


Strycher:
Bring the Heltec RadioCore RC52 (nRF52840 + HT-RA62A / SX1262 + FEM) up as a working **headless** board — radio, power and boot proven on hardware. No display; that is the sibling epic.

Created 2026-08-19 when the owner restructured **#625 from an Epic to a Feature**. These children previously hung directly off #625, which skipped the Epic level.

## Deliverable

An RC52 that boots reliably, joins the mesh, reports battery honestly, and has its FEM characterised — in the repeater / room-server / BLE-companion / USB-companion roles. **One branch, one PR, at epic completion.**

## Children

| Issue | Scope | State |
|---|---|---|
| #853 | Registry bootstrap — `rc52-bench-1`, identity by serial | **closed** |
| #854 | Headless variant scaffold — `variants/heltec_rc52` + board JSON | code complete, unpushed |
| #855 | CrashLog boot hang — the RC32 fault is not ruled out on nRF52 | open |
| #857 | Battery path confirmed on hardware — 4.9 / HIGH are derived, not measured | open |
| #858 | FEM bring-up — `FEM_EN`, `VFEM_CTRL`, RXEN on the LNA line | open |

## Branch

`epic/879-rc52-headless` — the single branch for all of the above. Task commits go on it; **no PR until this epic's integration test passes.**

## Known state carried in from #854

Derived and cited, safe to build on:

- `VFEM_CTRL` is **active HIGH** — drives EN on U14, a TLV75733PDBVR. `[verified: RC52-L62_V1.02 schematic]`
- `NRF52BoardDCDC` is correct — L2/L4 are 10µH/80mA against the nRF52840 `DCC`/`DCCH` pins. `[verified: same]`
- Battery `4.90` / `ADC_CTRL_ENABLED HIGH` — schematic divider `(390K + 100K)/100K`, independently corroborated by the vendor BSP's `BAT_AMPLIFY 4.9F`.
- There is **no LED** on this board (`PIN_LED1 -1`). No status-LED heartbeat is possible — do not diagnose its absence as a fault.

Unverified, and tagged as such in the source — **do not inherit these as fact**:

- Whether `FEM_EN` floats without a pulldown, and whether floating costs anything on this module. `[hypothesis: untested]` — no HT-RA62A datasheet exists.
[...]

Issue #876, opened 2026-08-19
task: Epic verification — RC52 bring-up end to end on hardware


Strycher:
The **final task of epic #879** (RC52 headless bring-up) and its acceptance gate.

> An epic is NOT complete just because all its child tasks are closed. […] **Epic integration test.** End-to-end validation that all tasks delivered under this epic work together cohesively. This is not a re-run of unit tests — it is a verification that the combined deliverables function as a system. **The final task in every epic must be scoped as this integration test.** — CLAUDE-BASE, Epic Completion Protocol

Gates the epic's single PR: *"no PR is created until the epic integration test passes."*

**Rescoped 2026-08-19:** originally written against #625 when that was an Epic. #625 is now a **Feature**, so this covers the **headless** epic only. Display integration is tracked separately under #880.

## What is being integrated

| Child | Contributes |
|---|---|
| #853 | RC52 in the device registry, identity-resolvable for flashing |
| #854 | Headless variant — LoRa, FEM pins, battery, four roles |
| #855 | CrashLog boot behaviour settled |
| #858 | FEM bring-up — `FEM_EN`, `VFEM_CTRL`, RXEN |

## Acceptance criteria

Run on **real RC52 hardware**, against a **release-shaped image built from a tag** — not a bench build. (Bench builds git-describe off the previous tag and misreport their version; only the tagged image self-identifies correctly.)

1. **Boot reliability.** Boots and stays up across ≥10 consecutive cycles covering power-on, RST, **fast replug** and **slow replug**. The fast/slow split is not optional — it is the exact axis on which the RC32 CrashLog fault appears and clears (#855).
2. **Radio, end to end.** Joins the mesh, both **sends and receives** against a known-good peer. Repeater role forwards. TX power **measured at the antenna** and consistent with what #858 established — never assumed from the config value, because the FEM's gain is not published.
3. **Battery telemetry sanity — not a runtime burn.** Reported millivolts are plausible and track a meter at one or two points; divider confirmed **off at idle** (no standing drain). The measured discharge-to-shutdown curve is **#857**, which moved to epic **#929** on 2026-08-21 — owner-directed, and explicitly *not* a gate on this epic: *"I'm not gating #879 closure on battery runtime burn."*
4. **Companion client.** Real Offband client pairs over BLE, exchanges traffic, reports a sane device-info string.
5. **Config survives an app-level DFU re-flash.** Identity and settings intact.
[...]

PR #1003, opened 2026-08-26
task(#889): RC52 BLE companion diag was missing power telemetry

Strycher:
Task [#889](https://github.com/OffbandMesh/meshcore-firmware/issues/889) under Epic [#879](https://github.com/OffbandMesh/meshcore-firmware/issues/879). One flag, one env.

## The gap

RC52 carried `OFFBAND_POWER_TELEMETRY` on **zero** envs. RC32 and RCC6 each carry it on exactly one — their BLE companion diag. #889's bar was explicitly *"the union of both predecessors, not either one"* and named power telemetry in RC32's column, so RC52 was short of it. This was missed when the rest of #889's structural work landed.

## Scoped to one env, matching both predecessors

**Not** added to the shared `[rc52_diag]` block, deliberately. RC32 records why it is kept off the boot-diagnosis builds:

> a battery line every 30 s is noise when the question is "why will it not boot", and the tester has no battery instrumentation rig

So it belongs on the burndown env and nowhere else, which is where RC32 and RCC6 put it.

## Why it is load-bearing

[#857](https://github.com/OffbandMesh/meshcore-firmware/issues/857) defines runtime as *(last `[pwr]` timestamp − first on-battery `[pwr]` timestamp)*. Without this flag an RC52 burn cannot be measured the way RC32 (**21 h 34 m 31 s**) and RCC6 (**43 h 21 m 55 s**) were, and a number produced any other way is not comparable to either.

## Verified on the wire, not merely built

`[verified: rc52-bench-1, 2026-08-26 06:51Z]`

```
[4964]  [pwr] mv=4152 up=4s
[34964] [pwr] mv=4170 up=34s
[64964] [pwr] mv=4163 up=64s
```

Correct format, 30 s cadence, arriving over the UART mirror. Boot clean through `setup:DONE`, beacons intact.

## #889's other open criterion, closed by control

The last acceptance line was *"caplog output, on a role that is not the BLE companion."* Proven by building the identical env **without** `FORCE_CAPLOG`, keeping beacon and mirror:

| | with `FORCE_CAPLOG` | control |
|---|---|---|
| `[BEACON]`, `[boot]`, `[rc52disp]`, `[gauge]`, `[ina]`, `[xchk]` | ✅ | ✅ |
| **`DEBUG:` MeshLog stream** | **✅ from [4810]** | **❌ none** |

The `DEBUG:` stream vanished entirely while everything else stayed — so `FORCE_CAPLOG` is what puts it there, and this board does **not** carry `caplog_enabled` persisted, which was the alternative explanation. The control env was temporary, removed, and never entered a commit (worktree verified clean, `pio project config` shows zero occurrences).

## Review gate

**Taken** — the instrument costs power, and its cost is not the same on every board. The flag wakes the CPU, enables the divider, takes an ADC read and logs, so every measured runtime is shorter than the shipped firmware's would be. Worse, RC52 is nRF52840 while RC32/RCC6 are ESP32 — different ADC architecture, different wake and sample costs, so the overhead is **not guaranteed equal** across the three. Recorded in the comment, tagged `[hypothesis: untested]`, with the instruction to measure the overhead before believing any runtime difference that lands within a few percent. Also stated plainly: this is not a shipped-build runtime and must never be quoted to a user as one.

**Declined** — *"also add it to `_usb_diag`."* That would go **beyond** the predecessors, not match them: RC32's USB testdiag block says in terms *"Deliberately NOT included: `OFFBAND_POWER_TELEMETRY`."* Following the precedent means keeping it off the USB diag. The burndown methodology for both existing boards is BLE companion, and a USB-role number would not be comparable to either.

Clean on the other three questions checked: no effect on the shipped non-diag env, cadence consistent at 30 s, and no factual errors in the comment.

## Verification

`pio run -e heltec_rc52_companion_radio_ble_diag` **SUCCESS**. `envdump` confirms `OFFBAND_POWER_TELEMETRY`, `OFFBAND_FORCE_CAPLOG`, `OFFBAND_BOOT_BEACON` and `OFFBAND_LOG_MIRROR_UART` all present. Secrets/private-IP scan clean. Rebased onto `firmware-base`, 0 behind, single commit.

## Merge

**Ben merges.** Epic closure remains a human-only gate — [#876](https://github.com/OffbandMesh/meshcore-firmware/issues/876) reports, the owner decides.

### DIFF
--- a/variants/heltec_rc52/platformio.ini
+++ b/variants/heltec_rc52/platformio.ini
@@ -286,6 +286,43 @@ extends = env:heltec_rc52_companion_radio_ble
 build_flags =
   ${env:heltec_rc52_companion_radio_ble.build_flags}
   ${rc52_diag.build_flags}
+  ; -------------------------------------------------------------------------
+  ; POWER TELEMETRY -- this env only, matching RC32 and RCC6 (#889).
+  ;
+  ; Emits `[pwr] mv=... up=...s` every 30 s. It is NOT part of the shared
+  ; [rc52_diag] block on purpose: RC32 records why it is kept off the boot-
+  ; diagnosis builds -- "a battery line every 30 s is noise when the question is
+  ; 'why will it not boot'". So it belongs on the BLE companion diag and nowhere
+  ; else, which is exactly where RC32 and RCC6 put it:
+  ;   heltec_rc32_companion_radio_ble_diag  -D OFFBAND_POWER_TELEMETRY
+  ;   heltec_rcc6_companion_radio_ble_diag  -D OFFBAND_POWER_TELEMETRY
+  ;
+  ; WHY IT IS LOAD-BEARING, not decoration: #857's runtime burn defines runtime
+  ; as (last [pwr] timestamp - first on-battery [pwr] timestamp). Without this
+  ; flag the burn cannot be measured the way RC32 (21 h 34 m 31 s) and RCC6
+  ; (43 h 21 m 55 s) were, and a number produced any other way is not comparable
+  ; to either -- same cell, same cutoff, same role, same instrument, or it does
+  ; not go in the table.
+  ;
+  ; This was the last gap against #889's "union of both predecessors" bar. RC52
+  ; had the beacon, the mirror and forced caplog; power telemetry was missed.
+  ;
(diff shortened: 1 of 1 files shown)


Strycher:

Issue #1004, opened 2026-08-26
epic: RC52 battery runtime characterisation


Strycher:
Shippable deliverable: a measured RC52 runtime figure, directly comparable to the RC32 baseline (#833) and the RCC6 run (#903).

## Why this is its own Epic

#833 measured RC32 (ESP32-S3) at **21 h 34 m / ~116 mA average**. The INA228 on the bench has already caught RC52 on battery at **~10–16 mA** — a 7–10x difference that, if it holds, is the single most useful number this board produces. It is also a week-long run rather than an overnight one, so it needs its own tracking rather than riding on the bring-up epic.

## Prerequisites — both are real blockers, not incidental fixes

| | |
|---|---|
| Cutoff | `AUTO_SHUTDOWN_MILLIVOLTS` is **absent** on RC52. Without it the shutdown path compiles out, the run has no defined endpoint, and the cell is pulled below safe voltage instead of the board stopping itself. Same blocker #878 fixed for RCC6. |
| Telemetry | `OFFBAND_POWER_TELEMETRY` (#889) is what emits the `[pwr]` line. It must be on `firmware-base` so the burn image is reproducible from a branch rather than from an unmerged worktree. |

## Method — matched to #833 so the numbers are comparable

Same 2500 mAh cell, same 3400 mV cutoff, same role (**BLE companion, display active**), same external instrumentation: Feather UART sniffer plus MAX17048, now with an INA228 adding load current that #833 did not have.

Runtime is defined as #833 defines it — *last `[pwr]` timestamp − first on-battery `[pwr]` timestamp*. T0 is the USB unplug, identifiable in the log as `[ina] ma` going positive.

## Known uncontrolled differences — state these with the result

- **MCU family.** nRF52840 + DC/DC (both inductors fitted) against RC32's ESP32-S3. This is most of what the comparison measures.
- **Diag build.** The burn image is a `_diag` env, so the UART mirror is streaming continuously. #833 was also `_ble_diag`, so this is matched rather than controlled.
- **Display state must be pinned at T0.** A 45.7 mA sample on 2026-08-25 against ~10 mA on 2026-08-26 suggests display state varies. #833 was explicitly "display active".

## Definition of done

Every criterion has a recorded result — pass, fail, or N/A-with-reason. A measured runtime, the discharge curve, the average current, and a written comparison against #833 and #903, in `docs/radiocore/measurements/`.

Epic closure is a HUMAN ONLY gate.

[...]

Issue #878, opened 2026-08-19
task: add AUTO_SHUTDOWN_MILLIVOLTS=3400 to the RCC6 BLE companion envs


Strycher:
The RCC6 defines no `AUTO_SHUTDOWN_MILLIVOLTS`, and there is no shared default — `[verified: grep across variants/ and platformio.ini; only heltec_rc32 sets it, on three envs]`.

## Why it matters now

#833 measured RC32 runtime as **21 h 35 m to auto-shutdown** — a time-to-a-specific-voltage, terminating at `AUTO_SHUTDOWN_MILLIVOLTS=3400`. An RCC6 discharge run is about to start on the **same 2500 mAh cell** for a direct board-to-board comparison.

Without the same cutoff the RCC6 run terminates at a different endpoint — or does not auto-shutdown at all — and the two numbers are not comparable no matter how well the role matches. The comparison is the entire point of the run.

## Change

Add `-D AUTO_SHUTDOWN_MILLIVOLTS=3400` to the RCC6 **BLE companion** envs, mirroring where RC32 scopes it:

| RC32 env | RCC6 equivalent |
|---|---|
| `heltec_rc32_companion_radio_ble` | `heltec_rcc6_companion_radio_ble` |
| `heltec_rc32_companion_radio_ble_diag` | `heltec_rcc6_companion_radio_ble_diag` |
| `heltec_rc32_without_display_companion_radio_ble` | *(no RCC6 equivalent)* |

RC32 deliberately does **not** set it on repeater, room_server or USB-companion envs — those are mains/solar/USB-powered roles. Same reasoning applies here, so this is scoped to the two BLE envs and no others.

## Run parameters this enables

Matching #833 as closely as the hardware allows: `heltec_rcc6_companion_radio_ble_diag`, BLE companion, display active, same 2500 mAh cell, same 3400 mV cutoff. GPS is inherently matched (RCC6 has none).

**Known uncontrolled difference:** the RCC6 bit-bangs its panel over software SPI (no free SPI peripheral — #806), where RC32 uses hardware SPI. That CPU cost is arguably part of what the comparison measures rather than a confound, but it should be stated when the result is reported.

Parent: #624.

PR #1006, opened 2026-08-26
task(#1005): AUTO_SHUTDOWN_MILLIVOLTS=3400 on the RC52 BLE companion envs

### LINKED ISSUE #1005: task: AUTO_SHUTDOWN_MILLIVOLTS=3400 on the RC52 BLE companion envs
`AUTO_SHUTDOWN_MILLIVOLTS` does not appear anywhere in `variants/heltec_rc52/platformio.ini` — zero occurrences on `firmware-base`. RC32 and RCC6 both carry it at 3400.

## Why this blocks the burn (#1004)

Quoting #903, which hit the identical blocker on RCC6:

> without the cutoff the run had **no endpoint** (the shutdown path compiles out when the macro is undefined)

Two consequences:

1. **No comparable termination.** #833 terminated at *exactly* 3400 mV, cleanly, still routing traffic. Without the macro the RC52 run ends in an undefined brownout, which is not the same measurement.
2. **Cell damage.** The run pulls the LiPo below safe voltage instead of the board stopping itself. `PWRMGT_VOLTAGE_BOOTLOCK=3300` does **not** cover this — that is the *boot* gate (`checkBootVoltage()`), a different mechanism that only acts at startup.

## Scope — BLE companion envs only, mirroring the fleet convention

Established placement, verified on `firmware-base`:

| Board | Envs carrying the cutoff |
|---|---|
| RC32 | `companion_radio_ble`, `without_display_companion_radio_ble`, `companion_radio_ble_diag` |
| RCC6 | `companion_radio_ble` (#878) |

So RC52 gets it on `[env:heltec_rc52_companion_radio_ble]` and `[env:heltec_rc52_without_display_companion_radio_ble]`. The `_diag` and `_diag_nobuf` envs inherit it through `extends`, so it is **not** duplicated there.

Deliberately **not** applied to:
- **USB companion envs** — tethered by definition; there is no cell to protect.
- **Repeater / room server** — typically mains or solar. Self-shutdown on a voltage dip is worse behavior than continuing, and changing that is a fleet-wide policy decision, not a side effect of a battery test.

## Value

**3400 mV**, matching RC32 and RCC6 so the three RadioCore runtimes are directly comparable. The in-tree nRF52 boards (`lilygo_techo`) use 3300; comparability was judged the stronger argument, and the divergence is recorded here deliberately.

## Definition of done

Builds clean on the affected envs, macro confirmed present in the generated build flags, change is on `firmware-base` before the #1004 burn image is built.

Strycher:
Adds the runtime battery cutoff the RC52 discharge run ([#1004](https://github.com/OffbandMesh/meshcore-firmware/issues/1004)) needs to have an endpoint at all. Closes [#1005](https://github.com/OffbandMesh/meshcore-firmware/issues/1005).

## What was wrong

`AUTO_SHUTDOWN_MILLIVOLTS` appeared **nowhere** in `variants/heltec_rc52/platformio.ini` — zero occurrences on `firmware-base`. The shutdown path compiles out when the macro is undefined, so a discharge run had no defined termination and would pull the cell below safe voltage instead of the board stopping itself. Same blocker [#878](https://github.com/OffbandMesh/meshcore-firmware/issues/878) cleared for RCC6.

`PWRMGT_VOLTAGE_BOOTLOCK` (3300) does **not** cover this — that is the boot gate in `checkBootVoltage()`, a different mechanism that only acts at startup.

## The change

`-D AUTO_SHUTDOWN_MILLIVOLTS=3400` on the two **BLE companion** envs — with-display and without-display. `_diag` and `_diag_nobuf` inherit through `extends`, so it is not duplicated.

**3400 for comparability.** It is what makes the three RadioCore runtimes mean anything against each other:

| Board | MCU | Runtime to 3400 mV | Avg current |
|---|---|---|---|
| RC32 | ESP32-S3 | 21 h 34 m 31 s | ~116 mA |
| RCC6 | ESP32-C6 | 43 h 21 m 55 s | ~58 mA |
| RC52 | nRF52840 | *this PR enables it* | ~10–16 mA observed |

The in-tree nRF52 boards (`lilygo_techo`) use 3300; the divergence is deliberate and documented inline.

**Not applied** to USB companion envs (tethered, no cell to protect) or repeater/room_server (mains or solar — self-shutdown on a dip is worse behavior than continuing). This matches RC32 and RCC6, which exclude the same roles.

## Verification

Confirmed from the **resolved compiler defines** (`pio run -t idedata`), not read off the ini:

```
AUTO_SHUTDOWN_MILLIVOLTS=3400   <- this PR
OFFBAND_POWER_TELEMETRY         <- via #1003
OFFBAND_FORCE_CAPLOG
OFFBAND_LOG_MIRROR_UART=1
```

The macro resolves **once** in `heltec_rc52_companion_radio_ble_diag`, proving the `extends` inheritance rather than assuming it. Builds SUCCESS on all three affected envs (`..._companion_radio_ble`, `..._companion_radio_ble_diag`, `..._without_display_companion_radio_ble`).

## Gemini adversarial gate — 5 findings, all dispositioned

| # | Finding | Disposition |
|---|---|---|
| 1 | `extends` works; duplicate `-D` harmless | No defect; matches the empirical check |
| 2 | `;` comments inside `build_flags` "will break the build" | **Refuted.** Three envs build SUCCESS with them, RC32 already uses this pattern throughout, and the resolved defines are clean |
| 3 | 3400 above the 3300 boot gate causes a brownout boot loop | **Refuted by the code.** Below 3300 `checkBootVoltage()` calls `initiateShutdown()` → SYSTEMOFF, so it cannot loop; 3.3–3.4 V is far above the nRF52840's ~1.7 V brownout; and the stated mechanism had cell relaxation moving the wrong way. RC52 is the only board carrying both gates, so the analysis is recorded inline rather than left implicit |
| 4 | Excluding repeater/room_server is "indefensible" — a battery repeater runs its cell flat | **Referred to the owner, not silently dismissed.** The point is fair, but RC32 and RCC6 exclude these roles too, so this PR matches convention rather than opening a new gap. Widening it is a fleet-wide policy change |
| 5 | Single unaveraged ADC sample against a nominal 4.90 divider | **Accepted as caveats.** Tracked on [#982](https://github.com/OffbandMesh/meshcore-firmware/issues/982) and [#857](https://github.com/OffbandMesh/meshcore-firmware/issues/857). ±3.5% is ±119 mV at 3400, so **"3400" is nominal per board, not actual** — the MAX17048 reading at termination is the honest cross-board figure and will be reported alongside the runtime |

## Scope

One file, one macro, two envs. No source changes, no shared-driver changes, no behavior change on any board other than RC52 BLE companions.

Epic #1004 stays open and carries the burn.

### DIFF
--- a/variants/heltec_rc52/platformio.ini
+++ b/variants/heltec_rc52/platformio.ini
@@ -290,13 +290,21 @@ build_flags =
   ; once in checkBootVoltage() at startup; this acts during the run.
   ;
   ; RC52 is the only board in the fleet carrying BOTH (RC32/RCC6 have no boot
-  ; gate), so the interaction was reviewed rather than assumed. A cutoff ABOVE
-  ; the boot gate cannot boot-loop the board: below 3300 checkBootVoltage()
-  ; calls initiateShutdown() and the part enters SYSTEMOFF, and between 3300
-  ; and 3400 it boots and then shuts down cleanly again on the runtime path.
-  ; Neither is a brownout -- the nRF52840 browns out near 1.7 V, far below
-  ; both numbers. The visible effect is a short boot/shutdown cycle on a
-  ; near-flat cell, not damage. [reviewed 2026-08-26, #1005]
+  ; gate). The cutoff cannot boot-loop the board -- below 3300
+  ; checkBootVoltage() calls initiateShutdown() and the part enters SYSTEMOFF,
+  ; and the nRF52840 browns out near 1.7 V, far below either number.
+  ;
+  ; But do NOT read those two as the whole picture, because a THIRD gate runs
+  ; before both of them and is the one that actually fires:
+  ; SafeBoot::checkAndMaybeSleep() (simple_repeater/main.cpp, before
+  ; board.begin()) is compiled in on this board -- PIN_VBAT_READ is defined
+  ; above, which satisfies the SafeBoot.cpp gate -- and it sleeps below
+  ; SLEEP_MV 3400, higher than the 3300 bootlock and equal to this cutoff.
+  ; The board therefore never reaches begin() in the 3300-3400 band, which
+  ; makes PWRMGT_VOLTAGE_BOOTLOCK effectively unreachable: armed protection
+  ; that reads as live and is not. Three thresholds, three mechanisms, no
+  ; single owner. Being fixed as one coherent policy in #1007 (under #882) --
+  ; do not tune any of these three in isolation. [verified 2026-08-26, #1005]
   -D AUTO_SHUTDOWN_MILLIVOLTS=3400
 build_src_filter = ${Heltec_RC52_with_display.build_src_filter}
   +<helpers/nrf52/SerialBLEInterface.cpp>
--- a/variants/heltec_rc52/platformio.ini
+++ b/variants/heltec_rc52/platformio.ini
@@ -273,6 +273,31 @@ build_flags =
   -D MAX_GROUP_CHANNELS=40
   -D BLE_PIN_CODE=123456
   -D OFFLINE_QUEUE_SIZE=256
+  ; Battery cutoff. The firmware shuts the board down cleanly at this reading
+  ; rather than running the cell flat -- without it the shutdown path compiles
+  ; out entirely and a discharge run has no defined endpoint (#1005).
+  ;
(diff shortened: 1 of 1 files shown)


Strycher:

Issue #1012, opened 2026-08-27
bug: pio-flash resolves the device registry, flash history and backups per-worktree — audit trail is already split and a registry copy is 7 days stale


Strycher:
`pio-flash` addresses three durable, per-host artifacts **relative to the checkout it happens to be running from**. Every worktree that runs a flash forks all three. There are currently **80 worktrees**.

## Measured state, not hypothetical

`[verified: 2026-08-27]`

**The flash audit trail is already split:**

```
2805 entries   C:\Dev\meshcore-firmware\flash-history.jsonl                  <- canonical
   6 entries   C:\Dev\.worktrees\meshcore-firmware-1004\flash-history.jsonl  <- 2026-08-26/27
```

Those 6 records are real flash operations — `rc52-bench-1`, `heltec-t096-1`, `hv4-bench-1` — and they are **not in the canonical safety record.** The history is the audit trail of what firmware went onto which device and when; fragmenting it defeats its purpose.

**The device registry has three copies, one a week stale:**

```
55302 B  2026-08-27  meshcore-firmware/hardware-devices.yaml
55302 B  2026-08-27  .worktrees/meshcore-firmware-1004/hardware-devices.yaml
39483 B  2026-08-20  .worktrees/meshcore-firmware-prefix/hardware-devices.yaml   <- 7 days stale
```

A flash run from `meshcore-firmware-prefix` resolves identity against a registry that predates a week of device changes. Given #503 made identity resolution the core safety property, a stale registry is a safety regression, not an inconvenience.

`flash-backups/` has the same addressing and currently exists only in the primary clone — by luck, because backups have only been run there.

## Root cause — one assumption

```python
PROJECT_ROOT = Path(__file__).resolve().parent.parent
REGISTRY_PATH = PROJECT_ROOT / "hardware-devices.yaml"
FLASH_HISTORY_PATH = PROJECT_ROOT / "flash-history.jsonl"
```

The script derives its root from **its own location**, and the script is checked into the repo. Durable per-host state is therefore addressed through an ephemeral checkout. This is an assumption from when there was one clone, not a design decision.

Because these files are gitignored, nothing detects the divergence: each worktree silently gets its own, and staleness is invisible until someone compares them by hand.

## The correct pattern is already in the same file

Twenty lines below the constants above:

```python
FIRMWARE_DIR = Path(os.environ.get("PIO_FLASH_FIRMWARE_DIR", str(PROJECT_ROOT)))
[...]

PR #1013, opened 2026-08-27
fix(#1012): one registry, one flash history, outside every checkout

Strycher:
Fixed in `d30f69fc` on `epic/736-canonical-flash-state`.

## One registry, one history, this repo

```
C:\Dev\.offband\hardware-devices.yaml
C:\Dev\.offband\flash-history.jsonl
C:\Dev\.offband\flash-backups\
```

Outside every checkout. Not in the clone, not in any of the 80 worktrees, never read from or merged with `C:\Dev\LoRa` or `C:\Dev\wadamesh` — their device short-names are not guaranteed to mean the same hardware.

## Verified, not asserted

| Check | Result |
|---|---|
| All four consumers resolve one path | ✅ 1 unique path |
| Copies remaining in clone / worktrees | **0** |
| Path logic definitions in the tree | **1** (`scripts/offband_state.py`) |
| OTA guard IPs | **0 → 2** (it was silently fail-open after the move) |
| History consolidation | 2805 + 6 = **2811**, content-hash verified, 0 dupes, 0 lost, 0 foreign |
| Backups | 57 / 57 files |
| Registry | 27 devices, all fields intact |

Originals preserved at `C:\Dev\.offband\_migrated-2026-08-27\` — nothing deleted.

## Consumers repointed

`scripts/pio-flash.py`, `scripts/ota-push.py`, `.claude/hooks/block-raw-curl-ota.py`, and `scripts/companion_harness.py` — whose "find the clone that holds the registry" search was itself a workaround for this bug and is gone.

## Gemini gate — 6 findings

**Fixed (3):**
1. **`_default_state_dir` was copy-pasted into `pio-flash.py`** as well as living in `offband_state.py` — a second copy of the path logic, which is the same class of bug this change exists to remove. The most damning finding; it directly contradicted the change's purpose. `pio-flash` now imports.
2. **The OTA hook's `except ImportError` fallback reimplemented the path logic incorrectly** — it ignored `PIO_FLASH_REGISTRY` and the POSIX default, creating a new silent-failure path in a security hook. Removed; the hook now warns loudly and reports the guard as open rather than guessing.
3. **`sys.path.insert(0, ...)` in a security hook** allowed a file in `scripts/` to shadow a stdlib module. Changed to `append`.

**Refuted with evidence (2):**
4. *"Merging JSONL by non-unique `ts_unix` will cause data loss on collision."* — False. The merge is `sorted(a + b, key=ts)`; sorting a concatenation never drops records. Demonstrated with a deliberate same-second collision: 2 + 2 → 4 records, none lost. The real merge was independently verified by content hash — every source record present, zero extras.
5. *"Clearly a manual spot-check; 77+ worktrees never scanned."* — False. A full-tree sweep was run (`find C:/Dev -name flash-history.jsonl` and the same for the registry). It is what found the LoRa and wadamesh histories and the 7-day-stale `meshcore-firmware-prefix` registry.

**Referred to the owner (1):**
6. *"Hardcoded `C:\Dev` is indefensible; use `%LOCALAPPDATA%`."* — A fair convention argument. `C:\Dev\.offband` was chosen because it sits beside the clones and is discoverable; `%LOCALAPPDATA%` is more conventional but hidden. `OFFBAND_STATE_DIR` overrides either way. Owner's call, not the reviewer's.

## Residual risk, stated rather than hidden

The 80 existing worktrees hold **old copies of these scripts** that still compute checkout-local paths. They cannot be patched retroactively. Because the old files have been moved away, those old scripts now **refuse** with "registry not found" rather than reading stale data — which is the safe failure. The remaining exposure is someone running `pio-flash bootstrap` from an old worktree and creating a fresh empty registry there.

### DIFF
--- a/.claude/hooks/block-raw-curl-ota.py
+++ b/.claude/hooks/block-raw-curl-ota.py
@@ -50,7 +50,28 @@ import sys
 from pathlib import Path
 
 PROJECT_ROOT = Path(__file__).resolve().parent.parent.parent
-REGISTRY_PATH = PROJECT_ROOT / "hardware-devices.yaml"
+
+# #1012: the registry is per-host and canonical, NOT checkout-local.
+#
+# Resolution lives in scripts/offband_state.py and is NOT duplicated here: a
+# second copy of the path logic drifts, and a security hook reading a
+# different registry than the tool it guards is worse than no hook.
+#
+# sys.path.append (NOT insert(0)) so a file dropped in scripts/ cannot shadow
+# a stdlib module for this hook.
+sys.path.append(str(PROJECT_ROOT / "scripts"))
+try:
+    from offband_state import registry_path
+except ImportError as _e:  # no silent second implementation -- see below
+    print(
+        "WARN: block-raw-curl-ota: cannot import scripts/offband_state.py "
+        f"({_e}); OTA IP guard is OPEN. Fix the checkout.",
+        file=sys.stderr,
+    )
+    registry_path = None
+
+REGISTRY_PATH = registry_path() if registry_path else None
+
 WRAPPER_PATH = "scripts/ota-push.py"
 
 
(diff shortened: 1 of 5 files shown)


Strycher:

Issue #984, opened 2026-08-25
bug: RAK3401 WisMesh 1W repeater (wismesh1w-repeater) is not transmitting -- ~76 dB below a working board, unreachable on the mesh


Strycher:
## Summary

`wismesh1w-repeater` (RAK10724 WisMesh 1W Booster — RAK3401 nRF52840 + RAK13302 SX1262 + SKY66122 1W PA) accepts CLI commands and reports `OK - zerohop advert sent`, but produces essentially no RF. Confirmed by two independent observations at different layers.

## Evidence

`[verified: owner-operated bench, 2026-08-24, conducted measurement via tinySA]`

**RF layer — conducted, zero span at 910.525 MHz:**

| board | tx | measured | capture |
|---|---|---|---|
| **rc32-bench-1** (control) | 22 | **+20.30 dBm** | 10/10 |
| **wismesh1w-repeater** | 22 | **~ −55.60 dBm** | 7/10, marginal |

Same chain, same session, same settings. **Delta ~76 dB.**

**Mesh layer:** its adverts are **not picked up by any local node** (owner-observed). Independent of the instrument, and agrees.

## Method — validated before use

- **Zero span** at 910.525 MHz, RBW 100 kHz (must exceed the 62.5 kHz occupied bandwidth), `sweeptime 0.4`, `ext_gain -40.2`, trigger armed from the **measured** noise floor on every run.
- **Validated against a known value:** rc32-bench-1 at tx 16 read **+13.30** on this method vs **+13.10** from an earlier independent method — 0.2 dB apart, inside one quantization step, 10/10 capture.
- Chain: measured pad+cable **39.73 dB** (NanoVNA, repeated across two days, 0.25 dB agreement; isolation floor −64.62 dB → ~25 dB margin) plus an **estimated** 0.4–0.8 dB IPEX pigtail that cannot be isolated in this setup. Error bar ±0.6 dB.

## Other symptoms on this unit

- **Dropped off USB entirely mid-run, during transmission** (advert 9 of 10), and did not re-enumerate until physically replugged. Did not recur on later runs.
- Unchanged by `set radio 910.525,62.5,7,5` + `reboot`.
- Otherwise responsive: `get name` → `SCN-BASE-1W`, `ver` → v1.16.0 (Build: 6 Jun 2026), CLI answers normally.

## NOT established

**Cause is unknown.**

[...]

Issue #104, opened 2026-06-12
rak3401: add -D RAK_BOARD so WisBlock RAK12500 (I2C u-blox) GPS compiles in


Strycher:
## Summary

The `rak3401` variant is missing `-D RAK_BOARD`, so the WisBlock I2C GPS path — which drives the **RAK12500 / u-blox ZOE-M8Q** over I2C — is compiled out. A RAK12500 on any RAK3401-based board (e.g. the WisMesh 1W Booster, RAK10724) is therefore never detected, and the MeshCore app shows GPS as **"not available."**

## Root cause

In `src/helpers/sensors/EnvironmentSensorManager.cpp`:

```c
#if ENV_INCLUDE_GPS && defined(RAK_BOARD) && !defined(RAK_WISMESH_TAG)
#define RAK_WISBLOCK_GPS
#endif
```

`RAK_WISBLOCK_GPS` gates the entire I2C u-blox driver (`RAK12500LocationProvider`, `ublox_GNSS.begin(Wire)`, `rakGPSInit()` / `gpsIsAwake()`), and it requires `RAK_BOARD`.

`RAK_BOARD` is defined by `rak4631`, `rak_wismesh_tag`, and the `gat562_*` variants — **but not by `rak3401`** (which only defines `-D RAK_3401`). So `rak3401` builds compile the `#else` branch, `initBasicGPS()`, which is **UART/`Serial1`-only** and never scans I2C → an I2C-only module (RAK12500) is invisible. This is a variant oversight: `rak3401` was derived from `rak4631` but dropped `-D RAK_BOARD`.

## Fix

Add `-D RAK_BOARD` to the `[rak3401]` base `build_flags` in `variants/rak3401/platformio.ini` (mirroring `rak4631`). Applies to all `rak3401` envs (companion BLE/USB, repeater, room server, sensor). One-line build-flag change; no logic change.

## Hardware / verification

- First device: RAK10724 WisMesh 1W Booster (RAK3401 + RAK13302), companion BLE — see #103.
- GPS module: **RAK12500 (u-blox ZOE-M8Q)**, mounted in **Slot C** (Slot F is physically buried; A/B are poor layout; Slot D's socket isn't in the probe list). `gpsIsAwake()` walks `WB_IO2`(Socket A) / `WB_IO4`(Socket C) / `WB_IO5`(Socket F) — **Slot C = `WB_IO4` is covered.**
[...]

PR #1015, opened 2026-08-27
fix(#984): set RAK3401_NO_GPS on every RAK3401 env — the GPS probe kills TX

Strycher:
## Root cause: the GPS probe drives and then floats the radio's power rail

`EnvironmentSensorManager::gpsIsAwake()` drives WisBlock IO pins hunting for a GPS module. On RAK3401, `variant.h` aliases **`WB_IO2` to `PIN_3V3_EN`** — the switched 3V3_S rail and 5V boost feeding the SX1262 and its SKY66122 PA, driven HIGH on purpose by `RAK3401Board::begin()`, which runs **before** the probe.

The probe pulls that rail LOW for 500 ms and, when no GPS answers, abandons the pin as a floating `INPUT`.

**Only the two companion envs carried `RAK3401_NO_GPS`. The other five did not.** That is the single build difference that touches the radio — every RF define is identical across roles:

| | repeater | companion |
|---|---|---|
| `LORA_TX_POWER` | 22 | 22 |
| `SX126X_CURRENT_LIMIT` | 140 | 140 |
| `SX126X_DIO2_AS_RF_SWITCH` | via variant.h | via variant.h |
| `SX126X_POWER_EN` | via variant.h | via variant.h |
| **`RAK3401_NO_GPS`** | **absent** | **present** |

That is why the companion works and the repeater does not, on a factory-sealed unit.

## What is NOT claimed

The existing comment on this flag says the TCXO cannot start and RadioLib returns `-707`. **That was not observed here and is not asserted.** `simple_repeater/main.cpp:1207` calls `halt()` when `radio_init()` fails, so a genuine init failure would leave the board mute. The un-flagged build answered every CLI command sent to it — `set name`, `set radio`, `set tx`, reboot, six read-backs. `radio_init()` succeeded.

What the measurements support is that the radio comes up while its front end does not: firmware acknowledges every transmit and almost nothing leaves the board.

`[hypothesis: untested]` the floating rail leaves the PA unpowered or marginal while the SX1262 core still runs. **The exact electrical mechanism is not diagnosed.**

## Evidence

Two units, one factory-sealed, forced transmits via `advert.zerohop`, zero-span analyser, `send_failed=0` on every level:

- No relationship between commanded TX power and output — tx 8 read hotter than tx 18
- 2 of 11 cycles captured anything at all; five levels captured nothing
- Range −21.6 to −55.1 dBm against +20.30 from healthy boards
- Healthy boards the same night: `stale=0`, spread ≤1.0 dB, monotonic

A hypothesis that the build-visibility of `SX126X_DIO2_AS_RF_SWITCH` was the cause was **disproven** by preprocessor probe — the macro is visible and `setDio2AsRfSwitch(true)` is called on both roles.

## The fix

`-D RAK3401_NO_GPS` on all seven RAK3401 envs. `kiss_modem` had no `build_flags` block, so one was added inheriting the base solely to carry the flag.

Fixing the probe itself was investigated on this hardware, found to be a nest of conflicts, and **deferred indefinitely**. Disabling GPS here is the standing decision — it was simply never applied beyond the companions. This applies it; it does not change it.

## Verification

Preprocessor `#pragma message` in both branches of the `RAK_WISBLOCK_GPS` guard, built, then removed:

| env | probe |
|---|---|
| all seven RAK3401 envs | **COMPILED OUT** |
| `RAK_4631_repeater` (control) | COMPILED IN — unaffected |

Eight envs build SUCCESS. Flashed to `RAK3401-radio-3`, sha256 `f16e3fa37bba8a80ef42d838db973db1af27518b6b1d00e84455734d3e43a62c`; config read-back intact.

## ⚠ Still owed

**RF confirmation. The board has not been re-swept.** The fix is not yet proven to restore transmit — that requires a sweep on the fixed measurement tool, with `tx=10` on this unit (not 22) and adverts forced, since it is configured silent.

## Follow-ups not actioned here

- `EnvironmentSensorManager.cpp:205` still reads *"Repeater (has GPS) keeps the path"* — now false, and invites re-enabling the probe.
- A RAK3401 that genuinely has a GPS module fitted now cannot use it. `sensor` is the most affected role.

### DIFF
--- a/variants/rak3401/platformio.ini
+++ b/variants/rak3401/platformio.ini
@@ -6,7 +6,20 @@ build_flags = ${nrf52_base.build_flags}
   ${sensor_base.build_flags}
   -I variants/rak3401
   -D RAK_3401
-  -D RAK_BOARD               ; #104: compile the WisBlock RAK12500 (u-blox ZOE-M8Q) I2C GPS path. Probe is slot-agnostic; Slot D is the RAK19007 GNSS slot.
+  ; #984: -D RAK_BOARD REMOVED. #104 added it to compile the WisBlock RAK12500
+  ; I2C GPS path. It is the ONLY thing that gates gpsIsAwake(), whose probe drives
+  ; WB_IO2 == PIN_3V3_EN -- the 3V3_S rail and the 5V boost feeding the SX1262 and
+  ; its SKY66122 PA -- then abandons that pin floating when no GPS answers.
+  ;
+  ; Upstream MeshCore does NOT define RAK_BOARD for this variant, so upstream never
+  ; compiles the probe at all. Every other RAK3401 file here is functionally
+  ; identical to upstream (variant.cpp, target.cpp/h, RAK3401Board.h byte-identical;
+  ; variant.h and RAK3401Board.cpp differ by comments only). This define was the
+  ; whole divergence.
+  ;
+  ; RAK3401_NO_GPS is kept per-env as defense in depth -- it independently blocks the
+  ; probe if RAK_BOARD is ever reinstated. Re-enabling GPS needs the pin conflict
+  ; solved first; that work was investigated and deferred. See #984.
   ; #104: GPS pins mirror rak4631 -- the RAK12500 path references these (Serial1 setup + PIN_GPS_EN) even though the module is I2C.
   -D PIN_GPS_TX=PIN_SERIAL1_RX
   -D PIN_GPS_RX=PIN_SERIAL1_TX
--- a/variants/rak3401/platformio.ini
+++ b/variants/rak3401/platformio.ini
@@ -49,6 +49,7 @@ build_flags =
   -D ADVERT_LON=0.0
   -D ADMIN_PASSWORD='"password"'
   -D MAX_NEIGHBOURS=50
+  -D RAK3401_NO_GPS              ; #984: boot probe floats WB_IO2 (radio power) -> kills TX (-707)
   ; PIN_STATUS_LED now inherited from the [rak3401] base (was repeater-only) -- #7/#9
   ;-D MESH_PACKET_LOGGING=1
   ;-D MESH_DEBUG=1
@@ -68,6 +69,7 @@ build_flags =
   -D ROOM_PASSWORD='"hello"'
   ;-D MESH_PACKET_LOGGING=1
   ;-D MESH_DEBUG=1
+  -D RAK3401_NO_GPS              ; #984: boot probe floats WB_IO2 (radio power) -> kills TX (-707)
 build_src_filter = ${rak3401.build_src_filter}
   +<helpers/ui/SSD1306Display.cpp>
   +<../examples/simple_room_server>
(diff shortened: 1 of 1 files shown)


Strycher:

Strycher:
## Second commit: `-D RAK_BOARD` removed — this is the actual divergence

`e76a11e0`. The first commit set `RAK3401_NO_GPS` on every env. This one removes the define that made the probe compile in at all.

### Diffed against upstream rather than reasoned about

`RAK_BOARD` was added by #104 to compile the WisBlock RAK12500 I2C GPS path. **Upstream MeshCore does not define it for this variant**, so upstream never compiles `gpsIsAwake()` — the probe that drives `WB_IO2 == PIN_3V3_EN` (the 3V3_S rail and the 5 V boost feeding the SX1262 and its SKY66122 PA) and then abandons the pin floating.

Everything else in `variants/rak3401` is functionally identical to upstream:

| file | vs `upstream/main` |
|---|---|
| `variant.cpp` | identical |
| `target.cpp` | identical |
| `target.h` | identical |
| `RAK3401Board.h` | identical |
| `variant.h` | +3 −1, **comment only** |
| `RAK3401Board.cpp` | +5 −1, **comment only** |
| `platformio.ini` | our envs, SafeBoot, CrashLog, LED — **and `RAK_BOARD`** |

### Removing it costs no non-GPS capability

`RAK_BOARD` gates exactly three things, all GPS: the `rakGPSInit()`/`gpsIsAwake()` declarations in `EnvironmentSensorManager.h:45`, the `RAK_WISBLOCK_GPS` guard at `:172`, and an `ETHERNET_ENABLED` guard that is inert on this board.

`RAK3401_NO_GPS` is kept per-env as defense in depth — it independently blocks the probe if `RAK_BOARD` is ever reinstated.

### Verification

Preprocessor `#pragma message` in both branches of the `RAK_WISBLOCK_GPS` guard, built, then removed:

| env | GPS path |
|---|---|
| `RAK_3401_repeater` | **OUT** |
| `RAK_3401_companion_radio_ble` | **OUT** |
| `RAK_3401_sensor` | **OUT** |
| `RAK_4631_repeater` (control) | IN — unaffected |

Four envs build SUCCESS. Flash **56.5% → 52.7%** as the GPS stack left the image. Flashed to `RAK3401-radio-3`, sha256 `6245fb4ec840121a5800af34fa93581ddf2f2d43e2f849b9477ab67298cbc014`; config read-back intact.

### ⚠ Scope of what this closes

**This closes the REGRESSION. It does not close the RF fault.**

The board still measures roughly **36 dB below a FEM-less RAK4631**, with a *correct* transfer curve — 1.11 dB/step, R² 0.96, monotonic — where the pre-fix runs were noise (0.25 dB/step, R² 0.02). The radio modulates correctly; something between the SX1262 and the SMA is attenuating.

**That code is upstream and unmodified**, so it is not a defect we introduced. The FEM control rests on an uncited claim in `variant.h` (`CSD + CPS` tied together, HIGH = enabled; `CTX` on DIO2) with no datasheet or schematic referenced anywhere in the variant, and RAK3401 is the only SKY66122 board in the tree — no working reference exists to compare against.

Issue #1016, opened 2026-08-28
task: add RC52 to the release matrix — companion (BLE/USB) and repeater


Strycher:
RC52 has **zero entries** in `.github/release-envs.txt`. The file lists 102 envs; RC32 contributes 4 (+3 testdiag), RCC6 contributes 4 (+4 diag), RC52 contributes none.

**Consequence: cutting a release today produces no RC52 firmware at all.** The code is fully merged — 19 RC52 envs exist on `firmware-base`, including the display builds, with `AUTO_SHUTDOWN_MILLIVOLTS=3400` (#1005), `OFFBAND_POWER_TELEMETRY` (#889), `RC52_DISPLAY_BUFFER_INDEXED` and `RC52Display.cpp` with the colour splash (#948). The firmware is ready; the release wiring was never done.

## Scope — owner-specified

| env | role |
|---|---|
| `heltec_rc52_companion_radio_ble` | Companion, BLE |
| `heltec_rc52_companion_radio_usb` | Companion, USB |
| `heltec_rc52_repeater` | Repeater |

These are the **display** builds. Headless (`without_display_*`) is for production low-power or a headless bench request, not the default product.

**`heltec_rc52_room_server` deliberately EXCLUDED** — owner decision, on the strength of #965: the RC52 room server watchdog-resets once at ~29 s and renders nothing on that first boot, then recovers and stays up. Open, P1, reproducible, and it predates the display work. Revisit when #965 closes.

There is no RC52 sensor env; nothing to exclude there.

## Diag envs — NOT included, and that is deliberate

`release-envs.txt` states the rule and its limits in its own words:

> The `_diag` / `_diag_nocrashlog` envs are diagnostic instruments, not products. This is the RULE... TWO owner-approved EXCEPTIONS now exist, both scoped to beta hardware and neither inheritable... **Extending this to a third board is a fresh owner decision, not an inheritance from these two.**

RC52 would be the third board. No diag env is added here. If beta testers on RC52 need to produce evidence the way RC32 and RCC6 testers can, that is a separate owner decision and should be recorded as one.

## ⚠ Known gap this exposes — the shipped envs are not CI-built

`ci.yml` references RC52 four times, and **all four are headless diag envs**:

```
heltec_rc52_without_display_companion_radio_ble_diag
heltec_rc52_without_display_companion_radio_usb_diag
heltec_rc52_without_display_repeater_diag
heltec_rc52_without_display_room_server_diag
```

[...]

Issue #880, opened 2026-08-20
epic: RC52 display support — nRF52 NV3001B port and back buffer


Strycher:
Give the RC52 a working T108 / NV3001B panel on its real hardware SPI.

Created 2026-08-19 when the owner restructured **#625 from an Epic to a Feature**. Split from the headless bring-up epic because the display is a genuinely separate deliverable with a genuinely separate blocker — the driver does not compile for nRF52 at all today.

## Why this is its own epic, not a task

`NV3001BDisplay` does not build for nRF52 in **either** SPI mode.
`[verified: src/helpers/ui/NV3001BDisplay.{h,cpp} + framework-arduinoadafruitnrf52 SPI.h, 2026-08-19]`

| # | Site | Problem |
|---|---|---|
| 1 | `NV3001BDisplay.h:55` — `SPIClass spi;` | No default ctor on nRF52. **Unconditional member**, so `NV3001B_USE_SOFTWARE_SPI=1` does not sidestep it. |
| 2 | `.cpp` — `spi.begin(SCL, MISO, SDA, CS)` | nRF52 `begin()` takes no arguments. |
| 3 | `.cpp` — `spi.writeBytes()` | ESP32-only. |
| 4 | `.cpp` — `spi.writePattern()` | ESP32-only; needs an expand-to-buffer loop. |

That is a bus-layer port touching a shared driver used by RC32 and RCC6, plus a 56 KB allocation question on a 256 KB part. Multi-PR by any reasonable sizing.

## Children

| Issue | Scope |
|---|---|
| #872 | Port the `NV3001BDisplay` bus layer to nRF52 (SPI1 / SPIM3) |
| #856 | 56 KB back buffer — allocation on nRF52, and the fallback path that has never executed on hardware |

## Hardware facts to build on

`[verified: vendor variant.h + RC52-L62_V1.02 schematic, 2026-08-19]`

- TFT is on **SPI1 backed by SPIM3** (`SPI_32MHZ_INTERFACE 1`), LoRa on SPI0 — so unlike the RCC6 there IS a free hardware peripheral. **Do not port `NV3001B_USE_SOFTWARE_SPI=1`.**
[...]

PR #1017, opened 2026-08-28
feat(#1016): ship RC52 — add it to the release matrix and gate it in CI

### LINKED ISSUE #965: bug: RC52 room server watchdog-resets at ~29 s and renders nothing on the first boot

Strycher:
RC52 had **zero entries** in `.github/release-envs.txt`. The firmware has been merged for some time, but a release would have produced **no RC52 firmware at all**. The code was ready; the release wiring was never done.

## What ships

| env | role |
|---|---|
| `heltec_rc52_companion_radio_ble` | Companion, BLE |
| `heltec_rc52_companion_radio_usb` | Companion, USB |
| `heltec_rc52_repeater` | Repeater |
| `..._ble_diag` / `..._usb_diag` / `..._repeater_diag` | matching diag pairs |

Release matrix **102 → 108**. These are the **display** builds; `without_display_*` is for a production low-power build or a headless bench request, not the product.

## Exclusions, each documented in the file

- **`heltec_rc52_room_server`** and its diag — owner decision on [#965](https://github.com/OffbandMesh/meshcore-firmware/issues/965): watchdog-resets once at ~29 s and renders nothing on that first boot, then recovers. Open, P1, reproducible, predates the display work. The file carries a note to add it when #965 closes.
- **`..._ble_diag_nobuf`** — its own variant comment: *"BENCH EXPERIMENT ONLY... NOT in the release matrix. Do not 'complete the set' by adding twins of this."*
- **`without_display_*`** — not the product.

## The diag exception was asked for, not inherited

`release-envs.txt` states the RC32 (#824) and RCC6 (#806) diag-as-product exceptions are *"NOT inheritable... Extending this to a third board is a **fresh owner decision**, not an inheritance from these two."*

RC52 is that third board. The decision was requested and granted, and is recorded as such — **including that it remains non-inheritable to a fourth board.**

## CI — closing a real gap

Before this, **every** RC52 env in `ci.yml` was a `without_display` `_diag` env. Nothing gated the builds that ship: a change breaking them merged green.

All six shipped envs are added. The four `without_display` diag envs are **kept** — they link no display driver and are the only CI coverage of the headless path, so dropping them loses coverage rather than duplicating it. RC52 goes 4 → 10; build matrix → 54.

Two stale comments corrected: the block claimed RC52 was *"deliberately NOT into release-envs.txt until #880 (display) lands"* and that *"this board has no display support yet"*. Both false since #880/#948.

## Verification

- All six shipped envs build **SUCCESS**
- Every name in the manifest confirmed to resolve to a real `[env:...]` in the variant ini
- `ci.yml` parses as valid YAML — 4 jobs, `matrix.env` 54 entries
- Scripted check: **zero** shipped RC52 envs missing from CI
- Re-verified after rebase

## Gemini gate — six findings, NONE acted on

Reported for owner review rather than actioned.

| # | Finding | Disposition |
|---|---|---|
| 1 | A typo in `release-envs.txt` fails one matrix job while `release.yml`'s `if: always()` publishes anyway → **silently incomplete release** | **Confirmed by reading `release.yml`** (`fail-fast: false` + `if: ${{ always() && ... }}`). Not introduced here — all six names verified — but the mechanism is real. Worth a lint check, separate change |
| 2 | +6 jobs slows a required merge gate for everyone | Fair; the CI addition was owner-approved and the cost is real and cumulative |
| 3 | "Shipping diag builds is indefensible" — UART state leak, power/perf cost, user cannot disable logging | **Owner has ruled**; RC32 and RCC6 already ship diag on the same rationale. Recorded that the reviewer disagrees with standing policy |
| 4 | Excluding `room_server` from release while still CI-building its headless diag is a confusing signal | Partly fair. #965 is a *runtime* bug; CI still catches compile breakage |
| 5 | Prose comments as policy are fragile — someone will cite RC52 as precedent for a fourth board, or fix #965 and forget to re-add the envs | **Strongest finding.** A `config-lint` rule would be the durable answer |
| 6 | YAML valid, matrix will not break | Agreed |

Findings 1 and 5 share one cheap fix — a lint check that every manifest name resolves to a real env. Not written here; owner's call.

### DIFF
--- a/.github/release-envs.txt
+++ b/.github/release-envs.txt
@@ -199,6 +199,49 @@ heltec_rcc6_repeater_display_diag
 #   -- every other observer env is ESP32/S3. Recorded as a fact worth knowing, not
 #   as a caveat: it is verified running on hardware (#830).
 
+# Heltec RadioCore RC52 (nRF52840 + HT-RA62A LoRa) -- #625 / #1016
+# The DISPLAY builds. Headless (without_display_*) is for a production low-power
+# build or a headless bench request, not the default product.
+#
+# heltec_rc52_room_server -- DELIBERATELY EXCLUDED, owner decision. #965: the
+#   room server watchdog-resets once at ~29 s and renders nothing on that first
+#   boot, then recovers and stays up. Open, P1, reproducible, and it predates the
+#   display work. Add it when #965 closes -- do NOT "complete the set" before then.
+# No RC52 sensor env exists; nothing to exclude there.
+#
+# THE THIRD DIAG EXCEPTION -- owner decision 2026-08-28 (#1016).
+#   The exceptions above (RC32 #824, RCC6 #806) are explicitly NOT inheritable:
+#   "Extending this to a third board is a fresh owner decision, not an
+#   inheritance from these two." RC52 is that third board and the owner has now
+#   made the decision. It was ASKED FOR and GRANTED, not assumed from precedent.
+#
+#   Same rationale as the other two: a beta tester on unproven hardware has to be
+#   able to produce evidence. Same scope limit: while RC52 is beta. Revisit when
+#   the board leaves beta. THIS IS STILL NOT INHERITABLE -- a fourth board needs
+#   its own decision.
+#
+#   One diag env per shipped role, and no others:
+#     * heltec_rc52_room_server_diag -- excluded, because room_server itself is
+#       excluded (#965). Pairs follow the shipped set.
+#     * heltec_rc52_companion_radio_ble_diag_nobuf -- excluded. Its own comment in
+#       variants/heltec_rc52/platformio.ini says: "BENCH EXPERIMENT ONLY -- NOT a
+#       shipping env, NOT in CI, NOT in the release matrix. Do not 'complete the
+#       set' by adding twins of this."
+#     * without_display _diag envs -- excluded; headless is not the product.
+#
+# WARNING -- these three are NOT built by ci.yml. Every RC52 env in the CI matrix
+#   is a without_display _diag env, so nothing gates the builds shipped here: a
+#   change that breaks them merges green. Adding them to ci.yml is a separate
+#   owner-approved change (#477 makes the matrix a required merge gate).
+heltec_rc52_companion_radio_ble
+heltec_rc52_companion_radio_usb
+heltec_rc52_repeater
+
+# RC52 diag pairs (see the decision note above)
+heltec_rc52_companion_radio_ble_diag
+heltec_rc52_companion_radio_usb_diag
+heltec_rc52_repeater_diag
+
 # Heltec T096 (nRF52)
 Heltec_t096_companion_radio_ble
 Heltec_t096_companion_radio_usb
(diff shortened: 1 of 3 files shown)


Strycher:

Issue #1042, opened 2026-09-04
task: verify a diag-build dispatch publishes a downloadable asset


Strycher:
Parent: #1034 (bug) · Sibling of #1041 (implementation) · Epic: #736

**The test that gates #1034's closure.** #1041's unit test proves the change compiles;
it cannot prove the workflow *works*. This task proves the bug is actually dead.

## Why a local build is not the test

Gemini review finding 1 on #1041, and it is correct:

> *"A local build only verifies compilation and the `build.sh` packaging step. It cannot
> exercise the final, most critical step: publishing the asset to the GitHub
> pre-release... The build job could pass, showing a green checkmark, but
> `gh release upload` could fail due to an expired token, insufficient permissions, a
> breaking change in the `gh` CLI on the runner, or the `diag-builds` release being in a
> state `--clobber` cannot handle."*

The failure this bug caused was *silent to everyone until someone dispatched the
workflow*. A test that does not dispatch it reproduces nothing.

## One mechanical constraint, stated as a fact rather than a sequence

`gh workflow run diag-build.yml --ref <branch>` requires the ref to exist **on the
remote**. That is the only hard ordering constraint here: this test cannot run against
a branch that has not been pushed.

Everything else about when this runs, and what its result gates, is the owner's to
sequence.

## Test procedure

Dispatch against the PR branch, for both observer envs:

```
gh workflow run diag-build.yml --ref epic/736-diag-build-flag -f env=heltec_v4_companion_observer_wifi
gh workflow run diag-build.yml --ref epic/736-diag-build-flag -f env=Heltec_v3_companion_observer_wifi
```

## Acceptance criteria — each is a claim with its own evidence

- [ ] Run reaches **`Build diagnostic firmware`** and it passes — proves the `#error`
      tripwire no longer fires in CI, not just locally
- [ ] Run reaches **`Verify a diagnostic .bin was produced`** and it passes — this guard
      exists because `build.sh` masks `cp` errors with `2>/dev/null || true`, so a green
      compile can still leave `out/` empty
- [ ] Run reaches **`Publish to the diag-builds pre-release`** and it passes — the step
      no local build can exercise
- [ ] An asset named `<env>-diag-<sha>.bin` is present on the `diag-builds` release, and
      its `<sha>` matches the branch head
- [ ] The published asset **downloads anonymously** (no auth), since self-serve download
      by a tester is the entire point of the workflow
- [ ] Both envs succeed, not just one
[...]

Issue #1041, opened 2026-09-04
task: correct the diag-build workflow's log-mirror flag to the post-#888 name


Strycher:
Parent: #1034 (bug) · Epic: #736

The implementation slice of #1034: correct the flag name `diag-build.yml` injects, so
the workflow compiles again.

## Scope — deliberately just the code change

- `.github/workflows/diag-build.yml` line 83: `-DOFFBAND_MESHLOG_UART0` ->
  `-DOFFBAND_LOG_MIRROR_UART`
- the stale header comment naming the old flag
- a short note recording *why* it broke: a build-flag rename has to sweep `.yml`,
  because workflows inject `PLATFORMIO_BUILD_FLAGS` from outside the build config the
  tree can see — which is the gap #888's sweep could not have caught by grepping
  `.ini`/`.h`/`.cpp`

## Done when

- [ ] Flag corrected and the stale comment updated
- [ ] Repo-wide sweep shows no remaining `OFFBAND_MESHLOG_UART0` outside the tripwire
      itself and historical CHANGELOG text
- [ ] Both observer envs build green locally with the corrected triad — proving the
      `#error` tripwire in `LogMirrorUart.h` no longer fires
- [ ] Gemini review run on the diff, findings acted on, before this task closes

## Explicitly NOT done here — why the parent stays open

This task ends at "the change is correct and compiles." **#1034 is not resolved until
the fix is merged and a real dispatch verifiably publishes an asset** — which is a
different claim, and one that cannot be made from a local build.

That is why this task exists separately: the code slice genuinely completes at push
time, while the bug does not. The parent closes on the owner's authorization after the
merge and a verified dispatch, per the verification steps recorded on #1034.

Agent: CloudyGrove (session fcc043c7)

Strycher:
## Gemini review — 4 findings, 2 code changes, 1 already structural, 1 no-op

`scripts/llm-consult.py --backend gemini --model gemini-2.5-pro`, audit log
`docs/llm-consultations/2026-09-04-1041-diag-build-flag-gemini-gemini-2.5-pro.log`
(`finish_reason: STOP`). Reviewed the workflow plus `LogMirrorUart.h` so the reviewer
could see the guard the flag has to satisfy.

### 1. Local build cannot prove the publish step — ACCEPTED, and it is why #1034 stays open

> *"A local build only verifies compilation and the `build.sh` packaging step. It cannot
> exercise the final, most critical step: publishing the asset to the GitHub
[...]

Issue #1034, opened 2026-09-02
bug: diag-build.yml injects the pre-#888 flag name — every diagnostic build fails


Strycher:
`.github/workflows/diag-build.yml` injects a build flag that #888 renamed, so every
dispatch fails at compile. **No diagnostic build can be produced for any board.**

## Evidence

`[verified: runs 33535936734 and 33535945894, 2026-09-01, both conclusion=failure]`

Failing step is *Build diagnostic firmware*:

```
src/helpers/LogMirrorUart.h:57:4: error: #error "OFFBAND_MESHLOG_UART0 was renamed to
OFFBAND_LOG_MIRROR_UART (#888). Update this env's build_flags -- the old name is inert
and would leave this board with no log mirror."
*** [.../examples/companion_radio/main.cpp.o] Error 1
```

The workflow still sets the old name on line 83:

```yaml
PLATFORMIO_BUILD_FLAGS: "-DOFFBAND_FORCE_CAPLOG -DOFFBAND_BOOT_BEACON -DOFFBAND_MESHLOG_UART0"
```

and repeats it in the header comment on line 11.

## Why it went unnoticed

The tripwire in `LogMirrorUart.h` is doing exactly its job — it exists so a stale flag
name fails loudly instead of silently producing a board with no log mirror. Its own
comment reads:

> *"The rename itself is complete: no .ini/.h/.cpp in the tree sets the old name."*

The sweep covered `.ini`, `.h` and `.cpp` and **missed `.yml`**. The last successful
diag build was 2026-08-23, before the rename landed, and nothing dispatched the
workflow between then and 2026-09-01 — so the breakage sat undetected.

## Impact

This is the sanctioned route for handing a tester a diagnostic image without cutting a
release (#957). While it is broken there is **no way to get an instrumented build to a
tester at all**, which currently blocks the field diagnosis in #956 — open since
2026-08-23 with no new evidence since.

## Fix

One line, plus the matching comment:

- line 83: `-DOFFBAND_MESHLOG_UART0` -> `-DOFFBAND_LOG_MIRROR_UART`
- line 11: update the comment to the current name

## Verification

- Dispatch for an observer env and confirm the run reaches *Publish to the diag-builds
  pre-release*, not just that it compiles.
- Confirm a `.bin` lands in `out/` — `build.sh` masks `cp` errors, and the workflow's
  own guard step exists because a green build can otherwise produce nothing.
[...]

Issue #956, opened 2026-08-23
bug: observer 'sprocket' stops publishing to CoreScope after hours (1.5.0-beta5, post-#720/#848)


Strycher:
**DIAGNOSE** — field report from tester **shill**. Observer **"sprocket observer"** (CoreScope device name) stopped transmitting/publishing after a number of hours; CoreScope stopped receiving updates. Filed at DIAGNOSE per owner; not yet root-caused.

## Build — rules out the recent broker fixes
shill is on **offband 1.5.0-beta5**, which **postdates** #838/#848 (terminal-`Failed`) and #720 (dwell-reset rotation). Those fixes are already in his build, so **they are not candidates** for this. `[verified: owner-stated version 2026-08-23]`

## Evidence in hand — one serial capture (non-diagnostic build)
`debug_logs/shill Offband serial capture 2026-08-23-12-36.txt`, ~5 min at **~10 h uptime** (ts 36,235 s → 36,535 s), **no reboot** (uptime continuous).

- `[verified]` Radio RX alive: `RadioLibWrapper: noise_floor ≈ -90 dBm` throughout, plus `Clearing header/preamble IRQ`. The node is still hearing the mesh.
- `[verified]` GPS task alive: `[GPS] detected=0 …` every ~5 s.
- `[verified]` **Zero publish-path lines** — no `[hb]`, `[mqtt]`, WiFi/`StaConnected`, `[rot]`, `[stats]`, `heap`.

**Correction to an earlier over-read:** the absence of those lines is **NOT** evidence of a hang. shill is on a **non-diagnostic build**, which does not emit `[hb]`/`[mqtt]`/`[rot]`. So this log establishes only: device up at ~10 h, radio + GPS running, no reboot. It **cannot** tell us WiFi state, broker state, heap, or whether the publish task is alive. `[hypothesis: untested]` the publish pipeline (WiFi→MQTT) stalled; unconfirmed by this log.

## Diagnostic plan (highest-leverage first)

1. **Server-side, no device needed — we have CoreScope MQTT + DB access.** Query when `sprocket` last published, last topic/message, and last-seen. This dates the failure precisely and confirms it's a publish-stop vs a CoreScope-side gap. **This is the immediate next step and needs neither shill nor a new build.**
[...]

PR #1043, opened 2026-09-04
task(#1041): correct the diag-build log-mirror flag to the post-#888 name

Strycher:
Fixes the flag name `diag-build.yml` injects, so the diagnostic-build workflow compiles
again.

Parent bug: #1034 · Implementation task: #1041 (closed) · Verification: #1042 · Epic: #736

## What was broken

The workflow injected `-DOFFBAND_MESHLOG_UART0`. #888 renamed that flag to
`OFFBAND_LOG_MIRROR_UART` and left a deliberate tripwire in `src/helpers/LogMirrorUart.h`:

```c
#if defined(OFFBAND_MESHLOG_UART0)
  #error "OFFBAND_MESHLOG_UART0 was renamed to OFFBAND_LOG_MIRROR_UART (#888) ..."
#endif
```

So **every `workflow_dispatch` failed at compile** from that rename onward. The last
successful diagnostic build predates it, and no instrumented build could reach a field
tester — which is what #956 has been waiting on.

The tripwire did its job: without it the build would have succeeded and produced a
diagnostic image **silently missing its log mirror**.

## Why the rename missed it

The #888 sweep covered `.ini`, `.h` and `.cpp`. A workflow injects
`PLATFORMIO_BUILD_FLAGS` from **outside the build configuration the source tree can
see**, so no amount of grepping the tree would have found it. A note recording that is
added to the workflow header so the next flag rename does not repeat it.

## Changes

- `-DOFFBAND_MESHLOG_UART0` -> `-DOFFBAND_LOG_MIRROR_UART=1`
- header comment updated to the current flag name, plus the note above

One file, 10 insertions, 4 deletions.

## Gemini review — 4 findings, all addressed

Run with `scripts/llm-consult.py --backend gemini --model gemini-2.5-pro`; audit log at
`docs/llm-consultations/2026-09-04-1041-diag-build-flag-gemini-gemini-2.5-pro.log`
(`finish_reason: STOP`). Full write-up on #1041.

| # | Finding | Disposition |
|---|---|---|
| 1 | A local build cannot exercise the publish step; only an end-to-end dispatch is a valid test | **Accepted.** Scoped as #1042 rather than claimed here |
| 2 | Bare `-D` relies on the compiler's implicit `1`; brittle against a future guard change | **Fixed** — passed as `=1`, rebuilt green |
| 3 | Sweep every `PLATFORMIO_BUILD_FLAGS` injection site, not just the flag name | **Done** — `build.sh` lines 117 and 159 are the only others; debug `-U` undefines and version stamps, neither a diag flag, and both append rather than overwrite |
| 4 | Keep the added comment | **No change** |

Finding 2's failure case is the one worth reading: if the guard ever became a value
comparison, a bare `-D` would leave the mirror **silently disabled** — reintroducing
exactly the "quietly does nothing" class the tripwire exists to prevent.

## Evidence

Both observer envs built green with the corrected triad, read from the per-env SUCCESS
table rather than the exit code:

| Env | Result | Duration |
|---|---|---|
| `Heltec_v3_companion_observer_wifi` | SUCCESS | 00:02:34 |
| `heltec_v4_companion_observer_wifi` | SUCCESS | 00:01:37 |

Zero error lines — the `#error` tripwire does not fire. `heltec_v4` rebuilt green
(00:01:03) after the `=1` change.

## What this PR does NOT prove

That a dispatch publishes a downloadable asset. A local build cannot reach the
`gh release upload` step, so that claim is not made here — it is #1042, which runs
against this branch.

Agent: CloudyGrove (session fcc043c7)

### DIFF
--- a/.github/workflows/diag-build.yml
+++ b/.github/workflows/diag-build.yml
@@ -6,9 +6,15 @@ name: Diagnostic Build (on-demand)
 #
 # The diagnostic triad is injected via PLATFORMIO_BUILD_FLAGS, matching the
 # standing diag surface (owner directive 2026-08-22):
-#   -DOFFBAND_FORCE_CAPLOG   caplog pinned ON (serial/mesh log capture)
-#   -DOFFBAND_BOOT_BEACON    raw-UART0 boot beacon (survives early-boot crashes)
-#   -DOFFBAND_MESHLOG_UART0  mirror the mesh log to UART0
+#   -DOFFBAND_FORCE_CAPLOG      caplog pinned ON (serial/mesh log capture)
+#   -DOFFBAND_BOOT_BEACON       raw-UART0 boot beacon (survives early-boot crashes)
+#   -DOFFBAND_LOG_MIRROR_UART   mirror the mesh log to UART0
+#
+# NOTE (#1034): the mirror flag was OFFBAND_MESHLOG_UART0 until #888 renamed it.
+# LogMirrorUart.h carries an #error tripwire on the old name, so this workflow
+# failed at compile on every dispatch between that rename and 2026-09-03. When a
+# build flag is renamed, sweep .yml too -- workflows inject
+# PLATFORMIO_BUILD_FLAGS from outside the build config the tree can see.
 # The boot beacon + pinned caplog make a diag build self-announce at runtime, and
 # the output filename/artifact carry "-diag" so it is never mistaken for a release.
 # (A version-STRING diag marker is a firmware follow-up -- build.sh stamps the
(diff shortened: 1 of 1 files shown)


Strycher:
Agent: CloudyGrove (session fcc043c7)

Strycher:
Human approval received from Ben in session fcc043c7 via `/merge #1043`.

Re-validated against the remote immediately prior: head `513939dc` (unchanged since approval), 1 commit, 1 file (`.github/workflows/diag-build.yml`), base `firmware-base`, mergeable=MERGEABLE, mergeStateStatus=CLEAN, checks 57/57 SUCCESS with 0 failures.

Verification #1042 passed against this exact head before merge — both observer envs dispatched, all three critical workflow steps succeeded, and the published assets download anonymously (HTTP 200, full bytes). Results recorded on #1042.

Agent: CloudyGrove (session fcc043c7)

Issue #1040, opened 2026-09-04
task: Epic verification — CI matrix + non-C6 hardware confirmation of the ZLP fix


Strycher:
Epic verification child of #1037. Depends on the Fix task.

Verify the ZLP fix as the Epic's unit:
- `ci-green` matrix passes on the Fix PR (all curated envs, incl. the JTAG-console and UART-bridge boards + nRF52 degrade path).
- Hardware confirmation on a board whose `Serial` is the native USB-CDC console on a chip family other than C6 (an S3 configured for USB-CDC-on-boot) — C6 already 0/90.
- Human sign-off.

Blocks Epic + #1035 closure.

Refs #1037, #1035.

Issue #1039, opened 2026-09-04
task: terminate USB-CDC console replies with a ZLP (flushSerialConsole)


Strycher:
Fix child of #1037.

Add `src/helpers/CdcConsoleFlush.h::flushSerialConsole()` — drain the HWCDC ring, wait (2 ms bounded) for the FIFO to become writable, then emit the terminating ZLP via `usb_serial_jtag_ll_txfifo_flush()`. Gated on `SOC_USB_SERIAL_JTAG_SUPPORTED && ARDUINO_USB_CDC_ON_BOOT`; degrades to a plain flush on UART-bridge consoles (e.g. Heltec V3). Called at the four serial-CLI reply sites (repeater, companion, room server, sensor). Retains the `emit` diag command.

Bench: fix on → 0/90 holds on C6. Compile-verified C6/S3/nRF52. Gemini 2.5 Pro reviewed (one finding — the CDC_ON_BOOT guard — applied).

Implemented in commit 48651ec6.

Refs #1037, #1035.

Strycher:
Fix merged to firmware-base in PR #1044 (merge commit 3430d851, 57/57 CI green). Console ZLP terminator (flushSerialConsole) live at the four CLI reply sites. Citadel Crosswire-1za already closed on commit-land. Companion writeFrame() path is a separate scope question under epic #1037 (owner deciding). Agent: VioletBarn (session 82e07da7).

Issue #1038, opened 2026-09-04
task: diagnose the USB-CDC 64-multiple console-reply hold


Strycher:
Diagnosis child of #1037.

Root cause: the ESP32 USB-Serial-JTAG console (Arduino HWCDC) does not emit the terminating zero-length packet after a console reply whose on-wire length is a multiple of the 64-byte USB max packet size, so the host holds the final packet ("stuck in host memory", per Espressif). Held-not-lost, host-side, intermittent. `HWCDC::flush()` only drains the ring; no released arduino-esp32 (≤3.3.8) emits the ZLP.

Bench-verified on rcc6-bench-1 (C6): fix off, 64-multiple replies hang 10–33%; the mechanism is `usb_serial_jtag_ll_txfifo_flush()` again when writable (the Espressif-documented recipe). Full trail in #1035.

Status: DONE.

Refs #1037, #1035.

Strycher:
Diagnosis done (Crosswire-eh0=done; root cause in #1035); both downstream fixes merged — #1044 (console) and #1093 (companion, 3534a642). Closing per owner. Agent field was never recorded; asking the sessions that touched #1038 to self-stamp via AM. Agent: VioletBarn (session 82e07da7)

Issue #1037, opened 2026-09-04
epic: USB-CDC console reply held on 64-byte multiples (ESP32 USB-Serial-JTAG)


Strycher:
## Epic: USB-CDC console reply held on 64-byte multiples (ESP32 USB-Serial-JTAG)

Build-and-test-as-one-unit deliverable: **diagnose → fix → verify** the host-side hold of console replies whose on-wire length is an exact multiple of the 64-byte USB max packet size on the ESP32 USB-Serial-JTAG console (Arduino HWCDC). Held-not-lost, host-side (`usbser.sys`), intermittent; it strands synchronous serial tooling on ×64-length replies.

**Symptom this resolves:** #1035 — rf_sweep records a canonical row with no firmware identity when the `version` read returns only its echo. Root cause is this hold.

### Child tasks
- Diagnosis — root cause (HWCDC emits no terminating ZLP on a 64-multiple) + bench reproduction. **DONE.**
- Fix — `flushSerialConsole()` ZLP terminator at the four serial-CLI reply sites.
- Epic verification — `ci-green` matrix + S3 (native-USB-console) hardware confirmation + human sign-off.

### Acceptance
Fix merged; `ci-green` green; a 64-multiple console reply delivers immediately (bench: 0 holds, verified on C6); #1035's version-read-miss root cause resolved.

Refs #1035.

Strycher:
## Scope note from #1053: as written, this epic's fix would not reach a `_usb` companion

Raising this because #1053's tester symptom was about to be triaged onto this epic, and the fix as scoped does not cover his path.

`[verified: gh pr diff 1044, 2026-09-09]` PR #1044 adds `flushSerialConsole()` at four sites, and all four are **text console** reply paths:

| File | Site |
|---|---|
| `examples/companion_radio/MyMesh.cpp` | `checkObserverSerialCli()`, after `Serial.println(reply)` |
| `examples/simple_repeater/main.cpp` | after `Serial.print("  -> ")` |
| `examples/simple_room_server/main.cpp` | same |
| `examples/simple_sensor/main.cpp` | same |

[...]

PR #1044, opened 2026-09-04
epic(#1037): terminate USB-CDC console replies with a zero-length packet

Strycher:
Epic **#1037** — diagnose → fix → verify the ESP32 USB-Serial-JTAG console reply hold. This PR carries the Fix commit; Epic verification (#1040) gates the merge.

## What & why

On the ESP32 USB-Serial-JTAG console (Arduino HWCDC), a console reply whose on-wire length is an exact multiple of the 64-byte USB max packet size is emitted as full 64-byte packets with **no terminating zero-length packet (ZLP)**. The host (Windows `usbser.sys`) then holds the final packet until the next write — documented by Espressif as data "stuck in host memory". Held-not-lost, host-side, intermittent; it strands synchronous serial tooling (pio-flash identity read, provisioning) on ×64-length replies. This is the root cause of the #1035 symptom (rf_sweep records no firmware identity when the `version` read returns only its echo).

## The fix (#1039)

`src/helpers/CdcConsoleFlush.h::flushSerialConsole()`: drain the HWCDC ring, then wait (2 ms bounded) for the FIFO to become writable and emit the terminating ZLP via `usb_serial_jtag_ll_txfifo_flush()` — the Espressif-documented recipe from `usb_serial_jtag_ll.h`, the same primitive the **unreleased** arduino-esp32 `master` HWCDC ISR uses (no released version ≤3.3.8 has it; we are pinned to 3.1.3 / IDF 5.3.2 via pioarduino). `HWCDC::flush()` alone does **not** fix this — it only drains the ring.

Gated on `SOC_USB_SERIAL_JTAG_SUPPORTED && ARDUINO_USB_CDC_ON_BOOT` — the chip has the peripheral **and** `Serial` is its USB-CDC console; degrades to a plain flush on UART-bridge consoles (e.g. Heltec V3 = esp32-s3 + CP2102) and non-ESP32. Called at the four serial-CLI reply sites (repeater, companion, room server, sensor). Retains an `OFFBAND_FORCE_CAPLOG`-gated `emit` diag command used to reproduce/A-B the issue.

## Structure

| Child | Status |
|---|---|
| #1038 Diagnosis | done — root cause + bench reproduction |
| #1039 Fix (this commit) | implementation complete |
| #1040 Epic verification | **open — gates this merge** (CI matrix + non-C6 hardware + human sign-off) |

## Evidence

- **Bench-verified on ESP32-C6 (rcc6-bench-1):** fix OFF → 64-multiple replies hang 10–33% of reps; fix ON → **0/90** (bulk write, delay=1000 µs, 30 reps × {128,192,256}). A first cut used a single `writable()` check and left a 1/30 residual; hardened to the 2 ms bounded wait → 0/90.
- **Compile-verified** across C6 (`rcc6` repeater + companion), S3 (`Heltec_v3`, `Ebyte_EoRa-S3` room), and the sensor example; nRF52 exercises the plain-flush degrade path.
- The `ARDUINO_USB_CDC_ON_BOOT` guard is compile-time and keeps the C6 code path identical (rcc6 sets it =1), so the 0/90 result carries over.

## Gemini adversarial review

`gemini-2.5-pro` (log: `docs/llm-consultations/2026-09-04-1035-usb-cdc-zlp-fix-gemini-gemini-2.5-pro.log`). Verdict: production-quality. **One finding — guard scope** (chip-capability-only; a UART-bridge board would poke the idle JTAG peripheral). **Fixed** by adding the `ARDUINO_USB_CDC_ON_BOOT` gate — the repo's existing precedent for HV3. All other points (ISR race, 2 ms bound, `micros()` rollover, HAL include chain, `static inline` placement) passed with no change.

## Not in this PR / gating merge

- **#1040 Epic verification** must pass before merge: `ci-green` matrix + hardware confirmation on a non-C6 board whose `Serial` is the native USB-CDC console + human sign-off.
- #1035 (rf_sweep symptom) closes when this epic lands; its fail-open half is a separate task (`Crosswire-5ho`), untouched here.

Refs #1037, #1039, #1035. Merges: **Ben**, after Epic verification.

### DIFF
--- a/examples/companion_radio/MyMesh.cpp
+++ b/examples/companion_radio/MyMesh.cpp
@@ -4,6 +4,7 @@
 #include <Mesh.h>
 #include <MeshLog.h>  // #396: serial-capture buffer access for the caplog download
 #include <helpers/ClockSanity.h>  // #607: owner-path clock sets + audit log
+#include "helpers/CdcConsoleFlush.h"   // #1035: right-flush (ZLP) for the USB-Serial-JTAG console
 
 // Offband fork-only companion-API frame codes (config 0xC0 / GPS 0xC1 / block 0xC2)
 // + shared enums. Self-contained (only <stdint.h>); included unconditionally so the
@@ -3713,6 +3714,7 @@ void MyMesh::checkObserverSerialCli() {
       char reply[256];
       offband::cliPassthroughExecute(_obs_cli_buf, reply, sizeof(reply));
       Serial.println(reply);
+      flushSerialConsole();   // #1035: emit the USB-CDC terminator so a 64-multiple reply isn't held host-side
       _obs_cli_len = 0;
       _obs_cli_redact = false;
       continue;
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -18,6 +18,7 @@
 // ((void)0) otherwise -- so non-diag builds of this role are byte-identical.
 #define OFFBAND_BEACON_DEFINE_CTOR
 #include "helpers/BootBeacon.h"
+#include "helpers/CdcConsoleFlush.h"   // #1035: right-flush (ZLP) for the USB-Serial-JTAG console
 
 
 #ifdef PIN_STATUS_LED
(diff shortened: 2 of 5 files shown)


Strycher:
Agent: VioletBarn (session 82e07da7). Epic #1037 Fix commit (41014155). Merge gated on Epic verification #1040 (ci-green matrix + non-C6 hardware + human sign-off).

Strycher:
Human approval received from Ben (session 82e07da7). Re-validated against the remote immediately prior: 1 commit (41014155), 5 files, mergeable=MERGEABLE / mergeState=CLEAN, 57/57 checks SUCCESS (ci-green passed). Enabling auto-merge --rebase. Note: Epic #1037 verification (#1040) and SunnyHollow's companion writeFrame() scope question remain open by owner decision — this lands the console fix only.

Issue #1115, opened 2026-09-10
task: verify -- stock client shows zero protocol corruption, and the #1053 tester confirms (#1087)


Strycher:
Parent: #1113. Verification task — the test that proves bug #1087 is dead. **Bug #1087 cannot close until this does.**

## Pass criteria
1. Stock MeshCore client over USB on an ESP32 `_usb` companion: the client debug log shows **zero** "unexpected frame type" entries across connect, sync and idle. — **PROVEN 2026-09-10** on rc32-bench-1.
2. Board stays healthy on an independent UART0 capture throughout: no reset, no silence. — **PROVEN 2026-09-10.**
3. The fix exercised by the #1053 tester on his own hardware. — **NOT DONE.**

## Explicitly out of scope
Settings and the channel list still fail to populate after this fix. That is #1093 (VioletBarn), a distinct defect in `writeFrame` underneath this one. It is not a failure of this verification. This task must not be held open waiting for #1093, and must not be closed as though it covered it.

Agent: SunnyHollow (session e75a0ac9)

Issue #1114, opened 2026-09-10
task: fix -- guard CrashLog and raw setup prints off the shared console (#1087)


Strycher:
Parent: #1113. Fix task for bug #1087.

## Change
- `src/helpers/diagnostics/CrashLog.cpp` — `crashLogSerialWrite()` returns before touching `Serial` when `meshLogMirrorEnabled()` is false. The UART0 mirror above it is untouched, so diag builds keep every byte.
- `examples/companion_radio/main.cpp` — `meshLogSetMirror(false)` is the first logging-affecting statement in `setup()` on `ENABLE_USB_INTERFACE` builds, closing the boot window. The build stamp and the nRF52 reset-reason line route through `crashLogf()` when the console carries the protocol.

## What "done" means here
The code is written, builds, and has been reviewed. It does **not** mean the fix is verified in the field — that is the sibling Verification task, which stays open.

## Evidence
- Green: heltec_rc32_companion_radio_usb, heltec_rc32_companion_radio_ble, heltec_rc32_repeater, RAK_4631_companion_radio_usb.
- Gemini: two rounds, audit logs `2026-09-09-1087-crashlog-protocol-guard` and `2026-09-10-1087-crashlog-guard-r2` in `docs/llm-consultations/`.

Agent: SunnyHollow (session e75a0ac9)

Strycher:
Fix implemented and committed: `bd2dcaf1` on `epic/1113-crashlog-protocol-stream`. Green on heltec_rc32_companion_radio_usb, heltec_rc32_companion_radio_ble, heltec_rc32_repeater and RAK_4631_companion_radio_usb; Gemini-reviewed twice.

Closing per CLAUDE-BASE line 286: a child Task closes when its commit lands on the Epic branch.

**This is not a claim that the fix is verified.** That is the sibling #1115, which stays open, and bug #1087 cannot close until it does. Owner-authorized 2026-09-10.

Agent: SunnyHollow (session e75a0ac9)

Issue #1113, opened 2026-09-10
epic: keep log text out of the companion protocol stream on USB companions (bug #1087)


Strycher:
Peer Epic for bug #1087, per CLAUDE-BASE "Bugs that need real work". Parent: #623 (RadioCore RC32), where it was found and reproduced. The defect itself affects every ESP32 `_usb` companion.

## Unit
Keep log text out of the framed companion protocol stream on USB companions, and prove it on hardware.

## Children
- **Fix** — implemented and committed. Closes when its commit lands.
- **Verification** — the test that proves the bug dead. Stays open until it passes and the owner signs off.

Bug #1087 is blocked-by this Epic and closes only when the Verification child does.

## Status, 2026-09-10
- **Diagnosis: complete.** Root cause verified in source and on hardware; the stock client's "unexpected frame type" values decode as our own ASCII boot line.
- **Fix: written**, green on 4 envs, Gemini-reviewed twice (round 1 rejected on initialization order and drove the real fix; round 2's substantive findings refuted with evidence), flashed to rc32-bench-1.
- **Verification: partial.** Proven: "unexpected frame type" went from a continuous stream to zero. Not proven: the fix in front of the #1053 tester.

Agent: SunnyHollow (session e75a0ac9)

Issue #1093, opened 2026-09-10
task: does the USB-CDC 64-multiple hold affect the companion writeFrame() path?


Strycher:
Child of Epic #1037. Filed per owner decision (2026-09-10). **Solve-side is pending SunnyHollow's determination** — companion/client side or firmware `writeFrame()` side.

## The gap
PR #1044 fixed the USB-CDC console-reply ZLP hold at the four **text-console** reply sites. A `_usb` companion does not use those — its data path is the **framed protocol**, `ArduinoSerialInterface::writeFrame()` on `Serial` (HWCDC). `[verified: src/helpers/ArduinoSerialInterface.cpp:24-65]` `writeFrame()` writes `>` + len(2) + payload with **no flush / no terminator**, so a frame whose on-wire length `3 + len` is a multiple of 64 could hold host-side by the same mechanism the console fix addresses. The merged `CdcConsoleFlush.h::flushSerialConsole()` (self-contained) would cover it.

## Determination pending (SunnyHollow)
Whether the companion-write symptom is solved **on that side** (client/DTR — SunnyHollow's #1053 lead is a DTR reset on port-open by the web client, NOT this hold) or **here** (a firmware `writeFrame()` terminator). It is **untested** whether the hold reproduces on `writeFrame()` at all — the `emit` reproducer is wired to the repeater console, never the frame path.

## Steps once the side is determined
1. **Reproduce:** an `emit`-*frame* variant that writes a `3 + len ≡ 0 (mod 64)` frame with nothing following, on a `_usb` companion; classify held vs delivered.
2. If it holds → call a `flushSerialConsole()`-style terminator after `writeFrame()` (guarded, same as the console path).
3. If it is client/DTR-side → resolve there; close this with that evidence.

Same root mechanism as #1037; kept in-epic per owner.

Refs #1037, #1035, #1053.

Strycher:
Hardware evidence from the #1053 / #1087 bench line, moved here from a duplicate I filed (#1094) before seeing this issue. This issue is the holder; I am not taking it.

## Context: #1087 is fixed and confirmed, and it uncovered this

#1087 (CrashLog writing plain text into the framed protocol on `_usb` builds) is fixed, flashed and **confirmed on hardware**. The stock client's debug log now has **zero** "unexpected frame type" entries where it previously had a continuous stream of them. The channel is clean.

What remains is underneath it, and it is a separate failure.

## Bench setup

[...]

PR #1119, opened 2026-09-10
epic(#1113): keep log text out of the companion protocol stream on USB companions

Strycher:
> ## ⚠ NOT READY TO MERGE
> The fix is implemented and its narrow claim is proven on hardware. **Epic verification (#1115) is still open** — the fix has not been in front of the #1053 tester. Merge is the owner's call after verification and sign-off.

Epic #1113 · bug #1087 · Fix task #1114 (closed) · Verification #1115 (open)

## What and why

On a `_usb` companion, `Serial` **is** the framed companion protocol. Log text was being written into it, so a stock MeshCore client parsed our own debug output as frames and reported one "unexpected frame type" per byte. Decoding a bench capture: `117 112 59 32 114 101 115 101 116` is `up; reset` — our own `[boot]` line.

It was severe rather than cosmetic because CrashLog installs an ESP-IDF log capture at boot, so the entire IDF stream went down the unguarded path. One capture held 498 lines.

MeshLog already solved this: `main.cpp` sets its mirror from `isConsoleSharedWithProtocol()`. CrashLog and two raw prints never asked. This reuses that single predicate rather than inventing a second policy.

## Changes

- **`src/helpers/diagnostics/CrashLog.cpp`** — `crashLogSerialWrite()` returns before touching `Serial` when the console is shared. The UART0 mirror above it is untouched, so a diag build keeps every byte on the sniffer wire.
- **`examples/companion_radio/main.cpp`** — on `ENABLE_USB_INTERFACE` builds, `meshLogSetMirror(false)` is now the **first** logging-affecting statement in `setup()`, before any output. The build stamp and the nRF52 reset-reason line, both raw `Serial.print` calls that bypassed both loggers, now route through `crashLogf()` when the console carries the protocol.

## Evidence

**Build:** green on `heltec_rc32_companion_radio_usb`, `heltec_rc32_companion_radio_ble`, `heltec_rc32_repeater`, `RAK_4631_companion_radio_usb`.

**Hardware (rc32-bench-1, ESP32-S3 USB-Serial-JTAG), stock MeshCore client over USB:**

| | Before | After |
|---|---|---|
| "unexpected frame type" in client log | continuous stream | **zero** |
| Board on independent UART0 capture | healthy | healthy — no reset, no silence |

**Not proven by this PR:** settings and the channel list still fail to populate. That is #1093 (VioletBarn), a separate defect in `writeFrame` underneath this one, and it is deliberately out of scope here.

## Gemini adversarial gate — two rounds

**Round 1 rejected the fix, correctly.** It found the guard was established at `main.cpp:532` while `crashLogStandardInit()` runs at `:351` and the build stamp at `:287`, so every boot had a window in which the exact text under investigation was emitted before the guard existed. → **Fixed**: the safe state is now the default from the first line of `setup()`. `ENABLE_USB_INTERFACE` is a sound proxy because `usb_serial_interface.begin(Serial)` is unconditional on that path.

Round 1 also found the two raw `Serial.print` calls in `setup()`. → **Fixed.**

Round 1 claimed the CrashLog → MeshLog include was a layering violation that would break host builds. → **Rejected with evidence**: the native env already compiles both deliberately (`platformio.ini:245` and `:254`), with a comment stating MESH_DEBUG routes into `mesh_log_line()`. The dependency predates this change.

**Round 2**, on the revised code:

- *Guard is after `Serial.begin()`, leaving a window.* → **Refuted**: a grep of the span between `Serial.begin(115200)` and the guard finds no Serial writes. The fragility point is fair for future edits and is addressed by a comment pinning the guard as the first logging-affecting statement.
- *`ENABLE_USB_INTERFACE` over-broad; a USB+BLE build would silence its BLE debug console.* → **Refuted**: no env combines them. The `_companion_radio_usb_` and `_companion_radio_ble_` envs are strictly separate in every variant.
- *Routing the build stamp through `crashLogf()` hides it from a casual serial monitor.* → **Refuted**: on a `_usb` build `Serial` carries binary frames, not a human-readable console, so the stamp was never usefully visible there. It still reaches the retained ring and the UART0 mirror.
- *Third-party libraries can still write to `Serial`.* → **Out of scope**: pre-existing and unbounded; not something this change claims to address.

Audit logs: `docs/llm-consultations/2026-09-09-1087-crashlog-protocol-guard-*` and `2026-09-10-1087-crashlog-guard-r2-*`.

## Unaffected

`_ble` companions: `Serial` is a debug channel there, the mirror stays true, logging continues. Non-companion roles never call `meshLogSetMirror`, so the flag stays true and their behavior is unchanged.

Refs #1087, #1053, #1093, #1083.

Agent: SunnyHollow (session e75a0ac9)

### DIFF
--- a/examples/companion_radio/main.cpp
+++ b/examples/companion_radio/main.cpp
@@ -256,17 +256,50 @@ void setup() {
   Serial.setTxTimeoutMs(1);
 #endif
 
+#if defined(ENABLE_USB_INTERFACE)
+  // #1087: FAIL SAFE BEFORE THE FIRST LINE OF OUTPUT. This must stay the first
+  // statement in setup() that can affect logging -- everything below it, the
+  // build stamp included, depends on the guard already being in place.
+  //
+  // The authoritative test is isConsoleSharedWithProtocol(), but it compares the
+  // interface's bound stream against Serial and so cannot be asked until
+  // usb_serial_interface.begin(Serial) has run -- roughly 250 lines below, after
+  // crashLogStandardInit() has already emitted the boot banner and replayed any
+  // previous-boot ring. Gemini review found that window: a guard set after the
+  // damage is not a guard.
+  //
+  // ENABLE_USB_INTERFACE is decidable here and is sufficient, because begin() is
+  // called with Serial unconditionally on this path -- so on these builds the
+  // console IS the protocol, always. Start closed and let the authoritative
+  // check below re-affirm it, rather than starting open and hoping nothing
+  // speaks first.
+  meshLogSetMirror(false);
+#endif
+
   // #149: stamp the running build on the serial console at boot. Every test build
   // looks identical otherwise, so there's no way to confirm what's actually flashed.
+  // #1087: raw Serial.print, so it bypasses both loggers -- route it through
+  // crashLogf() where the console carries the framed protocol. The stamp still
+  // reaches the UART0 mirror and the retained ring, so nothing is lost on a diag
+  // build; it simply stops being injected into the client's frame decoder.
 #ifdef OFFBAND_VERSION
-  Serial.print("\n=== Offband build: "); Serial.print(OFFBAND_VERSION);
-  #ifdef OFFBAND_GIT_SHA
-  Serial.print(" sha "); Serial.print(OFFBAND_GIT_SHA);
-  #endif
-  Serial.println(" ===");
+  if (meshLogMirrorEnabled()) {
+    Serial.print("\n=== Offband build: "); Serial.print(OFFBAND_VERSION);
+    #ifdef OFFBAND_GIT_SHA
+    Serial.print(" sha "); Serial.print(OFFBAND_GIT_SHA);
+    #endif
+    Serial.println(" ===");
+  } else {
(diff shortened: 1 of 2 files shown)


Strycher:
**Human approval received from Ben in session e75a0ac9, 2026-09-10.** Re-validated against the remote immediately prior: 1 commit (`bd2dcaf1`), 2 files (`examples/companion_radio/main.cpp`, `src/helpers/diagnostics/CrashLog.cpp`), mergeable=MERGEABLE, 57/57 checks SUCCESS including `ci-green`.

**Verification deliberately deferred by the owner.** This merges with Epic verification #1115 still open: the fix has not yet been exercised by the #1053 tester. The owner made that call knowingly, to stop the branch drifting further while other work lands on base. It is a recorded decision, not an oversight. What *is* proven on hardware: the stock client's "unexpected frame type" entries went from a continuous stream to zero.

Branch is BEHIND base (#1098, comment-only, no overlap with these hunks) and base protection is strict, so it is being updated before auto-merge. **Epic #1113, bug #1087 and Verification #1115 stay open** — merging closes nothing.

Agent: SunnyHollow (session e75a0ac9)

Issue #1118, opened 2026-09-10
task: verify -- both pads proven on hardware, RC32 and RC52 positions (#1085)


Strycher:
Parent: #1116. Verification task — proves bug #1085 is dead. **Bug #1085 cannot close until this does.**

## Pass criteria
1. RC32 / RCC6 position: target on pad A, `rx_a` climbs, `rx_b` stays 0. — **PROVEN 2026-09-10** on rc32-bench-1 (`rx_a` climbing, `rx_b=0`, v4 banner).
2. RC52 position: target on pad B, `rx_b` climbs, `rx_a` stays 0. — **NOT DONE.** Needs rc52-bench-1 on the pad B tap.
3. Banner prints the resolved GPIOs: pad A GPIO38, pad B GPIO39 on the plain Feather ESP32-S3. — **not yet confirmed from a capture.**
4. The `>>> [src:X]` marker prints once at the first byte. — **not yet confirmed.**
5. The default-on idle pull-up produces no spurious bytes on the quiet pad. — **supporting evidence only**: `rx_b` held at 0 across many minutes of captures, but that was not a deliberate test.

Agent: SunnyHollow (session e75a0ac9)

Issue #1117, opened 2026-09-10
task: fix -- dual-pin RX, per-pad counters, retire SNIFFER_RC52 (#1085)


Strycher:
Parent: #1116. Fix task for bug #1085.

## Change
`tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino` — two RX-only UARTs (both `TX = -1`), per-pad `rx_a` / `rx_b` counters, `SNIFFER_RC52` retired, build ID `SNIFFER-v4`, idle pull-up via `gpio_pullup_en()` defaulting **on** (a documented deviation from #938's default-off, whose premise of no idle pad no longer holds), and a source-change marker rather than per-byte tags.

## What "done" means here
The code is written, compiles, and has been reviewed. It does **not** mean both pads are proven — that is the sibling Verification task, which stays open.

## Evidence
`esp32:esp32:adafruit_feather_esp32s3`, core 3.3.11: 395968 bytes flash, 56040 bytes RAM. Gemini audit log `2026-09-09-1085-sniffer-dual-pin` in `docs/llm-consultations/`.

Agent: SunnyHollow (session e75a0ac9)

Strycher:
Fix implemented and committed: `febab571` on `epic/1116-sniffer-dual-pin`. Compiles for `esp32:esp32:adafruit_feather_esp32s3` (395968 B flash); Gemini-reviewed with three findings acted on.

Closing per CLAUDE-BASE line 286: a child Task closes when its commit lands on the Epic branch.

**This is not a claim that the fix is verified.** Pad A is proven on an RC32; pad B has never been exercised. That is the sibling #1118, which stays open, and bug #1085 cannot close until it does. Owner-authorized 2026-09-10.

Agent: SunnyHollow (session e75a0ac9)

Issue #1116, opened 2026-09-10
epic: sniffer listens on both header positions, proven on both pinouts (bug #1085)


Strycher:
Peer Epic for bug #1085, per CLAUDE-BASE "Bugs that need real work". Parent: #623 (RadioCore RC32) — the sniffer is the capture instrument for the RC32 bench line and for #1076.

## Unit
The sniffer listens on both header positions at once, so it can never be pointed at the wrong pad, proven on both board pinouts.

## Children
- **Fix** — implemented and committed. Closes when its commit lands.
- **Verification** — proves both pads on hardware. Stays open.

Bug #1085 is blocked-by this Epic and closes only when the Verification child does.

## Status, 2026-09-10
- **Diagnosis: complete.** #938 specified dual-pin listening and closed without it landing; the sketch hardcoded `SNIFFER_RC52` and read `rx_bytes=0` for 13.3 hours against an RC32.
- **Fix: written**, compiles for the Feather ESP32-S3, Gemini-reviewed with 3 findings acted on, flashed by the owner.
- **Verification: partial.** Pad A proven on an RC32. Pad B never exercised.

Agent: SunnyHollow (session e75a0ac9)

Issue #1085, opened 2026-09-09
bug: #938's dual-pin sniffer fix never landed — the rig still compile-time picks one pin, and it produced a 13-hour blind capture


Strycher:
#938 is **closed**. Its fix was never implemented. The sketch on `firmware-base` still does the exact thing #938 said to stop doing, and today that cost a diagnosis.

## What #938 specified

> **Fix — listen on both positions at once, no switch.** The owner has no dipswitch to put in the chain, so the sketch stops choosing and listens to both.

Scope it named: a second UART opened RX-only on the alternate pad, both drained into one stream, **per-pin `rx_bytes` counters** in the heartbeat, and build ID **`SNIFFER-v3` to `SNIFFER-v4`**.

## What is actually on firmware-base

`[verified: tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino, 2026-09-09]`

| #938 said | Reality |
|---|---|
| Listen on both, no switch | `:113` `#define SNIFFER_RC52 1` — a compile-time either/or |
| Second UART, RX-only | `:748` one `Serial1.begin(...)`. No `Serial2`, no second `HardwareSerial` |
| Per-pin `rx_bytes` counters | `:269` a single `static uint32_t rx_bytes` |
| Build ID `SNIFFER-v4` | `:2`, `:726`, `:752` all still `SNIFFER-v3` |

The sketch's own header states the failure mode it was meant to remove: a rig soldered for one board "lands RX-to-RX on the other and receives **NOTHING** — which looks exactly like a dead board, and is the failure this sketch's #938 change exists to prevent." It then ships with the RC52 value hardcoded.

## What it cost, today

`[verified: live monitor on COM16, 2026-09-09]`
```
[hb] SNIFFER-v3 alive  n=4816  up=47910s  rx_bytes=0
```

Zero bytes across 13.3 hours and 4816 heartbeats while the rig was pointed at an RC32. With `SNIFFER_RC52` defined it listens on the **TX pad**; the RC32 tap lands on the **RX pad**. I was one step from reporting a dead tap cable to the owner. The single `rx_bytes` counter is exactly why it was ambiguous — the per-pin counters #938 specified would have said which wire was live at a glance.

## Fix

Implement #938 as written. Not a runtime switch command: the owner already ruled that out in #938, and dual-listen is the better design because it removes the mode question entirely instead of making it something someone has to remember to set.

[...]

PR #1120, opened 2026-09-10
epic(#1116): sniffer listens on both header positions, per-pad counters

Strycher:
> ## ⚠ NOT READY TO MERGE
> Pad A is proven on an RC32. **Pad B has never been exercised**, and that is half of what this change exists to do. Epic verification (#1118) is open and needs rc52-bench-1 on the pad B tap. Merge is the owner's call after verification and sign-off.

Epic #1116 · bug #1085 · Fix task #1117 (closed) · Verification #1118 (open)

## What and why

#938 specified that the Feather sniffer listen on **both** UART header positions at once, because Heltec reversed them between RC32/RCC6 and RC52. #938 closed without that landing. The sketch still compile-time selected one pad via `SNIFFER_RC52`, shipped hardcoded to the RC52 position, and a rig pointed at an RC32 reported `rx_bytes=0` for 13.3 hours — one step from being reported as a dead tap cable.

A single counter cannot separate "wrong pad" from "dead wire" from "silent board": all three read 0. That ambiguity was the real defect.

## Changes

`tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino`:

- Two UARTs, `Serial1` on pad A (RC32/RCC6) and `Serial2` on pad B (RC52), **both RX-only with `TX = -1`**, so neither pad can ever be driven.
- Per-pad `rx_a` / `rx_b` counters in the heartbeat.
- `SNIFFER_RC52` retired; there is no mode left to set wrong.
- Build ID `SNIFFER-v3` to `SNIFFER-v4` in the banner and PING reply.
- The banner prints both resolved GPIOs, since `TX`/`RX` are variant aliases that move silently if the wrong board is selected.
- Generic S3 WROOM gets GPIO15 as its second pin, justified against the documented pin exclusions.

## Evidence

**Build:** `esp32:esp32:adafruit_feather_esp32s3`, core 3.3.11 — 395968 bytes flash, 56040 bytes RAM.

**Hardware, pad A:** flashed by the owner; `rx_a` climbing continuously against rc32-bench-1 across multi-minute captures while `rx_b` held at 0.

**Not proven:** pad B, the banner GPIO line, the source marker, and the idle pull-up as a deliberate test. All listed in #1118.

## Gemini adversarial gate

- **`pinMode(INPUT_PULLUP)` after `begin()` can silently detach the pad from its UART.** Fixed: now `gpio_pullup_en()`, which touches only the pull. The failure mode would have been a deaf sniffer still printing a heartbeat.
- **The idle pull-up should default on.** Fixed, and a deliberate deviation from #938, which specified default-off to keep existing behavior. That premise held with one pin and no idle pad; this change creates a permanently idle pad, and a floating input frames noise and inflates the idle counter. Setting `SNIFF_IDLE_PULLUP 0` restores the original default.
- **Tag every byte with its source.** Rejected and solved differently: per-byte tags would break every existing capture and human reader. A marker prints only when the active pad changes, once in normal operation and repeatedly if both pads go live.
- `Serial2` availability, the GPIO15 choice and buffer timing were confirmed safe; no action.

Audit log: `docs/llm-consultations/2026-09-09-1085-sniffer-dual-pin-*`.

Refs #1085, #938, #1076.

Agent: SunnyHollow (session e75a0ac9)

### DIFF
--- a/tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino
+++ b/tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino
@@ -1,9 +1,11 @@
-// RC32 UART0 sniffer + remote reset -- Adafruit Feather ESP32-S3.
-// BUILD ID: SNIFFER-v3                                          (#740, #702)
+// RadioCore UART0 sniffer + remote reset -- Adafruit Feather ESP32-S3.
+// BUILD ID: SNIFFER-v4                                   (#740, #702, #938, #1085)
 //
 // v2 was listen-only. v3 adds a command surface so a host can assert the RC32's
 // RST and BOOT lines without a human pressing buttons, which is what unblocks
-// unattended reset cycles.
+// unattended reset cycles. v4 listens on BOTH header positions at once with a
+// byte counter per pad, so the rig never has to be told which board it faces
+// (#1085, implementing what #938 specified).
 //
 // FLASH WITH ARDUINO IDE, NOT PLATFORMIO. Two prior PlatformIO attempts put
 // `Serial` on the TinyUSB CDC peripheral while the enumerated port was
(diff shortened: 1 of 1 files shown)


Strycher:
**Human approval received from Ben in session e75a0ac9, 2026-09-10.** Re-validated against the remote immediately prior: 1 commit (`febab571`), 1 file (`tools/diag/sniffer/rc_uart_sniffer_v3/rc_uart_sniffer_v3.ino`), mergeable=MERGEABLE, 57/57 checks SUCCESS including `ci-green`.

**Verification deliberately deferred by the owner.** This merges with Epic verification #1118 still open: **pad B (the RC52 header position) has never been exercised**, and that is half of what this change exists to do. The owner made that call knowingly. It is a recorded decision, not an oversight. What *is* proven on hardware: pad A against an RC32, `rx_a` climbing with `rx_b` at 0.

Branch is BEHIND base (#1098, comment-only) and base protection is strict, so it is being updated before auto-merge. **Epic #1116, bug #1085 and Verification #1118 stay open** — merging closes nothing.

Agent: SunnyHollow (session e75a0ac9)

Issue #1141, opened 2026-09-11
epic: harden companion USB-CDC terminator verification — production-instance / main-loop-driven A/B


Strycher:
Follow-up to Epic #1037 / task #1093. Deferred **option B** from the #1093 `/pushpr` Gemini gate (owner picked option A — ship #1093 now, log this to come back to).

## Why
#1093 shipped the deferred, non-blocking USB-CDC frame terminator. Its hardware A/B used a diag reproducer (`emitw`) that instantiates its **own** `ArduinoSerialInterface` and services it with a **manual pump loop**. That faithfully exercises the real `writeFrame()` / `loop()` / `emitConsoleZlpIfReady()` code on hardware (0/40 held with the terminator vs 15/40 without), and the production loop-wiring is verified by code inspection:
- `MultiSerialInterface::loop()` calls `iface.instance->loop()` for every registered interface (`src/helpers/MultiSerialInterface.h`)
- `usb_serial_interface` is registered and `interface_manager.loop()` runs every main pass (`examples/companion_radio/main.cpp`)

But no test exercises the **production `usb_serial_interface` instance through `interface_manager.loop()`** at runtime — the Gemini reviewer's fair objection.

## Scope (evaluate + execute)
1. Rewire `emitw` so the **repeater's main loop** services the test interface (drop the manual pump) — tests the arm-in-`writeFrame` / emit-in-main-loop pattern at runtime, not a hand-pumped isolated instance.
2. Evaluate a **true companion-instance integration test**: flash `heltec_rcc6_companion_radio_usb_diag`, drive a naturally-64-multiple frame through the production path, observe delivery.
3. Fold in the minor #1093 Gemini nits: document the bounded, uncontended `availableForWrite()` cost in `loop()` (only while a terminator is pending); consider renaming `_zlp_ring_capacity` -> `_zlp_drain_threshold`; tighten the `enable()`/`disable()` state-machine comments.

## DoD
Option B executed (production-path A/B green) or consciously re-deferred with a recorded rationale; verification tightened or the split hardware-methods + wiring-inspection verification formally accepted.

Refs #1093, #1037.

PR #1143, opened 2026-09-11
task: defer the companion USB-CDC frame terminator (non-blocking) (#1093)

### LINKED ISSUE #1093: task: does the USB-CDC 64-multiple hold affect the companion writeFrame() path?

Strycher:
## What & why

A companion `_usb` frame whose on-wire length (`3 + payload`) is a multiple of the 64-byte USB max packet size is held host-side by `usbser.sys` until the next write — the same ESP32 USB-Serial-JTAG (HWCDC) no-ZLP behaviour #1044 fixed for text-console replies. The framed path can't borrow that fix: `writeFrame()` is the #149 non-blocking hot path, and `flushSerialConsole()` blocks (`Serial.flush` + up-to-2 ms wait).

**Fix (deferred, non-blocking):** `writeFrame()` arms a terminator on a 64-multiple frame; `ArduinoSerialInterface::loop()` emits the zero-length packet once the TX ring has fully drained (`availableForWrite()` back to the ring capacity) and the FIFO is writable. Drain is read from the ring — not inferred from ISR interrupt-enable state — so it's cross-core-safe and needs no critical section; the ring capacity is seeded from the HWCDC default (256) so it can't be under-learned under load. A stalled or not-yet-connected host just leaves the frame in the ring and `loop()` retries — never waits. Guarded on `OFFBAND_USJ_CONSOLE` + the shared-`Serial` check: a no-op on UART bridges / non-USJ boards, and correct across the USJ family (C3/S3/C6/H2).

`simple_repeater` gains `emitf`/`emitw` diag reproducers (kept as tools); `emitw` drives the real `writeFrame()`+`loop()` path the hardware A/B measures.

## Issues
- Closes #1093 · under Epic #1037 · follow-up Epic #1141 (tighter verification).

## Evidence
- **Build:** `heltec_rcc6_companion_radio_usb_diag` (C6) and `Xiao_S3_companion_radio_usb` (S3) green post-rebase.
- **Gemini adversarial gate:** 5 rounds converged the design (inline→threshold→quiescent→ring-drain→capacity-seed); the round-5 review verified correctness (race-free, disconnected-safe, no data corruption, under-learning defeated). A final pre-push review ran on the shipped diff.
- **Hardware A/B (rcc6-bench-1, ESP32-C6):** `emitw` drives the real `writeFrame()`+`loop()`. With the terminator serviced: **0/40 aligned (64,128) frames held → delivered**; without it: **15/40 held**; the 129-byte offset control never holds via the mechanism.

## Gemini pre-push findings — addressed / justified
1. **`emitw` uses a separate `ArduinoSerialInterface` + a manual pump, so the A/B doesn't exercise the production instance through `interface_manager.loop()`.** *Justified + tracked.* The A/B exercises the real class methods on real hardware; the production loop-wiring is verified by inspection — `MultiSerialInterface::loop()` calls `iface.instance->loop()` for every registered interface, `usb_serial_interface` is registered, and `interface_manager.loop()` runs every main pass. Same code, called the same way the pump calls it. A runtime test of the production instance is deferred to **#1141** (owner decision: ship this, log the tighter test).
2. **`loop()` calls `availableForWrite()` (a `tx_lock` mutex) on the hot path.** *Acknowledged, accepted.* It runs only while a terminator is pending (a brief window after ~1-in-64 frames) and the lock is uncontended on the single-writer companion; it's inherent to reading drain from the ring. Documentation/evaluation tracked in #1141.
3. **Minor:** `enable()`/`disable()` comment precision and a possible `_zlp_ring_capacity` → `_zlp_drain_threshold` rename — tracked in #1141.

## Not merging
This PR is for review; **Ben merges** (`/merge`).

### DIFF
--- a/examples/simple_repeater/main.cpp
+++ b/examples/simple_repeater/main.cpp
@@ -19,6 +19,7 @@
 #define OFFBAND_BEACON_DEFINE_CTOR
 #include "helpers/BootBeacon.h"
 #include "helpers/CdcConsoleFlush.h"   // #1035: right-flush (ZLP) for the USB-Serial-JTAG console
+#include "helpers/ArduinoSerialInterface.h"   // #1093: emitw drives the real writeFrame()+loop() path for the v4.1 A/B
 
 
 #ifdef PIN_STATUS_LED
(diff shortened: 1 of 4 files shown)


Strycher:

Strycher:
Human approval received from Ben in session 82e07da7. Re-validated against the remote immediately prior: 1 commit, 4 files, mergeable=MERGEABLE, mergeState=CLEAN, 72/72 checks SUCCESS. Agent: VioletBarn (session 82e07da7)

